PR #468, opened 2013-05-28
[MI] Scrape individual contact pages for Republican senator email

jakejohns:
This should add some emails to the senators. It makes another request to the individual contact pages of the legislator and scrapes their email address from there. 

It uses an xpath to find where the "mailto: link" should be and checks it to see if it is in fact a mailto link. 

This only works for Republicans, as I have not yet found a way to determine Democrat emails. In addition, while the Republicans sometimes publish an email address here, some do not, and some link to a contact form. 

As it stands, I believe this will add the availability of 9 emails. 

To write this, I had to extract the basic code, and  test it separately, as the methods described in the "Contributing" guide did not seem to work (eg.billy-update )

### DIFF
--- a/openstates/mi/legislators.py
+++ b/openstates/mi/legislators.py
@@ -103,5 +103,8 @@ class MILegislatorScraper(LegislatorScraper):
                            email=email)
 
 
+            if email:
+                leg.add_source(contact_url)
+
             leg.add_source(url)
             self.save_legislator(leg)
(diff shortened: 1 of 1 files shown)


rshorey:
Just came across this. Looks like the repubs page has totally changed and everyone is doing contact forms instead of e-mail addresses now, so closing. thanks!

PR #495, opened 2015-01-06
NE: A bill action of "Bill withdrawn" should map to the action type: bill:withdrawn

calebcassel:
Currently if the text is "Bill withdrawn", the type is "other". This should be "bill:withdrawn".


paultag:
Nice, thanks, @calebcassel ! You rock! Merged into master, we should see this update going forward!

PR #496, opened 2015-01-06
KY: Small events scraper fix, and error to prevent identical event scraping

mileswwatkins:
Currently all events are scraped three times, and attributed to multiple chambers. We can't fix this bug yet, though, since there isn't enough information up on the calendar site.


paultag:
try changing the signature to `def scrape(self, chamber, session):` - let's see if that helps first (since it's a trivial change)

PR #501, opened 2015-01-07
OK: 2015 update

JackBewley:


.gitignore
```diff
@@ -22,3 +22,4 @@ site/openstates
 +*.sublime*
 openstates.sqlite3
 dump/
+venv
```
paultag: I'll pull this in because it's harmless, but you should check out `virtualenvwrapper` -- it's really great! Helps with virtual env management

paultag:
I'll give this a test now, thanks! :D

mileswwatkins:
Thanks @JackBewley! Accepted, and then made a small tweak to your session and term slugs.
https://github.com/sunlightlabs/openstates/pull/503/files#diff-0

JackBewley:
Awesome. Thanks!

PR #503, opened 2015-01-07
OK: Tweaked session and term slugs

mileswwatkins:


paultag:
Hurm, despite this being different then before, it looks correct.  Merged up, thanks @mileswwatkins

PR #505, opened 2015-01-12
VT: Full rewrite of scrapers is complete, compatible with new session

mileswwatkins:
Committees, bills, legislators, and events are all ready to be scraped!


mroswell:
This is a small matter, but is there any way to lowercase the "and" words?
For instance:
Agriculture And Forest Products --> Agriculture and Forest Products

paultag:
@mroswell we usually mirror the state's data exactly -- if you find a better place to source that string, we can check out ways to pull from there :)

PR #506, opened 2015-01-12
SD: 90th Session grrrrrrrrrr-eat!

mileswwatkins:
Cleared for all (ie, both) scrapers to run


paultag:
`Jean Hunhoff` appears to be pulled in on 2015-2016

Holding on this PR

mileswwatkins:
I _believe_ that Jean Hunhoff isn't a mistake; went from the Senate (18th District) to the House (18th District), but still in the legislature as a whole. A different Hunhoff (Bernie) is now in the 18th Senate seat.

paultag:
good call, lookin' good. Merged.

Issue #387, opened 2012-07-16
RI: legislator url


ninapsoncak:
don't know if its possible to add legislator urls

http://www.rilin.state.ri.us/Ajello/
http://www.rilin.state.ri.us/Ajello/rep.html

Issue #384, opened 2012-07-13
LA: most legislators have no committees


twneale:
I think this is a recent quirk--maybe they changed their site. Could be low-hanging fruit, but not sure.

/la/legislators/LAL000053/regina-ashford-barrow/

PR #519, opened 2015-01-19
fixes for #384 and #387

alexandervlpl:


paultag:
Thanks, @alexandervlpl - in the future, please make seperate pull requests for different states, so that we can fix one PR without blocking other changes.

Thanks so much!

PR #523, opened 2015-01-21
VT: Bills with no activity now handled correctly

mileswwatkins:


paultag:
I'm seeing some funkey bill version stuff - looks like URL is in Name, and Name is in URL

Also, the URL looks invalid - we might have had a site change

mileswwatkins:
Which URL looks invalid?

PR #524, opened 2015-01-21
New Mexico leg/bills ready to go

rshorey:
there haven't been any votes yet besides this, which doesn't seem to match to any bill:

http://www.nmlegis.gov/Sessions/15%20Regular/votes/T3SN6.PDF

otherwise this is running and seems to be working correctly


paultag:
can we run `.strip()` on names? I see trailing spaces in names

rshorey:
added .strip() to names, also moved email from the top level to the office (happened to notice that while I was in there)

paultag:
:+1: - i'll take a look at the merge now

paultag:
Hurm; can we also get rid of `Senator` and `Representative` from the `full_name` field?

rshorey:
sen/rep removed from full_name

paultag:
:+1: awesome, thanks!

PR #525, opened 2015-01-22
AK: 29th Session cleared for departure

mileswwatkins:
No votes or documents have been created yet so the scraper might choke on those in the next week or two. However, the bill page hasn't changed since last session so we're probably all good!


paultag:
:+1: thanks @mileswwatkins ! :D

PR #526, opened 2015-01-22
Indiana

rshorey:
Super messy website, so the data should probably be checked more thoroughly than usual. I'm going to do some checking as well. Especially bills. They're super complicated.

Note that this scraper is still using requests for one call due to the header-passing weirdness in scrapelib.

Once we've found and fixed any data issues, we should be ready to flip on bills and legislators! Note however that IN is a pre-file state and they file things in an insanely complex way AND they have a slow website so we should not be surprised about occasional crashes. I've tried to be as elegant as possible about error handling, and have run it through once with no problem on my local.


rshorey:
forgot to add, we're not currently handling committee votes. it's going to take some work, so figured better to get it up and running and then get to those next.

Issue #392, opened 2012-07-19
NM: Deleted committees


paultag:
NMC000012 - REDISTRICTING COMMITTEE
NMC000038 - NATURAL GAS EMERGENCY INVESTIGATION TASK FORCE

Verify they're gone, please.

PR #527, opened 2015-01-23
fixes for #392 (MI committees) and AttributeError crash (GA committees)

alexandervlpl:
The scraper gets committee members from `<a>` tags. Looks like this broke when the site added `<a>` tags with phone numbers (`billy-update mi --committees` was crashing). Skipping those tags.


alexandervlpl:
Adding a fix for the crash in GA committees. The code was trying to `strip()` a `ArrayOfCommitteeListing` object, which failed. The object has a proper `__str__()` method and should be converted to a string before making the check.

paultag:
Thanks!

PR #531, opened 2015-01-26
KY: everything but events

rshorey:
Events not working because of Billy bug where committees don't resolve correctly when chamber is not passed. @paultag is filing that bug.

Also we're not currently parsing votes - I think the vote PDFs are terrible scans (example: http://www.lrc.ky.gov/record/15RS/SB1/vote_history.pdf)


rshorey:
actually hold off on this one please - was going to remove some extra spaces but forgot

paultag:
Uch; sorry. I merged this before I saw the comment

PR #532, opened 2015-01-26
KY: fixed double space issue

rshorey:
jumped the gun on the prev PR, now committee-member matching works for a few members incorrectly formatted on KY's site.


paultag:
In the future, we can do something like `re.replace("\s+", " ").strip()` on incoming names. I think we have a billy filter setup where we can do that

PR #533, opened 2015-01-26
CA: committees available and scraping

rshorey:
committees were unavailable and we had thus manually set to throw an exception. They're available now so commented out the exception. Also a few minor fixes due to changes in class names.


paultag:
:+1: painless!

PR #536, opened 2015-01-27
ME: updated link for house committee page

rshorey:
page didn't change but location did.

NOTE: the new page (legislature.maine.gov/house/hsecoms.htm) has the wrong legislature number at the top, but includes some newly elected members, so I'm willing to just go with it.


paultag:
:+1: thanks!

PR #539, opened 2015-01-27
WY: Bills and events back on-line

mileswwatkins:
One representative's name will have changed (I removed the "(SD02)" from the end of his name); will this require manual reconciliation?

### DIFF
--- a/openstates/wy/bills.py
+++ b/openstates/wy/bills.py
@@ -102,6 +102,12 @@ class WYBillScraper(BillScraper, LXMLMixin):
         except scrapelib.HTTPError:
             self.warning('no digest for %s' % bill['bill_id'])
             return
+        if all_text.strip() == "":
+            self.warning(
+                    'Non-functional digest for bill {}'.
+                    format(bill['bill_id'])
+                    )
+            return
 
         # Split the digest's text into sponsors, description, and actions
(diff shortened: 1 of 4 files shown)


openstates/wy/bills.py
```diff
@@ -165,7 +160,7 @@ def scrape_digest(self, bill):
                     actor = 'upper'
 
                 date = datetime.datetime.strptime(date, '%m/%d/%Y')
-                bill.add_action(actor, action, date,
+                bill.add_action(actor, action.strip(), date,
```
paultag: at some point we should pull out entities from this:

e.g. `"action": "Introduced and Referred to H07 - Corporations",` => `Corperations`, (although, that wouldn't resolve right :) )

Doesn't need to change right now, I'll file a `DATA-` ticket

paultag:
Can you add `events` to `feature_flags`, so that we invoke the `events` scraper?

openstates/wy/bills.py
mileswwatkins: Agreed; it'd be nice, but I wasn't sure if there was a {comm abbrev: comm name} mapping that we could use, and there aren't enough bills yet in the new system to build one from scratch.

PR #540, opened 2015-01-28
MD: events throw warning and do not scrape; minor bugfixes

rshorey:
Should we also turn events off? The hand-indented, blob-of-text that is MD's events schedule has completely changed, and will have to be completely redone if/when we want to scrape events


paultag:
:+1: thanks!

PR #541, opened 2015-01-28
AK: added committee alias

rshorey:
everything should be able to run. the more times I type "committee" the less able I become to spell it.


paultag:
:+1: sweet! Thanks, @rshorey

PR #542, opened 2015-01-28
WV: 2015 bill scraper now functioning correctly

mileswwatkins:
Negative karmaaaaaa


paultag:
Phew, that one took a while to merge, sorry about that. Totally missed this.

PR #543, opened 2015-01-28
MI: no longer fails when crap is appended to mtg time

rshorey:


mileswwatkins:
Whoops, I didn't see this; just patched the bug myself, in a slightly different way.

mileswwatkins:
https://github.com/sunlightlabs/openstates/pull/544/files

PR #545, opened 2015-01-30
DE: legislators no longer fails

rshorey:
now skips empty/blank legislators


paultag:
Hurm, where was this coming from? Vacancies or HTML structural quirks?

This makes me a bit scared we'll drop all people if an xpath breaks (although I guess we'll just get a scrape failure for no people)

rshorey:
Seems like a html quirk. they accidentally put an extra row with only the value " ." in their table. I can make a more specific test for exactly that string?

rshorey:
fixed more narrowly.

paultag:
:+1: thanks!

PR #546, opened 2015-02-01
NV: scraper rewrite after site upgrade

alexandervlpl:
Fix for [DATA-387](https://sunlight.atlassian.net/browse/DATA-387). Updated bills and committees. Looks like they're still upgrading and some info might not be available, added workarounds.


paultag:
Sweet, thanks @alexandervlpl

PR #547, opened 2015-02-02
TX: bills/events can be turned on

rshorey:
seems that TX has stopped adding new data to their FTP. These are the same reports, scraped off the web.


paultag:
:metal:

PR #550, opened 2015-02-02
MN: house votes scraping

rshorey:
senate votes...not so much. they're buried in journals.


paultag:
:+1: thanks!

paultag:
Run issued

PR #555, opened 2015-02-05
UT: scraper should run, added related bills to events

rshorey:
Related bills was an easy lift while I was in there, @paultag assume this is something we want?

### DIFF
--- a/openstates/ut/events.py
+++ b/openstates/ut/events.py
@@ -64,7 +64,7 @@ class UTEventScraper(EventScraper, LXMLMixin):
                         )
                 bills = document.xpath('@onclick')
                 for bill in bills:
-                    if "2015/bills/static" in bill:
+                    if "bills/static" in bill:
                         bill_name = bill.split("/")[-1].split(".")[0]
                         event.add_related_bill(bill_name,
                             type='consideration',
--- a/openstates/ut/events.py
+++ b/openstates/ut/events.py
@@ -14,7 +14,9 @@ class UTEventScraper(EventScraper, LXMLMixin):
         events = doc.xpath('//item')
 
         for info in events:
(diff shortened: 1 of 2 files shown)


openstates/ut/events.py
```diff
@@ -14,7 +14,9 @@ def scrape(self, session, chambers):
         events = doc.xpath('//item')
 
         for info in events:
-            (title, when) = info.xpath('title/text()')[0].split(" - ")
+            title_and_date = info.xpath('title/text()')[0].split(" - ")
+            title = title_and_date[0]
+            when = title_and_date[-1]
```
paultag: Out of curiosity, what exists twixt these two?

openstates/ut/events.py
```diff
@@ -36,18 +38,38 @@ def scrape(self, session, chambers):
...
+                url = url.group(1)
                 event.add_document(
                         name=document.xpath('text()')[0],
-                        url=re.search(r'(http://.*?pdf)', document.xpath('@onclick')[0]).group(1),
+                        url=url,
                         mimetype='application/pdf'
                         )
+                bills = document.xpath('@onclick')
+                for bill in bills:
+                    if "2015/bills/static" in bill:
```
paultag: Hurm; can we tweak this to just `/bills/static`, or does that capture too much?

paultag: (just for future upgrade sanity)

openstates/ut/events.py
rshorey: I think it was a time

rshorey: but it was not consistent, just once

rshorey: fixed

paultag:
:+1: Thanks!

PR #556, opened 2015-02-05
MD: committee spacing issue fixed

rshorey:
Also included committees labelled as "other" as joint committees bc most seem to have members from both parties.

Also also, this will cause us to create a bunch of duplicate committees that will presumably have to be removed.

### DIFF
--- a/openstates/md/committees.py
+++ b/openstates/md/committees.py
@@ -16,15 +16,21 @@ class MDCommitteeScraper(CommitteeScraper):
         for a in doc.xpath('//a[contains(@href, "cmtepage")]'):
             url = a.get('href').replace('stab=01', 'stab=04')
             chamber = a.xpath('../../..//th/text()')[0]
+            com_name = a.text
+            if com_name is None:
+                continue
+            com_name = com_name.strip()
             if 'Senate' in chamber and 'upper' in chambers:
                 chamber = 'upper'
-                self.scrape_committee(chamber, a.text, url)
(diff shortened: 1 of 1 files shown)


mroswell:
Did you mean "joint committees bc most seem to have members from both **chambers**."

(I'm also curious: why would that create duplicates?)

Also what happens with the "Senate Special" and "House Special" committees at 
http://mgaleg.maryland.gov/webmga/frmcommittees.aspx?pid=commpage&tab=subject7

paultag:
@mroswell yeah, I assume @rshorey means chamber not party.

It would create duplicates since we don't delete committees if we don't scrape them, but they'll be empty. That's only a major concern if it's Legislators, since it will cause issues during name resolution, causing mass unlinking of related entities. We merge in that case, but we remove empty committees.

paultag:
Looks good, thanks @rshorey!

rshorey:
Special House/Senate Special committees will be attached to the relevant chamber, since we're checking for the word "House" or "Senate" in the header of each committee. There's a warning if the scraper fails to find a chamber, so we should get a heads up if any committees don't get assigned to a chamber.

mroswell:
I'm not seeing the special committees at
http://openstates.org/api/v1/committees/?state=md&apikey=APIKEY

Equivalent on my developing site:
http://legislativecommittees.org/?state=md

(Aside: I just alerted the webmaster for 
http://mgaleg.maryland.gov/webmga/frmcommittees.aspx?pid=commpage&tab=subject7 
of a typo in spelling "committee" in the Senate Special committee listing.)

PR #560, opened 2015-02-06
NY: random links as comm members fixed

rshorey:
used to be pulling in "Twitter", "Facebook", "Contact", "RSS"

There was no obvious way to do this without hard-coding the bad ones, so if a rep comes up with some other service they want to link to, we'll have to add that.

### DIFF
--- a/openstates/ny/committees.py
+++ b/openstates/ny/committees.py
@@ -19,6 +19,10 @@ def parse_name(name):
     >>> parse_name('Owen H.\\r\\nJohnson (Vice Chairperson)')
     ('Owen H. Johnson', 'vice chairperson')
     """
+
+    if name in ("Contact", "RSS", "Flickr", "Twitter", "Facebook", "YouTube"):
+        return (None,None)
+
     name = re.sub(r'^(Hon\.|Assemblyman|Assemblywoman)\s+', '', name)
     name = re.sub(r'\s+', ' ', name)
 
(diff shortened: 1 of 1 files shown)


paultag:
Hurm, I see what you mean;

Could we try passing if it has `/contact/` in the URL, or is that not consistant as well?

rshorey:
Looking into that, also fixing another thing I missed from the jira issue. Stand by.

paultag:
so standing

On Fri, Feb 6, 2015 at 2:49 PM, Rachel <notifications@github.com> wrote:

> Looking into that, also fixing another thing I missed from the jira issue.
> Stand by.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sunlightlabs/openstates/pull/560#issuecomment-73300819>
> .
>



-- 
Paul Tagliamonte
Software Developer | Sunlight Foundation

rshorey:
rewrote the NY senate committee scraper from scratch. looks like it was previously being contorted to use stuff that was written for the assembly. by rewriting it to be tailored to the Senate, we avoid the facebook/twitter issue, deal with the other problems in JIRA DATA-396, and make the whole thing a little easier to follow.

paultag:
Top notch; nice, thanks!

On Fri, Feb 6, 2015 at 3:33 PM, Rachel <notifications@github.com> wrote:

> rewrote the NY senate committee scraper from scratch. looks like it was
> previously being contorted to use stuff that was written for the assembly.
> by rewriting it to be tailored to the Senate, we avoid the facebook/twitter
> issue, deal with the other problems in JIRA DATA-396, and make the whole
> thing a little easier to follow.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sunlightlabs/openstates/pull/560#issuecomment-73307446>
> .
>



-- 
Paul Tagliamonte
Software Developer | Sunlight Foundation

PR #565, opened 2015-02-09
WA: events working

rshorey:
I'm NOT scraping related bills that are considered in the executive session, only in the public hearing. It is not the same set of bills. This is an easy fix/addition to line 87 though if we want to change.

### DIFF
--- a/openstates/wa/events.py
+++ b/openstates/wa/events.py
@@ -8,10 +8,10 @@ import re
 
 from billy.scrape.events import EventScraper, Event
 
-event_page = "http://www.leg.wa.gov/legislature/pages/showagendas.aspx?chamber=%s&start=%s&end=%s"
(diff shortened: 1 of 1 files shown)


openstates/wa/events.py
```diff
@@ -43,97 +43,53 @@ def scrape_agenda(self, ols):
...
+
         start = now.strftime(print_format)
-        then = now + timedelta(weeks=4)
-        end = then.strftime(print_format)
-        url = event_page % (
-            cha,
-            start,
-            end
-        )
+        end = (now+timedelta(days=30)).strftime(print_format)
```
paultag: seems sane, out of curisosity, how far into the future do these guys go?

rshorey: right now they're showing schedules for feb and march, but march is empty.

PR #567, opened 2015-02-09
HI: events no longer crashing

rshorey:
event featuring a special temp committee was causing some trouble.


paultag:
pedantic note: can we change this to use `.get`?

rshorey:
done

PR #568, opened 2015-02-10
PA: fixed some address stuff, added social media

rshorey:
Addresses in PA are kind of messy. Simplified the code and fixed some issues. I spot-checked a bunch of legislators, but it's possible that some different address-formatting issues will crop up.

(We're relying on a few words not being used in addresses, but I tried to be careful. If anyone moves their office to "Office Hours Rd." we're screwed.)

### DIFF
--- a/openstates/pa/legislators.py
+++ b/openstates/pa/legislators.py
@@ -53,132 +53,55 @@ class PALegislatorScraper(LegislatorScraper):
             legislator['email'] = '%s@%s%s' % tuple(vals)
 
     def scrape_offices(self, url, doc, legislator):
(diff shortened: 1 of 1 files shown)


rshorey:
Oh, also, social media stuff was there and easy to grab. I'm currently storing it at the top level for the legislator, and I'm storing the whole (eg www.twitter.com/username, not just username). I'm open to switching to usernames, but am a little concerned about 1) youtube IDs, which can either be a username or a channel, and 2) inconsistent link formatting that causes trouble for pulling out the username. I haven't seen evidence of 2 yet, so maybe that's not going to be a problem?

PR #569, opened 2015-02-10
AR: legislators with no party now being saved

rshorey:
fixes issue DATA-419

### DIFF
--- a/openstates/ar/legislators.py
+++ b/openstates/ar/legislators.py
@@ -47,9 +47,6 @@ class ARLegislatorScraper(LegislatorScraper):
         full_name = ' '.join(name_and_party[1:-1])
 
         party = name_and_party[-1]
-        if party == "()":
-            self.warning("Uch, bad party. Ducking")
-            return
 
         if party == '(R)':
             party = 'Republican'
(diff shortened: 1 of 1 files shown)


paultag:
Hurm.

I kinda don't like the idea of `unknown` in the DB. Do we know the party they're caucusing with? We might be able to hard-code this away, and break if we don't have an explicit mapping

rshorey:
yeah, we know in both cases, I'll make that happen. The issue is that we were totally missing the legislators if they had no party, but only throwing a warning so we didn't know about it until someone found a bug. Should I make it crash if it can't find a party?

mileswwatkins:
For hard-coding:
http://arkdems.org/people/state/
http://www.arkansasgop.org/index.cfm?p=state-house-candidates
http://www.arkansasgop.org/index.cfm?p=state-senate-candidates

rshorey:
@mileswwatkins did you make changes to prevent future "unknowns" from going into the db? Or does that still need fixing?

mileswwatkins:
Huh, my commits before `git good` don't show up on this PR, but this issue was corrected:
https://github.com/sunlightlabs/openstates/commits/master/openstates/ar/legislators.py

rshorey:
awesome, thanks.

openstates/ar/legislators.py
```diff
@@ -47,9 +47,6 @@ def scrape_member(self, chamber, term, member_url):
         full_name = ' '.join(name_and_party[1:-1])
 
         party = name_and_party[-1]
-        if party == "()":
-            self.warning("Uch, bad party. Ducking")
-            return
 
         if party == '(R)':
```
dogweather: A complicated if/then/else can be replaced by a dict with a mapping of before:after.

PR #570, opened 2015-02-10
MO: senate committees now scraping

rshorey:
slight rewrite of senate comm scraper


rshorey:
oh, also, fixes DATA-421

PR #571, opened 2015-02-10
PR: 2015 is finally functioning

mileswwatkins:


mileswwatkins:
#karmaneutral

mileswwatkins:
Also, please turn on the PR scraper on Bamboo once this pull request is accepted.

openstates/pr/committees.py
```diff
@@ -1,77 +1,111 @@
...
-            comm = None
-            for line in link.findall('text'):
-                text = line.findtext('b')
-                if text is not None and text.startswith('Comisi'):
-                    comm = Committee('upper',text);
+        lines = convert_pdf(filename, 'text').split('\n')
+        comm = None
+        comm_name = ''
+        title = ''
+        for line in (x.decode('utf8') for x in lines):
```
paultag: so, while I agree with this change (of course); we should likely file an issue on `convert_pdf` to accept an `encoding=` kwarg, so this isn't needed (unicode sandwich)

openstates/pr/committees.py
```diff
@@ -1,77 +1,111 @@
...
-                    if line.text and line.text.startswith('Hon.'):
-                        line_text = line.text.replace(u'–','-')
-                        name_split = line_text.split(u'-',1)
+                    self.save_committee(comm)
+                    comm = None
+                    comm_name = ''
+                comm_name = line
+
+                # Remove "Committee" from committee names
+                comm_name = comm_name.\
```
paultag: in the future, we can use `()` rather than `\` for multiline stuff

openstates/pr/legislators.py
```diff
@@ -57,100 +58,113 @@ def scrape_senate(self, term):
...
+                    name='Oficina del Capitolio',
+                    phone=phone,
+                    fax=fax,
+                    email=email
+                    )
+
+            self.save_legislator(leg)
+
+        # If this `assert` breaks, then remove the special-casing code
+        assert special_case_still_needed
```
paultag: We should change this to be:

`assert special_case_still_needed, "Yashira M. Lebrón Rodríguez doesn't need to be special cased anymore"` so we can scan the tracebacks

PR #574, opened 2015-02-11
OH: Events rewritten

mileswwatkins:


rshorey:
Scrape is working, but there's an issue when we try to resolve. I think the issue is that we need to pass a chamber when we add a committee as a participant (see lines 81 and 167 in oh/events.py).

mileswwatkins:
Is it working for you now? Doing an events-only local run is good for me after this change.

PR #576, opened 2015-02-12
AL: leg rewritten

rshorey:
data should probably be double checked. this one was ugly.

also I'm stuck on e-mail addresses for reps.

### DIFF
--- a/openstates/al/__init__.py
+++ b/openstates/al/__init__.py
@@ -29,6 +29,11 @@ metadata =  {
                       '2013rs', '2014rs'],
          'start_year': 2011,
          'end_year': 2014,
+        },
+        {'name': '2015-2018',
+         'sessions': ['2015os','2015rs'],
+         'start_year': 2015,
+         'end_year': 2018,
         }
     ],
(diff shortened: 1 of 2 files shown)


mileswwatkins:
@paultag: we can't remember; do we title() committee names, or lower() them?

paultag:
We should maintain casing as exists on the state site

mileswwatkins:
Gosh darn, @rshorey, I just can't figure out how representative pages are generated. I think that they're coming from a POST request to the main representative page with the headers:
```__EVENTTARGET:ctl00$HseMainContent$tabByName$TabPanel1$gvINFO```
```__EVENTARGUMENT:cmdRep$5``` (change that last number for the representative's alphabetical order in the current term)

If this sparks something for you on Tuesday then make that edit, but I'm guessing you already thought of that. Otherwise, removing the committee name lower() on line 215 is the only edit I see that needs to be made before pulling this branch in.

paultag:
Guys, it's OK to have a weekend :)

mileswwatkins:
MAKE ME

paultag:
:heart:

PR #592, opened 2015-02-27
WI: Added Feb 2015 Extraordinary Session

mileswwatkins:


paultag:
Can we put this into ingored until the session is active? This will result in us leaving data unscraped in the old session. We can always scrape / re-add this later

mileswwatkins:
@paultag, look good?

PR #595, opened 2015-02-28
Pfcass add mo events

cforcey:


.gitignore
```diff
@@ -22,4 +22,6 @@ site/openstates
 +*.sublime*
 openstates.sqlite3
 dump/
-venv
\ No newline at end of file
+venv
+.idea/
```
cforcey: You might want to put these in your global github ignore and revert commits to this file -- otherwise they will be not so useful for the main branch or anyone not using Pycharm: https://robots.thoughtbot.com/global-gitignore

openstates/mo/events.py
```diff
@@ -0,0 +1,155 @@
+import re
+import datetime
+import requests
+
+from lxml import html
+from openstates.utils import LXMLMixin
+from billy.scrape.events import EventScraper, Event
+
+
```
cforcey: We zap all double and trip white spaces in our code as a rule -- always feely fussy but over the long term I like the effect.

openstates/mo/events.py
```diff
@@ -0,0 +1,155 @@
...
+
+class MOEventScraper(EventScraper, LXMLMixin):
+    jurisdiction = 'mo'
+
+    keywords = ["Committee:", "Chair:", "Vice Chair:", "Time:", "Location:", "Note:"]
```
cforcey: Nice -- like seeing all these constant strings in one place since they are most likely to drift.

cforcey:
Sorry -- pulled internal branch over by accident.  Working on MO event scraper in fork.

PR #598, opened 2015-03-02
TX: Fix crash from irregular event time string

mileswwatkins:
Trivial change. I'm going to pull this in right away so that our (many) Texas dependents can have their information back up; the scraper's been down since mid last week.


paultag:
:+1: thanks!

PR #600, opened 2015-03-04
TX: Scrape fiscal notes, witness lists, and analyses

mileswwatkins:
Made an easy enhancement to grab all the other documents available for a bill. PR-review this _after_ #599, since this builds on the FTP crawling developed therein.

### DIFF
--- a/openstates/tx/bills.py
+++ b/openstates/tx/bills.py
@@ -1,115 +1,86 @@
-import re
-from urlparse import urljoin
 import datetime
 import ftplib
+import re
+import xml.etree.cElementTree as etree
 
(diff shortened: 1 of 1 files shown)


paultag:
"Hurm, the lack of caching really hurts running this for development"

paultag:
We should find a way to add caching. Or surface this through scrapelib. We really need to not not use scrapelib throughout the codebase.

paultag:
Since this is currently busted, I'm pulling this in so we can get production up again, but we should really change this to use scrapelib (even if it means patching SL a bit)

PR #602, opened 2015-03-05
OH: bill scraper rewritten for 131st session

rshorey:
now using their undocumented api, here:

http://search-prod.lis.state.oh.us/solarapi/v1/

### DIFF
--- a/openstates/oh/bills.py
+++ b/openstates/oh/bills.py
@@ -13,27 +13,27 @@ import lxml.html
 class OHBillScraper(BillScraper):
     jurisdiction = 'oh'
 
-    def scrape(self, chamber, session):
+    def scrape(self, session, chambers):
         if int(session) < 128:
             raise AssertionError("No data for period {}".format(session))
 
-        base_url = 'http://www.lsc.state.oh.us/status%s/' % session
+        status_report_url = "http://www.legislature.ohio.gov/legislation/status-reports"
 
-        bill_types = {'lower': [('hb', 'bill'),
-                                ('hjr', 'joint resolution'),
-                                ('hcr', 'concurrent resolution')],
-                      'upper': [('sb', 'bill'),
-                                ('sjr', 'joint resolution'),
-                                ('scr', 'concurrent resolution')]}
+        #ssl verification off due to requests bug
+        doc = self.urlopen(status_report_url,verify=False)
+        doc = lxml.html.fromstring(doc)
+        doc.make_links_absolute(status_report_url)
 
-        for bill_prefix, bill_type in bill_types[chamber]:
-            url = base_url + '%s.xlsx' % bill_prefix
+        status_table = doc.xpath("//div[contains(text(),'{}')]/following-sibling::table".format(session))[0]
+        status_links = status_table.xpath(".//a[contains(text(),'Excel')]/@href")
+
+
+        for url in status_links:
 
             try:
                 fname, resp = self.urlretrieve(url)
             except scrapelib.HTTPError:
-                # if there haven't yet been any bills of a given type
-                # then the excel url for that type will 404
+                self.logger.warning("Missing report {}".format(report))
                 continue
 
             sh = xlrd.open_workbook(fname).sheet_by_index(0)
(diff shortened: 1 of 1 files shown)


mileswwatkins:
@paultag, is it OS protocol to leave code in scrapers that doesn't deal with the present session? Here, `old_scrape()`, `scrape_versions_old()`, and `scrape_votes_old()` (about 200 of the 500 LoC) don't do anything for the current session, but _can_ grab the previous two sessions.

Asking since I haven't seen similar code in other scrapers, and my instinct personally would be to "retire" this code into the Git history and leave only have code that's directly relevant to current sessions in the master branch. But I can definitely see the other side of the argument.

rshorey:
I left the old_ versions in there because the interface has totally changed, the new data is not in the old format, and the old data is not in the new format, so there's no historical data if we don't keep them. I'm not tied to them, but if we lose them, we have no way to recover previous sessions' data.

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+
+                        #this stuff is version-specific
+                        version_name = bill_version["version"]
+                        version_link = base_url+bill_version["pdfDownloadLink"]
+                        mimetype = "application/pdf" if version_link.endswith("pdf") else None
+                        bill.add_version(version_name,version_link,mimetype=mimetype)
+
+                    self.save_bill(bill)
+
+    def pages(self,base_url, first_page):
```
mileswwatkins: What if we just made our API requests with a `?per_page=1000` (default is 50) instead of using a `pages` function? It doesn't seem to be disallowed, and it would result in fewer calls to their API and allow the removal of a whole method.

paultag: Well we could change per page and also use that function -- to reduce load.

Nice idea!

rshorey: so it loaded really slow when I tried to do that, and I'm worried that whatever n we set, we can't be totally sure we've got enough - even checking the number of bills in other sessions won't be enough to be sure, since this page actually displays *versions*, not bills. but would certainly be simpler. I can make this change if it feels right to y'all.

paultag: Well, not only -- I'd raise the per_page *and* keep the method, so we can both be sure, and also reduce HTTP requests, saving both sides time

mileswwatkins: I definitely saw the page lagging when I opened it in the browser, but I think it'd make the code much easier to upkeep going forward, and we'd be able to revert to the `pages()` function if scrapelib started failing on it.

rshorey: ok, I'm going to add the per_page parameter and keep pages, that's the decision? anyone have a gut feeling about a good value of n? maybe like 300? my browser really didn't like 1000

paultag: I doubt scrapelib would fail on it, it can read chunked, it'd fail with the json module if anything.

I want to keep the pages iterator, if we have more pages, we need to keep going, regardless of the value.

mileswwatkins: Why not just set it to a very high value (10000?), remove the recursive pagination code, and then raise an error if `nextLink` exists? The highest pagination I saw in my scrape was ~200, so I don't think that we'd ever hit an extremely high number.

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                                                    date,
+                                                    type = action_type,
+                                                    committees = committees)
+
+                            self.add_document(all_amendments, bill_id,"amendment",bill,first_page)
+                            self.add_document(all_fiscals,bill_id,"fiscal",bill,first_page)
+                            self.add_document(all_synopsis,bill_id,"synopsis",bill,first_page)
+                            self.add_document(all_analysis,bill_id,"analysis",bill,first_page)
+
+                            #todo: votes, committee votes
```
mileswwatkins: I think this comment isn't necessary anymore.

openstates/oh/bills.py
paultag: It's a few lines of code, we could patch it in the time it's taken us to talk about it :)

It's not a big deal, let's leave it as is and bump the per_page up a few. 300 sounds good. 

Let's :shipit:

mileswwatkins: Cool cool, checking a few more things.

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                            vote_url = base_url+bill_version["votes"][0]["link"]
+                            vote_doc = self.get(vote_url)
+                            votes = vote_doc.json()
+                            self.process_vote(votes,vote_url,bill,legislators,chamber_dict,vote_results)
+
+                            vote_url = base_url+bill_version["cmtevotes"][0]["link"]
+                            vote_doc = self.get(vote_url)
+                            votes = vote_doc.json()
+                            self.process_vote(votes,vote_url,bill,legislators,chamber_dict,vote_results)
```
mileswwatkins: I think that (at least some) committee votes are another layer deep; you have to navigate to each `link` _within_ their `cmtevotes` page. For example:
http://search-prod.lis.state.oh.us//solarapi/v1/general_assembly_131/resolutions/hcr5/cmtevotes

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                                sponsor_name = self.get_sponsor_name(sponsor)
+                                bill.add_sponsor("primary",sponsor_name)
+
+                            cosponsors = bill_version["cosponsors"]
+                            for sponsor in cosponsors:
+                                sponsor_name = self.get_sponsor_name(sponsor)
+                                bill.add_sponsor("cosponsor",sponsor_name)
+
+                            #actions
+                            blacklist = ["/solarapi/v1/general_assembly_131/resolutions/lr_131_0035-1/actions"]
```
mileswwatkins: I'm getting a 500 for `http://search-prod.lis.state.oh.us//solarapi/v1/general_assembly_131/resolutions/lr_131_0019/actions` as well. I'm surprised that the default `self.accept_response()` isn't handling this better, but maybe you could implement [something like in AL](https://github.com/sunlightlabs/openstates/blob/master/openstates/al/bills.py#L49)?

openstates/oh/bills.py
mileswwatkins: I don't think there are many non-committee votes, but this might happen for floor votes as well (once they pop up in the LIS). That said, I haven't seen an instance of this happening yet.

openstates/oh/bills.py
mileswwatkins: Actually, maybe this is a different matter; `accept_response()` is correctly not accepting the `500` responses, so maybe you need to just `continue` upon a `500` response?

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                            "refer_209":"other",
+                            "intro_108":["bill:introduced","bill:passed"],
+                            "intro_103": ["bill:introduced","bill:passed"],
+                            "msg_reso_503": "bill:passed",
+                            "intro_107": ["bill:introduced","bill:passed"],
+                            "imm_consid_360": "bill:passed"}
+
+            
+
+            base_url = "http://search-prod.lis.state.oh.us/"
```
mileswwatkins: Some of the URLs created have double-slashes: one slash from the end of `base_url` and another from the beginning of the path they're accessing. Maybe delete one or the other?

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+
+                            subjects = []
+                            for subj in bill_version["subjectindexes"]:
+                                try:
+                                    subjects.append(subj["primary"])
+                                except KeyError:
+                                    pass
+
+
+                            bill = Bill(session,chamber,bill_id,title,subjects=subjects,type=doc_type)
```
mileswwatkins: `bill_id` (at least in this instantiation argument) should be uppercase, and have a space between the letters and numbers. This still seems to be their convention, despite their LIS's API; [see a search page's return on their main website](https://www.legislature.ohio.gov/legislation?28&pageSize=10&start=1&sort=LegislationNumber&dir=asc&statusCode&legislationTypes=HB&generalAssemblies=131).

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                        #key=bill_id
+                        #value = dict of all versions, where keys are versionids
+                        #and values are the bill data from the api.
+                        #then loop through that to avoid duplicate bills
+
+                        bill_id = v["number"]
+                        version_id = v["versionid"]
+                        if bill_id in bill_versions:
+                            if version_id in bill_versions[bill_id]:
+                                self.logger.warning("There are two versions of {bill_id} called the same thing. Bad news bears!".format(bill_id=bill_id))
```
mileswwatkins: I think it'd be worth raising an error instead of just a warning here just since version name duplication would make the bill's data really unclear (and is also very unlikely to happen). But a not-necessary change, for sure.

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                            title = bill_version["shorttitle"] or bill_version["longtitle"]
+
+                            title = title.strip()
+
+                            chamber = chamber_dict[bill_version["chamber"]]
+
+                            subjects = []
+                            for subj in bill_version["subjectindexes"]:
+                                try:
+                                    subjects.append(subj["primary"])
```
mileswwatkins: Worth collecting other subjects? Or is `primary` the only field that holds them?

openstates/oh/bills.py
paultag: There are quite a few states that duplicate names (but **not** URLs) -- I see no problem in storing the data as shown on the state site

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                for b in bill_versions:
+                    bill = None
+                    for bill_version in bill_versions[b].values():
+                        if not bill:
+                            bill_id = bill_version["number"]
+                            title = bill_version["shorttitle"] or bill_version["longtitle"]
+
+                            title = title.strip()
+
+                            chamber = chamber_dict[bill_version["chamber"]]
```
mileswwatkins: Is the version list sorted ascending or descending by time? If descending (and the most recent version was from conference committee or the opposite chamber), could the chamber be incorrect for the `Bill`?

I guess we could always pull bill chamber from its `bill_id[0]`.

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                                action_doc = self.get(base_url+bill_version["action"][0]["link"])
+
+                                actions = action_doc.json()
+                                for action in actions["items"]:
+                                    actor = chamber_dict[action["chamber"]]
+                                    action_desc = action["description"]
+                                    try:
+                                        action_type = action_dict[action["actioncode"]]
+                                    except KeyError:
+                                        raise AssertionError("Unknown action {desc} with code {code}. Add it to the action_dict.".format(desc=action_desc, code=action["actioncode"]))
```
mileswwatkins: ARGH SO ELEGANT CAN'T TAKE IT

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                                if len(disapprove_json["items"]) > 0:
+                                    raise AssertionError("Whoa, a disapprove! We've never gotten one before. Go write some code to deal with it: {}".format(disapprove_url))
+
+
+
+
+                        #this stuff is version-specific
+                        version_name = bill_version["version"]
+                        version_link = base_url+bill_version["pdfDownloadLink"]
+                        mimetype = "application/pdf" if version_link.endswith("pdf") else None
```
mileswwatkins: I _think_ validation will error if there is no `mimetype` on a document, so maybe just turn this into an `assert` or tweak it in another way?

paultag: We can always fall back to 'application/octet-stream', which is saying "OMG
DATA" :)

On Thu, Mar 5, 2015 at 4:12 PM, Miles Watkins <notifications@github.com>
wrote:

> In openstates/oh/bills.py
> <https://github.com/sunlightlabs/openstates/pull/602#discussion_r25901871>
> :
>
> > +                                if len(veto_json["items"]) > 0:
> > +                                    raise AssertionError("Whoa, a veto! We've never gotten one before. Go write some code to deal with it: {}".format(veto_url))
> > +
> > +                            if "disapprove" in bill_version:
> > +                                disapprove_url = base_url+bill_version["disapprove"][0]["link"]
> > +                                disapprove_json = self.get(disapprove_url).json()
> > +                                if len(disapprove_json["items"]) > 0:
> > +                                    raise AssertionError("Whoa, a disapprove! We've never gotten one before. Go write some code to deal with it: {}".format(disapprove_url))
> > +
> > +
> > +
> > +
> > +                        #this stuff is version-specific
> > +                        version_name = bill_version["version"]
> > +                        version_link = base_url+bill_version["pdfDownloadLink"]
> > +                        mimetype = "application/pdf" if version_link.endswith("pdf") else None
>
> I *think* validation will error if there is no mimetype on a document, so
> maybe just turn this into an assert or tweak it in another way?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sunlightlabs/openstates/pull/602/files#r25901871>.
>



-- 
Paul Tagliamonte
Software Developer | Sunlight Foundation

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+        try:
+            documents = documents[bill_id]
+        except KeyError:
+            return
+        for item in documents:
+            if type_of_document == "amendment":
+                name = item["amendnum"] + " " + item["version"]
+            else:
+                name = item["name"] or type_of_document
+            link = first_page+item["link"]+"?format=pdf"
```
mileswwatkins: You might want to add a check to make sure the PDF version exists, although from what I've seen it does in all cases.

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+                vote["no_count"] += 1
+            if "absent" in v:
+                for voter_id in v["absent"]:
+                    vote["other_votes"].append(legislators[voter_id])
+                    vote["other_count"] += 1
+            if "excused" in v:
+                for voter_id in v["excused"]:
+                    vote["other_votes"].append(legislators[voter_id])
+                    vote["other_count"] += 1
+
```
mileswwatkins: Potential check to add: assert that there are no legislator IDs outside of the fields that we read from (`yeas`, `nays`, `absent`, `excused`). Just a thought, since the API and session are new and there may be vote objects that we haven't seen yet.

openstates/oh/bills.py
```diff
@@ -13,43 +13,333 @@
...
+    def add_document(self,documents,bill_id,type_of_document,bill,first_page):
+        try:
+            documents = documents[bill_id]
+        except KeyError:
+            return
+        for item in documents:
+            if type_of_document == "amendment":
+                name = item["amendnum"] + " " + item["version"]
+            else:
+                name = item["name"] or type_of_document
```
mileswwatkins: For fiscal notes and analyses, you could use the `legacyver` field to identify which stage of the bill the document's tied to, and make the names more descriptive, but this'd be icing.

openstates/oh/bills.py
rshorey: per_page parameter added at 300. (I'm making notes here as I finish these to keep track)

openstates/oh/bills.py
rshorey: yep. fixed. thanks.

openstates/oh/bills.py
rshorey: fixed.

openstates/oh/bills.py
rshorey: I added a try/except for a scrapelib.HttpError, which seems to have taken care of it. don't know why that was being weird before - I must have made some other changes that fixed that weirdness.

openstates/oh/bills.py
rshorey: error handling added for all cases. error handling with multi-level json uncertainty is messy :(

openstates/oh/bills.py
rshorey: So I'm not really sure how to do this. @mileswwatkins, any thoughts? The vote types are on the top level, there's no sub-level for votes. I don't like the idea of checking all keys to see if they point to an array of integers since that seems error-prone, but can't think of any other way to do this.

openstates/oh/bills.py
rshorey: add_document defaults mimetype to None. @paultag, do we prefer application/octet-stream?

paultag: It's OK to keep document MIME at None, but version has a required `mimetype` param.

openstates/oh/bills.py
mileswwatkins: Uhhhhh, good question :P

I'd probably go with looking for lists of four-digit integers in any of the fields, and then raise a `self.warning()` if one appeared? None of the other fields _should_ have that, AFAIR, but I understand that it's dicey anyway.

rshorey: OK, this is happening now

openstates/oh/bills.py
rshorey: done. there are indeed a few bad links!

openstates/oh/bills.py
rshorey: ok changed to octet-stream for version if it doesn't have a pdf extension

openstates/oh/bills.py
rshorey: done

openstates/oh/bills.py
rshorey: good call. pulling it off the bill id instead.

openstates/oh/bills.py
rshorey: done

rshorey:
OK, I've pushed all my changes. I think I hit all of Miles' comments. This was a messy one, thanks for the thorough review!

mileswwatkins:
:star2:

paultag:
Is this ready to go? Can we :shipit: ?

rshorey:
From my perspective, ready to go.

paultag:
@mileswwatkins did the review; seems only sensable he merge it!

PR #605, opened 2015-03-09
updated Louisiana events / legislators

iCandyLabs:
This should be reviewed to ensure that my changes proper.


openstates/la/__init__.py
```diff
@@ -20,13 +20,14 @@
...
-            "name": "2012-2015",
+            "name": "2012-2016",
             "start_year": 2012,
-            "end_year": 2015,
+            "end_year": 2016,
             "sessions": [
                 "2012",
                 "2013",
                 "2014",
+		"2015",
```
paultag:
```diff
+		"2015",
```

PEP8 convention is to use spaces -- not tabs

Could you change all the tabs to four spaces? I'm a bit suprised this runs, actually!

openstates/la/legislators.py
```diff
@@ -109,21 +111,24 @@ def scrape_lower_legislator(self, url, leg_info, term):
...
             "Independent": "Independent"
         }
 
-        if leg_info['name'].startswith("Vacant"):
+        if leg_info['name'].startswith(("Vacant","Miguez")):
             return
 
-        party = 'other'
+        #party = 'other'
+	party = 'Republican'
```
paultag:
```diff
-        partyblk = filter(lambda x: "District" in x,
-                          page.xpath('//p[@align="center"]//text()'))[0]
+        #partyblk = filter(lambda x: "District" in x,
+        #                  page.xpath('//td[@width="10%"]//text()'))[0]
```

and

```diff
-        party = 'other'
+        #party = 'other'
+	party = 'Republican'
```

What's up with these guys? Can we fix the party scraping bit?

openstates/la/__init__.py
iCandyLabs: Just this line?

openstates/la/legislators.py
iCandyLabs: The site has changed and I they are party affiliation is no longer on this page for lower chamber.

openstates/la/__init__.py
paultag: There should be no tabs in the file at all :)

openstates/la/legislators.py
paultag: Am I reading this as always hardcoding `Republican`?

iCandyLabs: That is correct.  It is only an attempt to "work around" until I could get your input.  The party affiliation is no longer listed on that page.

iCandyLabs: This is the page I am referring to for lower chamber party data scraping:
http://house.louisiana.gov/H_Reps/H_Reps_FullInfo.asp

As you can see the web team at the Louisiana legislature have neglected to offer the party affiliation info for House members.

Another, note is that this is also true of Upper chamber.

paultag: I don't like hardcoding everyone to Republican - we're widly used, we can't serve bad data like that. 

Let's get the PR fixing events in first. We'll talk over how to fix Legislators in a different PR

PR #607, opened 2015-03-09
removed special case for McGonagle, seems to be standard now

rshorey:
thanks VERY much to @mileswwatkins for including the following error:
"AssertionError: Remove the special-casing code for Joseph McGonagle"


paultag:
:+1:! Thanks, team awesome!

PR #611, opened 2015-03-09
fixed Events.py for Louisiana 2015 legislative session

iCandyLabs:
Please check code to ensure it is proper.

### DIFF
--- a/openstates/la/__init__.py
+++ b/openstates/la/__init__.py
@@ -22,11 +22,12 @@ metadata = {
         {
             "name": "2012-2015",
             "start_year": 2012,
-            "end_year": 2015,
+            "end_year": 2016,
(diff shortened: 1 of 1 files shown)


openstates/la/__init__.py
```diff
@@ -22,11 +22,12 @@
         {
             "name": "2012-2015",
             "start_year": 2012,
-            "end_year": 2015,
+            "end_year": 2016,
```
paultag:
```diff
-            "end_year": 2015,
+            "end_year": 2016,
```

I seem to recall LA being a max of 60 calendar days starting sometime in April,
so I think the end year here is 2015. I'll correct this before I merge.

paultag:
Great, thanks!

PR #612, opened 2015-03-10
OH: bills scraper more robust to things that don't really matter

rshorey:
not erroring on things we don't really care about any more.


paultag:
Nice, thanks!

PR #613, opened 2015-03-10
TX: Fix FTP timeout for bill document fetching

mileswwatkins:


openstates/tx/bills.py
```diff
@@ -22,14 +23,15 @@ class TXBillScraper(BillScraper):
     def _get_ftp_files(self, root, dir_):
         ''' Recursively traverse an FTP directory, returning all files '''
 
-        ftp = ftplib.FTP(self._FTP_ROOT)
+        for i in range(3):
+            try:
+                ftp = ftplib.FTP(root)
+                break
+            except (EOFError, ftplib.error_temp):
+                time.sleep(2 ** i)
```
paultag: You likely want an:

```
else:
    raise
```

-- this is why we have people review PRs :)

mileswwatkins: Cool :)

I figured since it would break anyway (line 26 `name ftp unknown`) it wasn't necessary, but I'll add this in.

PR #616, opened 2015-03-11
TX: Fix chamber mis-assignment

mileswwatkins:
Let's throw out all 84th-session `bill`s from the database, and then a fixed run will pull them all back in.


mileswwatkins:
@paultag, I'm going to pull this in, just so we minimize the amount of bad data in Open States between now and when we do the one-off database fix.

paultag:
LGTM; :+1:

PR #618, opened 2015-03-11
AL: Handle missing bill pages and fix vote fetching

mileswwatkins:


rshorey:
this looks fine to me

PR #621, opened 2015-03-12
TX: Fix time-string parsing, and remind us that we're dealing with regexes

mileswwatkins:
There was a tiny issue where we were replacing the word `"adjourn."`, and because of the replacement method the program replaced `r'adjourn.'` instead of the word `"adjourn"` with a `.`. I added `r""` notation to the other replacement strings to remind us that we're playing … the most dangerous game.


paultag:
pedantically, `r""` means `raw`, not `regex` :)

mileswwatkins:
Touche :P

PR #622, opened 2015-03-12
OH: Default action categorize to 'other'

mileswwatkins:
I'm unhappy about doing this, but over the last few days we've seen multiple new action codes per day, which has prevented _any_ of our builds from working, even if I add and rerun within a single day.


rshorey:
late to the party, but agreed. I'll pull and run on my local again soon and update some of them

PR #623, opened 2015-03-12
OK: Fix senate motion names, avoid fake votes

mileswwatkins:
`SENATE RES` and other such strings were getting captured as voters, and the wrong lines were being used for senate measure names.


openstates/ok/bills.py
```diff
@@ -155,11 +155,12 @@ def scrape_votes(self, bill, url):
                 motion_index = 8
             else:
                 chamber = 'upper'
-                motion_index = 9
+                motion_index = 13
```
rshorey: @mileswwatkins can you comment this? It's weird and takes some doing to figure out

PR #624, opened 2015-03-12
IN: scrapers running, but missing lots of stuff for bills

rshorey:
All scrapers using api.

Not dealing with anything that requires a pdf (versions, votes, documents) due to api-key related issues.


openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+                title = bill_json["latestVersion"]["shortDescription"]
+
+            #and if that doesn't work, use the bill_id but throw a warning
+            if not title:
+                title = bill_id
+                self.logger.warning("Bill is missing a title, using bill id instead.")
+
+
+            original_chamber = "lower" if bill_json["originChamber"].lower() == "house" else "upper"
+            bill = Bill(session,original_chamber,bill_id,title)
```
mileswwatkins: `bill_id` should be instantiated with a space between the letters and numbers, and with leading zeros stripped. For example, `SB 15` instead of `SB0015`.

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+            "hj":"resolutions/house/joint",
+            "sb":"bills/senate/",
+            "sr":"resolutions/senate/simple/",
+            "scr":"resolutions/senate/concurrent",
+            "sjr":"resolutions/senate/joint",
+            "sc":"resolutions/senate/concurrent",
+            "sj":"resolutions/senate/joint"
+            }
+        bill_id = bill_id.lower()
+        try:
```
mileswwatkins: You might be able to do this more simply with a filter, a la `bill_id = "".join([c for c in bill_id if c.isdigit()])`

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+        else:
+            bill_prefix = bill_id[:2]
+            bill_num = bill_id[2:]
 
-                if bill_title.strip() == "":
+        try:
+            url += urls[bill_prefix]
+        except KeyError:
+            raise AssertionError("Unknown bill type {}, don't know how to make url".format(bill_id))
+        url += bill_num
```
mileswwatkins: For URLs, `bill_num` needs to have no leading zeros. Eg, `http://iga.in.gov/legislative/2015/bills/senate/15` instead of `http://iga.in.gov/legislative/2015/bills/senate/0015`.

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+            #and if that doesn't work, use the bill_id but throw a warning
+            if not title:
+                title = bill_id
+                self.logger.warning("Bill is missing a title, using bill id instead.")
+
+
+            original_chamber = "lower" if bill_json["originChamber"].lower() == "house" else "upper"
+            bill = Bill(session,original_chamber,bill_id,title)
+            
+            bill.add_source(api_source,note="API key needed")
```
mileswwatkins: Sliiiick

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
-                try:
-                    bill_title = li.xpath('a/strong')[0].tail.rstrip().lstrip(': ')
-                except IndexError:
-                    continue
+    def make_html_source(self,session,bill_id):
+        url = "http://iga.in.gov/legislative/{}/".format(session)
+        urls = {
+            "hb":"bills/house/",
+            "hr":"resolutions/house/simple/",
+            "hcr":"resolutions/house/concurrent",
```
mileswwatkins: I think these `concurrent` and `joint` links need a `/` at the end.

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+            bill_id = b["billName"]
+            bill_link = b["link"]
+            api_source = api_base_url + bill_link
+            bill_json = client.get("bill",session=session,bill_id=bill_id.lower())
+            
+            title = bill_json["title"]
+            #sometimes title is blank
+            #if that's the case, we can check to see if
+            #the latest version has a short description
+            if not title:
```
mileswwatkins: I'm seeing some cases of `title` being `"NoneNone"`, but I'm not sure what's causing this. Check out HC 15 for an example.

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+                date = datetime.datetime.strptime(date,"%Y-%m-%dT%H:%M:%S")
+                
+                action_type = []
+                d = action_desc.lower()
+                committee = None
+                if "first reading" in d:
+                    action_type.append("bill:reading:1")
+
+                if ("second reading" in d
+                    or "reread second time"):
```
mileswwatkins: Oh noes! This will always resolve to `True`, so all actions will inadvertently be marked as `bill:reading:2`.

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+                action_desc = a["description"]
+                chamber = "lower" if a["chamber"]["name"].lower() == "house" else "upper"
+                date = a["date"]
+                date = datetime.datetime.strptime(date,"%Y-%m-%dT%H:%M:%S")
+                
+                action_type = []
+                d = action_desc.lower()
+                committee = None
+                if "first reading" in d:
+                    action_type.append("bill:reading:1")
```
mileswwatkins: It seems that resolutions (at least) can be voice-vote passed after just the first reading. Could add a check for `"adopted voice vote"` that is agnostic of reading number, maybe?

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+                if "third reading" in d:
+                    action_type.append("bill:reading:3")
+                    if "passed" in d:
+                        action_type.append("bill:passed")
+                    if "failed" in d:
+                        action_type.append("bill:failed")
+
+                if ("referred" in d and "committee on" in d
+                    or "reassigned" in d and "committee on" in d):
+                    committee = d.split("committee on")[-1].strip()
```
mileswwatkins: Not sure how `billy`'s internal matching works, but in `action_desc` the committee names are title-cased. Could this line be changed so that it references `action_desc` instead of `d`?

openstates/in/legislators.py
```diff
@@ -1,111 +1,47 @@
...
+            party = leg["party"]
+            link = leg["link"]
+            api_link = api_base_url+link
+            html_link = base_url+link.replace("legislators/","legislators/legislator_")
+            html = get_with_increasing_timeout(self,html_link,fail=True,kwargs={"verify":False})
+            doc = lxml.html.fromstring(html.text)
+            doc.make_links_absolute(html_link)
+            address, phone = doc.xpath("//address")
+            address = address.text_content().strip()
+            phone = phone.text_content().strip()
```
mileswwatkins: This `phone` data isn't added to the legislator. I don't think it'd be inappropriate to add an otherwise empty office (no address, no email, no fax, just phone), right?

mileswwatkins: I think you missed an `add_office()` call for this address and phone information.

openstates/in/legislators.py
```diff
@@ -1,111 +1,47 @@
...
+        client = ApiClient(self)
+        t = next((item for item in self.metadata["terms"] if item["name"] == term),None)
+        session = max(t["sessions"])
+        base_url = "https://iga.in.gov/legislative"
+        api_base_url = "https://api.iga.in.gov/{}".format(session)
+        chamber_name = "Senate" if chamber == "upper" else "House"
+        r = client.get("chamber_legislators",session=session,chamber=chamber_name)
+        all_pages = client.unpaginate(r)
+        for leg in all_pages:
+            print leg
```
mileswwatkins: Is this debug code?

openstates/in/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    #access the page we need
+    #if it's false, throw a warning and keep going
+    timeout_length = 2
+    html = None
+    while timeout_length < 65 and html is None:
+        try:
+            html = scraper.get(link,timeout=timeout_length,**kwargs)
+        except requests.exceptions.ConnectTimeout:
+            old_length = timeout_length
+            timeout_length **= 2
```
mileswwatkins: Wouldn't've guessed that this existed. Neato!

openstates/in/utils.py
```diff
@@ -0,0 +1,22 @@
+import requests
+
+
+def get_with_increasing_timeout(scraper,link,fail=False,kwargs={}):
```
mileswwatkins: Why do we use this function instead of a `scrapelib.Session()`?

openstates/in/committees.py
```diff
@@ -0,0 +1,115 @@
...
+                if mem_name not in comm_members:
+                    comm_members.append(mem_name)
+                    comm.add_member(mem_name)
+
+            api_source = api_base_url + comm_name
+            html_source = html_base_url + comm_name.replace("committee_","")
+            comm.add_source(api_source,note="requires API key")
+            comm.add_source(html_source)
+            self.save_committee(comm)
+
```
mileswwatkins: So a man walks into a bar, and the bartender looks at him and asks, "Why the blank space?"

openstates/in/committees.py
```diff
@@ -0,0 +1,115 @@
...
+            else:
+                chamber = "joint"
+
+            name = comm_json["name"]
+            try:
+                owning_comm = subcomms[name]
+            except KeyError:
+                comm = Committee(chamber,name)
+            else:
+                comm = Committee(chamber,owning_comm,subcommittee=name)
```
mileswwatkins: Should we strip the word "Subcommittee" from subcommittees?

openstates/in/committees.py
```diff
@@ -0,0 +1,115 @@
...
+            
+            chamber = comm_json["chamber"]["name"]
+            if chamber == "Senate":
+                chamber = "upper"
+            elif chamber == "House":
+                chamber = "lower"
+            else:
+                chamber = "joint"
+
+            name = comm_json["name"]
```
mileswwatkins: The standing committees don't start with `"Committee on "`, but the Statutory Committees do. I'd be in favor of removing `"Statutory Committee on "` from `name`.

openstates/in/committees.py
```diff
@@ -0,0 +1,115 @@
...
+                owning_comm = subcomms[name]
+            except KeyError:
+                comm = Committee(chamber,name)
+            else:
+                comm = Committee(chamber,owning_comm,subcommittee=name)
+
+            chair = self.process_special_members(comm,comm_json,"chair")
+            vicechair = self.process_special_members(comm,comm_json,"viceChair")
+            ranking = self.process_special_members(comm,comm_json,"rankingMinMember")
+
```
mileswwatkins: Is there any way we can `assert` that there are no more special titles? It'd be unfortunate to hardcode these and then miss other special members.

openstates/in/committees.py
```diff
@@ -0,0 +1,115 @@
...
+    jurisdiction = 'in'
+
+    def process_special_members(self,comm,comm_json,role_name):
+        try:
+            mem = comm_json[role_name]
+        except KeyError:
+            return
+        if mem:
+            person = mem["firstName"]+" "+mem["lastName"]
+            comm.add_member(person,role=role_name)
```
mileswwatkins: The `role`s are all spelled out (eg, `Ranking Minority Member` instead of `rankingMinMember`) on the committees' webpages (eg, `http://iga.in.gov/legislative/2015/committees/rules_and_legislative_procedures_1800`); we should use the proper names (`Chair`, `Vice Chair`, and `Ranking Minority Member`) instead of the JSON names.

openstates/in/committees.py
```diff
@@ -0,0 +1,115 @@
...
+            comm_members = [m for m in [chair,vicechair,ranking] if m]
+
+            for mem in comm_json["members"]:
+                mem_name = mem["firstName"]+" "+mem["lastName"]
+                if mem_name not in comm_members:
+                    comm_members.append(mem_name)
+                    comm.add_member(mem_name)
+
+            api_source = api_base_url + comm_name
+            html_source = html_base_url + comm_name.replace("committee_","")
```
mileswwatkins: If it is a subcommittee, I believe that `"committee_"` has to remain in the URL. The first is currently generated and fails, the latter is the actual URL:
`http://iga.in.gov/legislative/2015/committees/provider_services_sub3963`
`http://iga.in.gov/legislative/2015/committees/provider_services_subcommittee_3963`

openstates/in/committees.py
```diff
@@ -0,0 +1,115 @@
...
+        session = max(t["sessions"])
+
+        subcomms = self.get_subcommittee_info(session)
+
+        api_base_url = "https://api.iga.in.gov/{}/committees/".format(session)
+        html_base_url = "http://iga.in.gov/legislative/{}/committees/".format(session)
+        client = ApiClient(self)
+        r = client.get("committees",session=session)
+        all_pages = client.unpaginate(r)
+        for comm_info in all_pages:
```
mileswwatkins: There's a fake Senate committee named "Withdrawn" that's getting created when I run this locally. Not sure what its dealio is, but it's got a valid URL and everything. Could special-case ignore it.
`http://iga.in.gov/legislative/2015/committees/withdrawn_9901`

openstates/in/committees.py
```diff
@@ -0,0 +1,115 @@
...
+            comm_name = comm_link.split("/")[-1]
+            comm_json = client.get("committee",session=session,committee_name=comm_name)
+            
+            chamber = comm_json["chamber"]["name"]
+            if chamber == "Senate":
+                chamber = "upper"
+            elif chamber == "House":
+                chamber = "lower"
+            else:
+                chamber = "joint"
```
mileswwatkins: I don't see any joint committees for the state, even for 2014, so I think it'd be safe to change this from `chamber = "joint"` to a `raise` of some kind.

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
...
+
+                #blacklisted actions - don't correspond to anything we know
+                if ("authored" in d
+                    or "referred to the senate" in d
+                    or "referred to the house" in d
+                    or "sponsor" in d
+                    or "coauthor" in d
+                    or ("rule" in d and "suspended" in d)):
+                    #not really actions we track
                     continue
```
mileswwatkins: `pass` instead of `continue`? I'm thinking of situations where one of these blacklisted actions _and_ another action type appear in the same `d`; we'd want to still `add_action()` instead of skipping, right?

openstates/in/bills.py
```diff
@@ -45,198 +45,185 @@ class INBillScraper(BillScraper):
     # Can turn this on or off. There are thousands of subjects and it takes hours.
     SCRAPE_SUBJECTS = True
```
mileswwatkins: This isn't used anymore, so you can chuck it or re-implement it. I didn't notice any slowness due to scraping subjects, so maybe the comment is just out-of-date.

openstates/in/__init__.py
```diff
@@ -71,9 +67,16 @@
 )
 
 def session_list():
-    from billy.scrape.utils import url_xpath
-    # cool URL bro
-    return url_xpath('http://www.in.gov/legislative/2414.htm', '//h3/text()')
+    import requests
+    import os
+    apikey = os.environ['INDIANA_API_KEY']
```
mileswwatkins: Are we going to store Sunlight's INDIANA_API_KEY on Gitlab or something? I don't know where these sorts of things would live.

openstates/in/bills.py
rshorey: fixed

openstates/in/bills.py
rshorey: leading zero issue fixed

openstates/in/bills.py
rshorey: NoneNone comes straight out of their database. I'm treating those NoneNone's like missing titles now.

openstates/in/bills.py
rshorey: oops. Also this was masking a lot of missing issue information. I changed the error below this to a warning, too, so new ones don't crash this scraper all the time (see terrible times in ohio)

openstates/in/bills.py
rshorey: rewrote this all due to the everything-evaluating-to-true error that was masking a bunch of other stuff

openstates/in/bills.py
rshorey: subject scraping on/off removed

openstates/in/bills.py
rshorey: I'm not 100% that the "committee on" is consistently cased (or won't change someday, given other data consistency issues). So if Billy can handle potentially different casing for committees in the matching, I think it'd be safer to leave it this way. @paultag can you confirm? Otherwise I'll make the change.

openstates/in/bills.py
rshorey: added this and a check to make sure we're referring to a bill reading, not an amendment.

openstates/in/bills.py
rshorey: now in the format of SB 15

openstates/in/bills.py
rshorey: ok, changed, although it needs to happen twice because I need both the prefix and the number

openstates/in/bills.py
rshorey: :)

openstates/in/legislators.py
rshorey: I did indeed miss add_office. it's there now.

openstates/in/legislators.py
rshorey: yep, removed debug code

openstates/in/utils.py
rshorey: Does a scrapelib.Session() give me the option to retry a lot like this? I want to be able to hit them with increasingly long timeouts.

openstates/in/committees.py
rshorey: So I thought no so we know they're subcommittees but I don't know how we typically do this.

openstates/in/utils.py
mileswwatkins: I believe it should; check out `scrapelib.RetrySession`, from which `scrapelib.Session` inherits:
https://github.com/sunlightlabs/scrapelib/blob/master/scrapelib/__init__.py#L171

openstates/in/committees.py
mileswwatkins: Since they're already in the `subcommittee` field I had thought it was fine, but this is definitely something we should check with @paultag/@jamesturk.

openstates/in/committees.py
rshorey: and the horse replies "have you ever tried to type with hooves?"

openstates/in/committees.py
rshorey: "statutory committee on" is no more.

openstates/in/committees.py
mileswwatkins: "Nay"

openstates/in/committees.py
rshorey: I can't think of a way to catch errors here since they're fieldnames in the JSON, so we have to go looking for them specifically (I mean, I could loop through all JSON fieldnames and look for a person's name, but that feels like more risk than reward because who knows what else could show up).

Oh, also also, the positions seem to be null when not filled, so it's unlikely that there will be weird ones on just one committee.

openstates/in/committees.py
rshorey: role names fixed

openstates/in/committees.py
rshorey: Substrings are the worst. Fixed.

openstates/in/committees.py
rshorey: yeah, wasn't sure what to do there since it looked kinda official with that page and all? "withdrawn" is now special-cased out.

openstates/in/committees.py
rshorey: yeah, that feels right, no joint committees, thanks.

openstates/in/__init__.py
rshorey: re API_KEY, this might be working somewhere because the scraper wasn't failing before. Although maybe it just never called anything that needed that. Tossing this one to @paultag

openstates/in/utils.py
rshorey: yeah, **=2 is probably mean to future peeps and looks kind of like an emoticon (for dude with really weird facial hair and strange smile?). I'm leaving it but put a comment in :)

openstates/in/utils.py
rshorey: OK, I'll check this out in the morning and see if I can get it working that way.

openstates/in/committees.py
mileswwatkins: Good logic, makes sense to me.

openstates/in/committees.py
mileswwatkins: It's allegedly a standing committee, but isn't on the standing committee index page, so I feel pretty good that it's just a deprecated artifact.

mileswwatkins:
I'll do a local run of this now that you've patched it, but I feel like we've caught pretty much everything.

openstates/in/utils.py
rshorey: OK, @mileswwatkins and I have conferred on this in person - we're trying to do something different here than scrapelib sessions deal with, since we're extending the time we wait for a timeout, not the time between attempts. So I think this line can be deemed resolved.

openstates/in/committees.py
```diff
@@ -0,0 +1,110 @@
...
+                raise AssertionError("Unknown committee chamber {}".format(chamber))
+            name = comm_json["name"]
+            try:
+                owning_comm = subcomms[name]
+            except KeyError:
+                name = name.replace("Statutory Committee on","").strip()
+                comm = Committee(chamber,name)
+            else:
+                name = name.replace("Statutory Committee on","").strip()
+                comm = Committee(chamber,owning_comm,subcommittee=name)
```
mileswwatkins: Subcommittee names still have `" Subcommittee"` on the end of them, which should be removed.

openstates/in/committees.py
```diff
@@ -0,0 +1,110 @@
...
+            comm_json = client.get("committee",session=session,committee_name=comm_name)
+            
+            chamber = comm_json["chamber"]["name"]
+            if chamber == "Senate":
+                chamber = "upper"
+            elif chamber == "House":
+                chamber = "lower"
+            else:
+                raise AssertionError("Unknown committee chamber {}".format(chamber))
+            name = comm_json["name"]
```
mileswwatkins: For whatever reason, the House's Joint Rules has `"Committee on "` at the start of its name, although no others do (and neither does the Senate's Joint Rules).

openstates/in/legislators.py
```diff
@@ -1,111 +1,49 @@
...
+            html_link = base_url+link.replace("legislators/","legislators/legislator_")
+            html = get_with_increasing_timeout(self,html_link,fail=True,kwargs={"verify":False})
+            doc = lxml.html.fromstring(html.text)
+            doc.make_links_absolute(html_link)
+            address, phone = doc.xpath("//address")
+            address = address.text_content().strip()
+            address = "\n".join([l.strip() for l in address.split("\n")])
+            phone = phone.text_content().strip()
+            district = doc.xpath("//span[@class='district-heading']")[0].text.lower().replace("district","").strip()
+            legislator = Legislator(term,
```
mileswwatkins: You can add `photo_url` for everyone, which I believe is just replacing `/legislators/` with `/portraits/`. Hadn't noticed this earlier!

openstates/in/bills.py
```diff
@@ -42,201 +42,214 @@ class INBillScraper(BillScraper):
...
+                    or "signed by the president" in d
+                    or "signed by the speaker"
+                    or "authored" in d
+                    or "sponsor" in d
+                    or "coauthor" in d
+                    or ("rule" in d and "suspended" in d)
+                    or "removed as author" in d
+                    or ("added as" in d and "author" in d)
+                    or "public law" in d):
+                        action_type.append("other")
```
mileswwatkins: Do we need to add `"other"` to `action`s that already have a `type`? I would argue not, but I could see the other side of the argument.

mileswwatkins:
Alright, everything looks square outside of my most recent comments!

openstates/in/committees.py
rshorey: subcommittee removed.

openstates/in/committees.py
rshorey: So this kind of makes me want to leave it there - it's the official name of the committee according to the site, it's in the url, and that's how it's alphabetized. Unless there's strong objection or reason to go in a different direction, I'm going to consider this resolved.

mileswwatkins: Hmmm. It definitely _is_ anomalous in that it's the only one starting with `"Committee on "` on [the index page](https://iga.in.gov/legislative/2015/committees/). My argument would be that we do strip `"Statutory Committee on "` from others in that same list that have it in their official names. But you make a good point; I'm fine either way here. Professor @paultag?

openstates/in/bills.py
rshorey: agreed, action "other" not needed if something else is there, fixed.

openstates/in/legislators.py
rshorey: good catch. photo_urls added.

openstates/in/committees.py
paultag: Interesting.

I'm tempted to make it consistant, but I really don't like changing the official data. Let's keep it as is, but email them asking if that was intentional

rshorey: ok, in that case I've pushed my other changes, and I'll e-mail them. @mileswwatkins if you don't see other issues, I think we're ready to go.

PR #625, opened 2015-03-13
AR: Remove legislator special casing, add events

mileswwatkins:


openstates/ar/events.py
```diff
@@ -42,36 +38,43 @@
...
-
+    def scrape(self, session, chambers):
         url = "ftp://www.arkleg.state.ar.us/dfadooas/ScheduledMeetings.txt"
         page = self.urlopen(url)
         page = csv.reader(StringIO.StringIO(page.bytes), delimiter='|')
 
         for row in page:
+            # Deal with embedded newline characters, which cause fake new rows
+            LINE_LENGTH = 11
+            while len(row) < LINE_LENGTH:
```
rshorey: I'd be a bit more comfortable if we could make something bad happen if we end up with more than 11 cells so we can confirm that there were no legit formatting changes - throw an error if the eventual length isn't 11?

rshorey: forgot to @mileswwatkins that last one!

PR #630, opened 2015-03-17
AL: Improve testing for empty bill index pages

mileswwatkins:
This should fix some of the intermittent errors we've seen with Alabama.

### DIFF
--- a/openstates/al/bills.py
+++ b/openstates/al/bills.py
@@ -48,6 +48,7 @@ class ALBillScraper(BillScraper):
     CHAMBERS = {'H': 'lower', 'S': 'upper'}
     DATE_FORMAT = '%m/%d/%Y'
 
+    # Tweak which responses are acceptible to the scrapelib internals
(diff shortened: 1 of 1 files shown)


mileswwatkins:
Refactoring further, per suggestion of @paultag. Do not review until more commits have been pushed.

mileswwatkins:
@paultag/@rshorey: ready to go

openstates/al/bills.py
```diff
@@ -50,30 +49,25 @@ class ALBillScraper(BillScraper):
     DATE_FORMAT = '%m/%d/%Y'
 
     def accept_response(self, response, **kwargs):
```
rshorey: add a comment about how this is a scrapelib function we're redefining for easier future maintainability?

rshorey:
This looks OK to me. Also I learned a lot about the Alabama leg dress code by reviewing these bills :)

PR #633, opened 2015-03-19
Indiana: bills now have votes, versions and documents

rshorey:
questions (probably for @mileswwatkins):

1) should we rewrite the vote scraper not to refer to line numbers? It's failing in a very small number of cases and I think this is best, but I'm open to suggestion

2) they just make a lot of documents, so checking has been slow. I haven't done a super thorough job and am continuing to do some checks.


openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+                    for v in voters:
+                        if v.strip():
+                            vote[currently_counting].append(v.strip())
+
+            assert (len(vote["yes_votes"]) == vote["yes_count"],
+                "Yes vote counts ({count}) don't match count of actual votes ({actual})".format(count=vote["yes_count"],actual=len(vote["yes_votes"])))
+            assert (len(vote["no_votes"]) == vote["no_count"],
+                "No vote counts ({count}) don't match count of actual votes ({actual})".format(count=vote["no_count"],actual=len(vote["no_votes"])))
+            assert (len(vote["other_votes"]) == vote["other_count"],
+                "Other vote counts ({count}) don't match count of actual votes ({actual})".format(count=vote["other_count"],actual=len(vote["other_votes"])))
```
mileswwatkins: Same issue that the Prof noted in my recent pull request:
Even `assert (False, "message")` will pass, because Python assumes you're checking for an empty tuple, not that you're just doing a line continuation.

openstates/in/bills.py
```diff
@@ -36,6 +39,7 @@ def check_vote_counts(vote):
         pass
```
mileswwatkins: What if we did a `self.warning()` instead of `pass`ing on these?
Also, `check_vote_counts` can be replaced with `vote.validate()`, which checks for this exact issue. I mentally facepalmed when I found that out, since I've written at least three or four vote validators myself, haha.

openstates/in/bills.py
rshorey: ah, crap, I always forget that parentheses do unexpected things to asserts.

openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+            'OVERRIDDEN': True,
+            'ADOPTED': True,
+        }
+
+        for r in rollcalls:
+            proxy_link = proxy["url"] + r["link"]
+            (path, resp) = self.urlretrieve(proxy_link)
+            text = convert_pdf(path, 'text')
+            lines = text.split("\n")
+            os.remove(path)
```
mileswwatkins: I feel like this should be an `openstates.utils` function, a la `lxmlize()`. Maybe we should add it.

openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+
+            chamber = "lower" if "house of representatives" in lines[0].lower() else "upper"
+            date_parts = lines[1].strip().split()[-3:]
+            date_str = " ".join(date_parts).title() + " " + lines[2].strip()
+            vote_date = datetime.datetime.strptime(date_str,"%b %d, %Y %I:%M:%S %p")
+
+            for res,val in result_types.items():
+                if res in lines[3].upper():
+                    passed = val
+                    break
```
mileswwatkins: How about an `else` in case no codes are found?

openstates/in/bills.py
rshorey: Ah, I think that's actually legacy code. Sorry. I accidentally rewrote exactly this below. I will consolidate.

openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+
+            motion = " ".join(lines[4].split()[:-2])
+            try:
+                yeas = int(lines[4].split()[-1])
+                nays = int(lines[5].split()[-1])
+                excused = int(lines[6].split()[-1])
+                not_voting = int(lines[7].split()[-1])
+            except ValueError:
+                self.logger.warning("Vote format is weird, skipping")
+                continue
```
mileswwatkins: How often does this come up? Was it just that single weird bill you had talked about with the me the other day?

rshorey: once that I saw.

openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+    def process_votes(self,rollcalls,bill,proxy):
+
+        result_types = {
+            'FAILED': False,
+            'DEFEATED': False,
+            'PREVAILED': True,
+            'PASSED': True,
+            'SUSTAINED': True,
+            'NOT SECONDED': False,
+            'OVERRIDDEN': True,
```
mileswwatkins: Could you explain this one a bit more in a comment? Are we talking about veto override? Objection override? How do we know that this is always going to be a positive for the bill?

openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+                "Yes vote counts ({count}) don't match count of actual votes ({actual})".format(count=vote["yes_count"],actual=len(vote["yes_votes"])))
+            assert (len(vote["no_votes"]) == vote["no_count"],
+                "No vote counts ({count}) don't match count of actual votes ({actual})".format(count=vote["no_count"],actual=len(vote["no_votes"])))
+            assert (len(vote["other_votes"]) == vote["other_count"],
+                "Other vote counts ({count}) don't match count of actual votes ({actual})".format(count=vote["other_count"],actual=len(vote["other_votes"])))
+            
+
+            #vote should have the majority if it passed
+            if vote["passed"] and vote["yes_count"] <= vote["no_count"]:
+                raise AssertionError("The vote should have at least a majority if it passed.")
```
mileswwatkins: Smart

openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+
+            #if vote has the majority and didn't pass, throw a warning
+            #this is possible in supermajority cases,
+            #but we might want to think about it if it starts happening lots
+            if not vote["passed"] and vote["yes_count"] > vote["no_count"]:
+                self.logger.warning("Vote got a majority but didn't pass. Probably worth a look.")
```
mileswwatkins: As long as we're doing these checks, can we do a hard `assert` to make sure that if a bill _does_ have supermajority that it absolutely _does_ pass?

From poking around I don't think that Indiana _has_ anything that requires a supermajority, even veto overrides or constitutional amendments, but I could be wrong (`http://www.in.gov/ipas/2425.htm`).

openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+        docs = {"Amendment":version["amendments"],
+                "Fiscal Note":version["fiscal-notes"],
+                "Committee Amendment":version["cmte_amendments"],
+                "Floor Amendment":version["floor_amendments"],
+                "Committee Report":version["committee-reports"]
+                }
+        for doc_type,doc_list in docs.items():
+            for doc in doc_list:
+                title = "{doc_type}: {name}".format(doc_type=doc_type,name=doc["name"])
+                url = proxy["url"] + doc["link"]
```
mileswwatkins: I'm seeing some repeated URLs. For example, HB 1372 has the following document link twice, once as an `Amendment` and once as a `Committee Amendment`:
`http://in.proxy.openstates.org/2015/bills/hb1372/versions/hb1372.01.intr/amendments/HB1372.01.INTR.AMH01`

I'm not sure what de-duping strategy would work best; your instinct is probably right after spending all this time on Indiana!

openstates/in/bills.py
```diff
@@ -77,9 +81,127 @@ def get_name(self,random_json):
...
+
+        #votes
+        votes = version["rollcalls"]
+        self.process_votes(votes,bill,proxy)
+
+
     def scrape(self, session, chambers):
         
         api_base_url = "https://api.iga.in.gov"
+        proxy = {"url":"http://in.proxy.openstates.org"}
```
mileswwatkins: Maybe add a comment about the proxy, where its code is stored, and why it's needed? Seems somewhat opaque to outsiders at the moment.

mileswwatkins:
I think that's it as far as review. Slick job on handling such an annoying problem with the documents!

openstates/in/bills.py
rshorey: asserts are fixed, as is the error this was masking

openstates/in/bills.py
rshorey: removed old methods that have been incorporated elsewhere.

openstates/in/bills.py
rshorey: In general agreed. I'm going to leave it for now, though and PR it later after I've done some research on how much (ie line splitting or not) we typically want to include so we can get IN done and out.

mileswwatkins: Oh, yeah, totally.

openstates/in/bills.py
rshorey: a little more complicated than an else, but done.

openstates/in/bills.py
rshorey: This was already in the code when I got it and doesn't conflict with anything I've seen.

openstates/in/bills.py
rshorey: comment explains the proxy situation. possibly in way too much detail :)

openstates/in/bills.py
rshorey: ugh, Indiana, your duplicate documents were a pain! but fixed. good catch, @mileswwatkins

rshorey:
ok I think I've addressed all of @mileswwatkins comments (plus a few more things!) Indiana's never going to have the world's cleanest scraper, but I think this is a decently runable, maintainable attempt.

PR #635, opened 2015-03-20
NV: deals with petition bills and a few other tweaks

rshorey:
@mileswwatkins will have a field day that I willingly used a regex

### DIFF
--- a/openstates/nv/bills.py
+++ b/openstates/nv/bills.py
@@ -142,7 +142,7 @@ class NVBillScraper(BillScraper):
     def scrape_assem_bills(self, chamber, insert, session, year):
 
         doc_type = {1: 'bill', 3: 'resolution', 5: 'concurrent resolution',
-                    6: 'joint resolution'}
+                    6: 'joint resolution',9:'initiative petition'}
         for docnum, bill_type in doc_type.iteritems():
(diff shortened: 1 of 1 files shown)


mileswwatkins:
All my regexes live in Texas, that's why I'm in Tennessee.

mileswwatkins:
[`from re import funtimez`](http://funfunfunfunfunfunfunfun.funfunfunfunfunfunfunfun.com/)

openstates/nv/bills.py
```diff
@@ -245,8 +245,18 @@ def scrape_actions(self, root, bill, actor):
                         action_type = atype
                         break
 
+
+                if "Committee on" in action:
+                    committees = re.findall("Committee on ([a-zA-Z, ]*)\.",action)
```
mileswwatkins: What if there's text _after_ the committee name? How do you know when to stop the `re` search?

mileswwatkins:
That's my only comment. Still can't run locally, which is a pain, so I can't look at the data directly, unfortunately.

openstates/nv/bills.py
rshorey: That's what the literal period is for! I checked a bunch and they all had extra text, but the committee was always the end of a sentence.

rshorey:
I did some additional data check since you can't see them, and saw no errors, so I'm good with a merge if you are.

PR #636, opened 2015-03-20
AL: Defend against bad BIR actions from the LIS itself

mileswwatkins:
B-b-b-b-baaaad, bad to the bone.


mileswwatkins:
Basically, in exactly one case (at the moment), the actions of other bills are dumped into the BIR action list elsewhere. It's an internal ALISON issue that we need to mitigate ourselves.

rshorey:
looks good!

PR #644, opened 2015-03-24
OH: Tweak subcommittee name

mileswwatkins:
I think this is all there is to the fix, but as of now the Ohio House website is completely down. So wait until that's back up and test the `legislator` scraper before pulling this change in.


rshorey:
hang on sloopy.

PR #645, opened 2015-03-24
MT: Correct committee PDF parsing, and handle usage of Unicode

mileswwatkins:


openstates/mt/committees.py
```diff
@@ -156,6 +160,109 @@ def is_legislator_name(line):
...
+        and fourth columns are joint comittees that are scraped from
+        the Senate document, so the only column that needs to be
+        inserted this way is the second.
+        '''
+
+        assert "Revised: January 23, 2015" in text,\
+            "House committee list has changed; check that the special-case"\
+            " fix is still necessary, and that its text is still correct"
+
+        second_page_second_column = '''State Administration (cont.)
```
paultag:
```diff
+        second_page_second_column = '''State Administration (cont.)
```

This is also valid:

```diff
+        second_page_second_column = '''\
+State Administration (cont.)
```

openstates/mt/committees.py
```diff
@@ -156,6 +160,109 @@ def is_legislator_name(line):
         if comm['members']:
             self.save_committee(comm)
 
+    def _fix_house_text(self, text):
```
paultag: so I don't super love this.

Why can't we do this on the fly?

mileswwatkins:
@paultag: Here's my attempt at on-the-fly. I _would_ try to do automated column detection, but that would be challenging since after `pdftotext -layout` there isn't a consistent line width or number of spaces between lines; we'd have to calculate the modal character on which the lines began, and use that to guess. Better to just feed the program the text of the header column, in practice, IMO.

paultag:
Also :+1: - I remember reviewing this, forgot to merge. Thanks!

PR #646, opened 2015-03-25
UT: Skip votes if their webpage returns a 404 error

mileswwatkins:
This seems to be a fix for a one-off issue, where [SB 235](http://le.utah.gov/~2015/bills/static/SB0235.html) had a duplicate vote listed in its actions. The first of these had a [valid link to the vote](http://le.utah.gov/DynaBill/svotes.jsp?sessionid=2015GS&voteid=1765&house=S), and the second one had a [dead link](http://le.utah.gov/~2015/status/sbillsta/0). Included a `self.warning()` so that this doesn't proliferate without explicitly notifying us.


rshorey:
looks good!

PR #650, opened 2015-03-27
indiana bill versions now with dates

rshorey:


paultag:
Nice. We should watch for this to fix prod.

rshorey:
so this didn't fix prod, although the dates are now in there. I'm going to reverse the list and see if that works.

PR #651, opened 2015-03-27
TN: not crashing anymore

rshorey:
It seems that TN is kind of fickle in terms of page loading - occasionally we're missing some information from a bill. The committed code is functional and no worse than we were before. Should we be doing some kind of check for incomplete pages, though? Anyone have thoughts on the best way to do that?

I spotchecked the data in openstates, and the ones I checked look complete, so possibly it's just a temporary fluke.


paultag:
Nice. :+1:, thanks!

PR #653, opened 2015-03-30
IN: bill version order reversed

rshorey:


mileswwatkins:
Could you give an example of a current-session `bill` with multiple `version`s?

rshorey:
http://openstates.org/in/bills/2015/SB101/

mileswwatkins:
Otherwise, the code update looks great! Merge it in once you make that one tweak.

rshorey:
so it looks like we've been really inconsistent about this - I grep'd for it, and we've used both as actor relatively frequently. Could not find documentation on one vs the other. @paultag, is "governor" or "executive" preferred for actor when a bill is signed?

rshorey:
confirmed with @paultag that "executive" is preferred here.

PR #656, opened 2015-03-30
Remove last of the urlopen() calls

mileswwatkins:
These two states had open PRs when I was removing `urlopen()`, and as a result I must've planned to add the change directly to their PRs instead of to the main one. Anyway, here're those changes now.


rshorey:
looks good

PR #657, opened 2015-03-30
OR: committee scrapers now getting members

rshorey:


rshorey:
also this resolves data-449 (mostly this is a reminder to myself to close it!)

PR #658, opened 2015-03-31
LA: Remove dependency on BackoffScraper

mileswwatkins:
Trivial change; @rshorey, could you run locally and then merge in?


paultag:
Wait, why? The LA server was kicking me out hard -- in fact, the original backoff code was for LA

paultag:
Also, this seems fine, looks like there was a typo in the class inheritance from eyeballing it, this wouldn't have ever run

PR #660, opened 2015-03-31
FL: Increase precision of allowed vote codes for Senate committee rolls

mileswwatkins:
This only allows "X" (standard vote) and "VA" (vote after roll call) as valid votes; it omits "AV" (abstained from voting) and other codes. The code _does_ confirm that the totals match the head counts, so after a full run I'm confident this change is correct.

### DIFF
--- a/openstates/fl/bills.py
+++ b/openstates/fl/bills.py
@@ -229,7 +229,7 @@ class FLBillScraper(BillScraper, LXMLMixin):
                         ''', line).group(1)
                 # Usually non-voting members won't even have a code listed
                 did_vote = bool(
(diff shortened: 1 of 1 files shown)


paultag:
when you say that it omits `AV`, it'll still add it to `other`, right?

mileswwatkins:
Affirmative:
```python
if did_vote:
    …
else:
    votes['other'].append(member)
```

rshorey:
Ran this locally and didn't see any issues. @paultag, any reason not to merge?

paultag:
fire away

PR #662, opened 2015-03-31
UT: fixed leg scraper

rshorey:
Was not scraping senators; now is using json to scrape both houses (except for a few details needed from rep pages)

### DIFF
--- a/openstates/ut/legislators.py
+++ b/openstates/ut/legislators.py
@@ -1,5 +1,5 @@
 from billy.scrape.legislators import LegislatorScraper, Legislator
-
+from scrapelib import HTTPError
 import lxml.html
 
 from openstates.utils import LXMLMixin
@@ -32,9 +32,19 @@ class UTLegislatorScraper(LegislatorScraper,LXMLMixin):
             else:
                 leg_url = (senate_base_url +
                         "senators/district{dist}.html".format(dist=district))
-                leg = Legislator(term, 'upper', district, leg_name,
-                         party=party, photo_url=photo_url, url=leg_url)
-                leg.add_source(leg_url)
+                try:
+                    self.head(leg_url)
+                except HTTPError:
+                    warning_text = "Bad link for {sen}".format(sen=leg_name)
+                    self.logger.warning(warning_text)
+
+                    leg = Legislator(term, 'upper', district, leg_name,
+                         party=party, photo_url=photo_url)
+                else:
+                    leg = Legislator(term, 'upper', district, leg_name,
+                         party=party, photo_url=photo_url,url=leg_url)
+                    leg.add_source(leg_url)
+
                 address = leg_info["address"]
                 try:
                     cell = leg_info["cell"]
--- a/openstates/ut/legislators.py
+++ b/openstates/ut/legislators.py
@@ -63,6 +63,9 @@ class UTLegislatorScraper(LegislatorScraper,LXMLMixin):
             self.save_legislator(leg)
 
     def scrape_house_member(self, leg_url, leg):
+        #JSON is complete for senators, not for reps
+        #so we still have to hit the rep's page
+        #to get office info
 
         leg_doc = self.lxmlize(leg_url)
         email = leg_doc.xpath('//a[starts-with(@href, "mailto")]')[0].text
--- a/openstates/ut/legislators.py
+++ b/openstates/ut/legislators.py
@@ -61,7 +61,7 @@ class UTLegislatorScraper(LegislatorScraper,LXMLMixin):
                 #them in order of how likely we think they are
                 #to actually get us to the person we care about
                 phone = (cell or home_phone or work_phone)
-                
+
                 email = leg_info["email"]
                 leg.add_office('district', 'Home',
                     address=address, phone=phone, email=email, fax=fax)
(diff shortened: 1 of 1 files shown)


openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
+                    address=address, phone=cell, email=email)
+                conflict_of_interest = (senate_base_url +
+                        "disclosures/2015/{id}.pdf".format(id=leg_id))
+
+                leg["links"] = [conflict_of_interest]
+
+            leg.add_source(json_link)
             self.save_legislator(leg)
 
+    def scrape_house_member(self, leg_url, leg):
```
mileswwatkins: Add a quick comment for posterity, basically what you just told me? Otherwise it's a tad unclear why we're running an extra partial scraper on `lower`s.

openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
-            leg_url = a.get('href')
-            email = tds[2].xpath('span[@class="email"]/a/text()')
-            if email:
-                email = email[0]
-            else:
-                email = ''
+        leg_doc = self.lxmlize(leg_url)
+        email = leg_doc.xpath('//a[starts-with(@href, "mailto")]')[0].text
+        address = leg_doc.xpath('//b[text()="Address:"]')[0].tail.strip()
+        cell = leg_doc.xpath('//b[text()="Cell Phone:"]')
```
mileswwatkins: Looks like some members have `Work Phone`s instead; eg, http://le.utah.gov/house2/detail.jsp?i=POWELK.

openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
-            try:
-                leg_html = self.get(leg_url).text
-                leg_doc = lxml.html.fromstring(leg_html)
-                leg_doc.make_links_absolute(leg_url)
-                photo_url = leg_doc.xpath('//p[@class="photo"]/img/@src')[0]
-            except:
-                self.warning('could not fetch %s' % leg_url)
-                photo_url = ''
+        conflict_of_interest = leg_doc.xpath("//a[contains(@href,'CofI')]/@href")
+        leg["links"] = [conflict_of_interest]
```
mileswwatkins: :clap:

openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
-                             party=party, photo_url=photo_url, email=email,
-                             url=leg_url)
-            leg.add_office('district', 'Home', address=address, phone=cell)
-
-            leg.add_source(url)
-            leg.add_source(leg_url)
+                leg_url = (senate_base_url +
+                        "senators/district{dist}.html".format(dist=district))
+                try:
+                    self.head(leg_url)
```
mileswwatkins: Slick.

openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
+                except HTTPError:
+                    warning_text = "Bad link for {sen}".format(sen=leg_name)
+                    self.logger.warning(warning_text)
+
+                    leg = Legislator(term, 'upper', district, leg_name,
+                         party=party, photo_url=photo_url)
+                else:
+                    leg = Legislator(term, 'upper', district, leg_name,
+                         party=party, photo_url=photo_url,url=leg_url)
```
mileswwatkins: Should we put `leg_url` in both the `url` top-level field _and_ the `sources` list? This might be a question for @paultag, since I'm not sure.

openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
+                    leg.add_source(leg_url)
+
+                address = leg_info["address"]
+                try:
+                    cell = leg_info["cell"]
+                except KeyError:
+                    cell = None
+                email = leg_info["email"]
+                leg.add_office('district', 'Home',
```
mileswwatkins: How do we know that the office _is_ their home, and not a different district office?

openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
+    def scrape(self, term, chambers):
+        #utah seems to have undocumented JSON!
+        house_base_url = "http://le.utah.gov/house2/"
+        senate_base_url = "http://senate.utah.gov/"
+        json_link = "http://le.utah.gov/data/legislators.json"
+        leg_json = self.get(json_link).json()
+        for leg_info in leg_json["legislators"]:
+            leg_name = leg_info["fullName"]
+            district = leg_info["district"]
+            party = leg_info["party"]
```
mileswwatkins: Need to crosswalk from "R" and "D" to full party names.

openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
-            if cell:
-                cell = cell[0].tail.strip()
+    def scrape(self, term, chambers):
+        #utah seems to have undocumented JSON!
+        house_base_url = "http://le.utah.gov/house2/"
+        senate_base_url = "http://senate.utah.gov/"
+        json_link = "http://le.utah.gov/data/legislators.json"
+        leg_json = self.get(json_link).json()
+        for leg_info in leg_json["legislators"]:
+            leg_name = leg_info["fullName"]
```
mileswwatkins: Using `formatName` instead of `fullName` seems a tiny bit better, since it reports legislators' names in a clean {first} {middle} {last} order, instead of {last}, {first} {middle}.

openstates/ut/legislators.py
```diff
@@ -1,125 +1,81 @@
...
+                    leg.add_source(leg_url)
+
+                address = leg_info["address"]
+                try:
+                    cell = leg_info["cell"]
+                except KeyError:
+                    cell = None
+                email = leg_info["email"]
+                leg.add_office('district', 'Home',
+                    address=address, phone=cell, email=email)
```
mileswwatkins: At least some legislators have `fax` in their JSON. There're also `work` and `home` phones, though I'm not sure if we want to prefer those over `cell`.

openstates/ut/legislators.py
rshorey: I agree, but we've been scraping the fullName for a while. If we switch to formatName, will we have to MOM everyone, and is it worth it?

openstates/ut/legislators.py
rshorey: I stole this logic from the previous scraper, but I just now google maps'd a few of these, and I'm pretty confident those are people's houses.

openstates/ut/legislators.py
rshorey: good catch, fixed.

openstates/ut/legislators.py
rshorey: good call, comment commented

openstates/ut/legislators.py
rshorey: ok, added, with comments explaining what we discovered about work phones

openstates/ut/legislators.py
rshorey: phone/fax situation investigated and resolved

openstates/ut/legislators.py
rshorey: confirmed that having it in both places is cool.

rshorey:
fixed the _scraped_name formatting thing, so I think this is good to go.

PR #665, opened 2015-04-01
LA: Remove BackoffScraper dependency in committee scraper as well

mileswwatkins:
Amateur hour over here; can't believe I missed this other import on my other PR.


paultag:
Also broken MRO order I think :)

PR #666, opened 2015-04-01
NM: Remove errant text after filenames

mileswwatkins:


paultag:
:fire: :metal: :skull: :metal: :fire:

PR #668, opened 2015-04-01
MA: Handle newly malformed legislator page

mileswwatkins:


paultag:
:confetti_ball: 

:trophy: to MA for continuing their outstanding tradition of leading the nation in LISs

paultag:
For those playing at home, [the site is actually busted](https://malegislature.gov/People/Profile/jfk0)

PR #670, opened 2015-04-01
TN: Correct Joint Government Ops Committee member names

mileswwatkins:
I still had so much to learn about XPath back in January.

Fixes Jira bug 447:
https://sunlight.atlassian.net/browse/DATA-447


rshorey:
looks good!

PR #671, opened 2015-04-03
NY: Add new event type when parsing listings

mileswwatkins:


rshorey:
@mileswwatkins runs fine locally, is there a specific event I should check to make sure I'm getting the right data?

rshorey:
Also it looks like Summit is misspelled. Is that our mistake or matching theirs? If it's theirs, could you comment that?

mileswwatkins:
@rshorey, the scraper should be failing on [this page](http://assembly.state.ny.us/leg/?sh=hear), since it doesn't look for "Public Forum"s.

rshorey:
OK, looks good on both fronts now, I'll merge.

PR #672, opened 2015-04-06
IN: working around bad API response

rshorey:
looks like it was confined to one page, can't access it in the browser either.

While I was in there, threw a .lower() on the governor check just in case.


paultag:
:+1: thanks!

PR #676, opened 2015-04-07
TN: Allow date-only event times

mileswwatkins:


openstates/tn/events.py
```diff
@@ -95,17 +95,24 @@ def scrape(self, chamber, session):
                 description = metainf['type'].text_content()
 
                 dtfmt = "%A, %B %d, %Y %I:%M %p"
+                dtfmt_no_time = "%A, %B %d, %Y"
                 if time == 'Cancelled':
                     self.log("Skipping cancelled event.")
                     continue
                 else:
+                    print(datetime_string)
```
rshorey: remove print statement?

paultag: :+1:

mileswwatkins: Heh.

rshorey:
probably a pre-existing issue, but it looks like for documents name and url are reversed. For example I'm getting:

    "documents": [
        {
            "name": "http://wapp.capitol.tn.gov/apps/Calendar/CalendarMain.aspx?committeeid=980000&calendardate=4/7/2015&ga=109",
            "type": "agenda",
            "url": "Agenda"
        }
    ]

paultag:
we should fix that up, nice catch!

mileswwatkins:
Alright, changed from positional to named parameters!

rshorey:
looks good, thanks!

PR #687, opened 2015-04-13
WV: workaround for missing bill titles

rshorey:
For one bill there seems to be no title anywhere (although all other information incl bill text is good and valid) so I'm just using the ID as the title in this case. @paultag, is this OK? Super non-urgent.


paultag:
Hurm. Interesting question. I *think* this may be OK. Let me play with it
quickly.

paultag:
OK, after looking into this, it seems sane.

[one example without a title](http://www.legis.state.wv.us/Bill_Status/Resolution_History.cfm?year=2015&sessiontype=RS&input4=37&billtype=cr&houseorig=s&btype=res)

PR #695, opened 2015-04-17
TX: Allow additional datetime strings

mileswwatkins:


rshorey:
texas, why do you have to add new text every day!!!

(looks good, merging)

PR #696, opened 2015-04-17
OH: Quicken API calls

mileswwatkins:
I think that we've been hitting a timeout issue on big API calls, so I decreased the `per_page` from 300 to the default (50(?)).


rshorey:
looks good to me!

PR #700, opened 2015-04-21
OK: empty districts and whitespace issues

rshorey:
moved the check for a bad district earlier so we don't try to do stuff with an empty dist!


paultag:
:+1:, thank you!

PR #703, opened 2015-04-22
Fix OCD division ID mappings

mileswwatkins:


paultag:
Awesome. Thanks, @mileswwatkins !

PR #705, opened 2015-04-23
FL: Accept changed votes in committee vote parsing

mileswwatkins:


rshorey:
looks reasonable to me!

PR #707, opened 2015-04-27
de: throw warning instead of failing on missing vote docs

rshorey:
my new favorite vote passage status is "Neither passed nor defeated by Senate". Thanks Delaware.


mileswwatkins:
Works great, but seems a little unclear on reading. Could you instead try a `HEAD` and check for `<404>` on the document page?

rshorey:
@mileswwatkins pushed a slightly different solution, since scrapelib's "head" throws an error on 404 and I think resetting accept_headers is both more confusing and likely to cause unexpected issues down the line.

mileswwatkins:
:smiley_cat:

PR #708, opened 2015-04-27
MA: Fix bill text scraping

mileswwatkins:
As reported on [Open States Google Group](https://groups.google.com/d/msg/fifty-state-project/hI1CwlNLsWA/1NDpBUcLxccJ). Thanks!


paultag:
:+1: nice!

PR #710, opened 2015-04-28
DC: fix actors on committees

rshorey:


paultag:
:+1: thank you!

mileswwatkins:
This is coming up with some weird actors for me. When I `grep` for "legislator" inside the `bills` JSON, most of the legislator names I get are "the Mayor" (wouldn't that be "executive" instead?). I also get one "legislator" named "the University of the District of Columbia" (for bill `21_upper_CA21-0010`).

paultag:
@mileswwatkins was that present before this changeset too? We might need to do another PR

mileswwatkins:
I'm not sure, but there's a good chance that's the case.

mileswwatkins:
I'll file a Jira bug.

paultag:
:heart: - let's get another chanceset addressing issues

PR #711, opened 2015-04-28
DC: correct actor when mayor or committee withdraws a bill

rshorey:
this doesn't fix the University of District of Columbia as a legislator who withdraws a bill - the only way I can think of to do that is to hard-code them in. @paultag how big a deal is it if we have them as a (unresolved) related_entity of type legislator in one bill?


paultag:
Not bad at all. No worries

PR #714, opened 2015-05-01
GA: Order bills

paultag:
I think this solves it. untested.


paultag:
The issue here is the "latest text" link is pointing to some random bill. Let's sort by their ID, that should be good enough for now

PR #717, opened 2015-05-05
OH: Fix bill source URLs

mileswwatkins:


mileswwatkins:
The Ohio API's endpoint is `/bills/`, not `/bill/`.

rshorey:
How about resolutions? Does this also hold? If so, maybe we should just clean up the code that removes the 's' directly above the line you changed.

mileswwatkins:
It _does_ hold for resolutions as well. Does this code strike your fancy?

rshorey:
yes!

PR #719, opened 2015-05-06
IN: switched from https to http to (hopefully) fix SSL error

rshorey:
getting a SSL error on production but not locally. Indiana works with http and https when visiting through the browser, so figured I'd try http to see if that fixes it.


mileswwatkins:
Worked fine for me, we'll see how production feels.

PR #720, opened 2015-05-11
WI: Adapt to new committee index pages

mileswwatkins:


rshorey:
looks good, runs fine!

PR #722, opened 2015-05-11
IA: Allow motion names without question marks

mileswwatkins:
Even though it says `On the question of`!!


mileswwatkins:
grinds my gears

mileswwatkins:
@rshorey, all's good now. Thanks for prodding me on this!

rshorey:
ok, sounds good, merging!

PR #725, opened 2015-05-12
Changed 'Open State' to 'Open States'

joemcl:
Changed 'Open State' to 'Open States'


jamesturk:
ha, that's an old one (at one point the name was Open State Project!), thanks for catching it

joemcl:
James, you're welcome

PR #730, opened 2015-05-18
Add script to check for name duplicates that will break OS-OCD conversions

mileswwatkins:
There is already a potential-duplicate script, but this is different enough that I figured I should give it its own file.


paultag:
totes legit, thanks!

PR #733, opened 2015-05-21
MN: added 2015 special session

rshorey:
note that the session doesn't have any bills yet.

Also we should rerun 2015-2016 one more time, because they appear to have added a few more actions after they put this special session in the session list.


mileswwatkins:
No bills yet in this session, so let's keep an eye on it to make sure it goes green soon.

PR #736, opened 2015-05-22
FL: Added new session, which is currently in progress

mileswwatkins:
Also clean up titles from sponsor names.

### DIFF
--- a/openstates/fl/__init__.py
+++ b/openstates/fl/__init__.py
@@ -23,7 +23,7 @@ metadata = dict(
          'sessions': ['2013', '2014', '2014A'],
          'start_year': 2013, 'end_year': 2014},
         {'name': '2015-2016',
-         'sessions': ['2015'],
+         'sessions': ['2015', '2015A'],
          'start_year': 2015, 'end_year': 2016}
     ],
     session_details={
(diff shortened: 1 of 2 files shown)


rshorey:
When I run this locally I get:

AssertionError: Bill search pagination needed but not used

mileswwatkins:
I added that `warning` just now. Basically, I want it to highlight the fact that the pagination isn't being used in the bill search. (Specifically, that the "next" button isn't found.) It's innocuous right now because there _are_ genuinely few bills, but I think it's more of a `warning` than a `debug`. Good way to keep the code fragile.

mileswwatkins:
Oh, er, uh, I'll look into that. Misread :)

rshorey:
Indeed when I downloaded the newer version of libxml2 it worked.

rshorey:
OK, now that I've got it working...we're not scraping any bills from the house. Looks like they're not even acknowledging the special session yet. Is that of any concern?

mileswwatkins:
Both the House and the Senate are in on it, but AFAICT only the Senate has bills filed in it so far.

rshorey:
OK, all looks good.

PR #739, opened 2015-05-26
TX: Refactor to just remove all text after datetime

mileswwatkins:
Dang tootin' datetime event text has annoyed us for the last time. I just took @rshorey's advice and removed our explicit handling, as it was doing far more harm than good; we were just removing all the text anyways. :fire: #offwithitshead :fire:


paultag:
:+1:

PR #740, opened 2015-05-26
MN: senate changed the format of their legislators csv

rshorey:


mileswwatkins:
All Senators have their capitol office listed on the index page, which might be worth capturing in addition to (or instead of) from the spreadsheet, which only lists district _or_ capitol. (I think the one address that's listed on the spreadsheet is their preferred mailing address during the recess.) Also, there's often additional contact information (phone numbers, additional addresses) on individual member pages. If you think this is worth the effort, either implement it or (more likely) create a Jira issue to add this later.

mileswwatkins:
The change looks good, though, so I'm pulling it in.

PR #744, opened 2015-05-29
Updated to show business phone if there isn't a home phone available.

linzjax:


rshorey:
Change looks good! @linzjax, could you get rid of that extra print statement in line 69 while you're in there, though (you didn't add it, but might as well clean it up)? When you've done it, just make a commit to the same branch and push it, and then make a comment here so I know you've made the change, and I'll pull it in!

mileswwatkins:
Could you clean out some of the excess whitespace? Also, it looks like there was an existing `print` in the scraper; could you remove that?

Otherwise, works great for me!

rshorey:
man, team openstates on the same page today :)

mileswwatkins:
:trophy:

linzjax:
@rshorey Done! Today I learned about linters and branches. It was a good day.

rshorey:
Great! thanks.

PR #745, opened 2015-05-30
RI: Updated xpath for phone number collection

linzjax:
Not sure if this is the cleanest way, but all the other information is being pull from a spreadsheet that didn't contain phone numbers, and individual bio pages weren't guaranteed to contain phone numbers.

### DIFF
--- a/openstates/ri/legislators.py
+++ b/openstates/ri/legislators.py
@@ -96,7 +96,7 @@ class RILegislatorScraper(LegislatorScraper, LXMLMixin):
                 contact_url = 'http://webserver.rilin.state.ri.us/Email/SenEmailListDistrict.asp'            
             if chamber == 'lower':
                 contact_url = 'http://webserver.rilin.state.ri.us/Email/RepEmailListDistrict.asp'
-            contact = lxml.html.fromstring(self.get(contact_url).text)
+            contact = self.lxmlize(contact_url)
             contact_phone = contact.xpath('//tr[@valign="TOP"]//td[@class="bodyCopy"]/text()')
 
             for el in contact_phone:
--- a/openstates/ri/legislators.py
+++ b/openstates/ri/legislators.py
@@ -102,8 +102,8 @@ class RILegislatorScraper(LegislatorScraper, LXMLMixin):
             for el in contact_phone:
                 if full_name in el:
                     number = contact_phone.index(el)
-                    d['phone'] = contact_phone[number + 1]
-                    d['phone'] = d['phone'].strip()
+                    phone = contact_phone[number + 1]
+                    phone = phone.strip()
 
             if d['email'] is not None:
                 kwargs['email'] = d['email']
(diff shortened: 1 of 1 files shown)


openstates/ri/legislators.py
```diff
@@ -98,6 +92,19 @@ def scrape(self, chamber, term):
                 "town_represented": d['town_represented'],
             }
 
+            if chamber == 'upper':
+                contact_url = 'http://webserver.rilin.state.ri.us/Email/SenEmailListDistrict.asp'            
+            if chamber == 'lower':
+                contact_url = 'http://webserver.rilin.state.ri.us/Email/RepEmailListDistrict.asp'
+            contact = lxml.html.fromstring(self.get(contact_url).text)
```
rshorey: self.lxmlize can be used instead here. It's a function we wrote (you can see the code here if you're interested: https://github.com/sunlightlabs/openstates/blob/master/openstates/utils/lxmlize.py)

rshorey:
It looks like the phone number is being added to both the legislator's office (which we want) and as a top-level field in the legislator's JSON (which we don't). Can you make it stop being added to the top level, please?

openstates/ri/legislators.py
```diff
@@ -98,6 +92,19 @@ def scrape(self, chamber, term):
                 "town_represented": d['town_represented'],
             }
 
+            if chamber == 'upper':
+                contact_url = 'http://webserver.rilin.state.ri.us/Email/SenEmailListDistrict.asp'            
+            if chamber == 'lower':
+                contact_url = 'http://webserver.rilin.state.ri.us/Email/RepEmailListDistrict.asp'
```
rshorey: Can you also add the contact URL to the source for the legislator please?

rshorey:
A few legislators' phone numbers seem to be missing. For example, Anastasia Williams' number is available here: http://webserver.rilin.state.ri.us/Email/RepEmailListDistrict.asp but isn't being scraped. I think it may have to do with her having a middle initial in one source but not the other. Maybe it would make sense to match on district instead of first name on the phone list instead (although who knows what problems that creates). I'm OK with missing a few if there's no good way to do it, just wanted to check.

linzjax:
Yeah I noticed a few homepage urls aren't being pull because of differences in how names appear "Cynthia Armour Coyne" vs "Cynthia A. Coyne".
It's also not pulling Da Ponte's number because he's listed as DaPonte on the contact sheet.

I'll poke around for a bit.

rshorey:
this is a super irritating problem. I hate when state legislatures make us match two different pages to each other based on inconsistent names or something. Let us know what you come up with, thanks!

linzjax:
homepage_url feels a little sloppy, but William W. O'Brien was the only one not working with this new method because of a typo in his name : "William W.O'Brien", which is why he has his own code.

linzjax:
Ugh... Frank Lombardo III... hang on...

mileswwatkins:
Lil' Rhody!!

linzjax:
Hah! Take that Frank Lombardo III....

rshorey:
Great job with the Frank-slaying! One more thing - this one's a tiny bit nitpicky (sorry), but since phone is no longer on the top level, we've got a bit of unused code we should remove to prevent later confusion. In line 55, we can make the whole thing an empty dictionary:

d = {}

and we can get rid of the check to see if phone exists (lines 122-123). After that, ready to go!

mileswwatkins:
I was about to say that she could refer to the OS API documents, but they're apparently they don't note which fields are deprecated and which aren't. Still, a good source of information otherwise:
https://sunlightlabs.github.io/openstates-api/

linzjax:
Done! Let me know if there's anything else I should update.

rshorey:
Looks good!

PR #747, opened 2015-06-02
AR: Added Independent Party option

linzjax:


paultag:
Sweet, thanks @linzjax !

PR #749, opened 2015-06-02
TN: deal with vacant seat

rshorey:


mileswwatkins:
https://www.youtube.com/watch?v=FT25niTg2IM

mileswwatkins:
Output looks good to me. Any idea why every legislator is hitting the website twice?

PR #752, opened 2015-06-03
DE: bad de link

rshorey:
one bad link. surprisingly painful to fix. delaware.

### DIFF
--- a/openstates/de/bills.py
+++ b/openstates/de/bills.py
@@ -9,7 +9,6 @@ from billy.scrape.bills import BillScraper, Bill
 from billy.scrape.votes import Vote
 import scrapelib
 import actions
-import requests
 
 class DEBillScraper(BillScraper, LXMLMixin):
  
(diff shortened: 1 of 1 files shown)


paultag:
Doesn't this add 2x overhead to the normal scrape? Since you have to HEAD and GET every page?

openstates/de/bills.py
```diff
@@ -50,8 +50,13 @@ def scrape_bill(self,link,chamber,session):
 
         base_url = "http://legis.delaware.gov"
         text_base_url = "http://legis.delaware.gov/LIS/lis{session}.nsf/vwLegislation/{bill_id}/$file/legis.html?open"
-        page = self.lxmlize(link)
+
```
paultag: mind triming whitespace out?

rshorey:
@paultag here's another way that doesn't increase overhead but puts lxmlize in a try which I don't love, but is probably better.

paultag:
Yeah, slightly more ugly, but would reduce network overhead by half, so I'm cool with it :)

paultag:
Thanks!

PR #754, opened 2015-06-04
IN: added new names for ignored sessions

rshorey:
Indiana apparently changed the names of some old sessions. I just dumped the new names in ignored_sessions with the old names. @paultag is this an acceptable way of dealing with them?


paultag:
I think the way that file is now, this is right - but let's change how that file is

Let's throw them into the `_scraped_name` in the session dict, I think this predates that bit

rshorey:
@paultag moved them to session_details

paultag:
#swag

PR #756, opened 2015-06-04
MD: Fixed indent so scraper saved joint committees as well

linzjax:
Side note: I think this might have added all "House Delegation" committees to 'Joint'. If that's wrong, I'll fix it.

### DIFF
--- a/openstates/md/committees.py
+++ b/openstates/md/committees.py
@@ -27,14 +27,12 @@ class MDCommitteeScraper(CommitteeScraper):
             com_name = com_name.replace("Committee on ", "")
             com_name = com_name.replace(" Committee", "")
             com_name = com_name.strip()
-            print com_name
             if 'Senate' in chamber_name:
                 chamber = 'upper'
             elif 'House' in chamber_name and 'Delegation' not in chamber_name:
                 chamber = 'lower'
             elif 'Joint' or 'Statutory' or 'Special Joint' or 'Other' in chamber_name:
                 chamber = 'joint'
-                print chamber
             else:
                 self.logger.warning("No committee chamber available for committee '%s'" % com_name)
                 continue
@@ -61,13 +59,13 @@ class MDCommitteeScraper(CommitteeScraper):
             for row in rows[1:]:
                 name = row.getchildren()[0].text_content().strip()
                 if name.endswith(' (Chair)'):
-                    name = name.replace(' (Chair)','')
+                    name = name.replace(' (Chair)', '')
                     role = 'chair'
                 elif name.endswith(' (Vice Chair)'):
-                    name = name.replace(' (Vice Chair)','')
+                    name = name.replace(' (Vice Chair)', '')
                     role = 'vice chair'
                 elif name.endswith(' (Co-Chair)'):
-                    name = name.replace(' (Co-Chair)','')
+                    name = name.replace(' (Co-Chair)', '')
                     role = 'co-chair'
                 else:
                     role = 'member'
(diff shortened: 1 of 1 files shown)


openstates/md/committees.py
```diff
@@ -14,25 +14,33 @@ def scrape(self, term, chambers):
...
+                url = url.replace('stab=01', 'stab=04')
             com_name = a.text
             if com_name is None:
                 continue
+            com_name = com_name.replace("Joint ", "")
+            com_name = com_name.replace("Committee on ", "")
+            com_name = com_name.replace(" Committee", "")
             com_name = com_name.strip()
-            if 'Senate' in chamber and 'upper' in chambers:
+            print com_name
```
mileswwatkins: Whoops! Your debug `print` is showing.

linzjax:
Ew gross

openstates/md/committees.py
```diff
@@ -14,25 +14,33 @@ def scrape(self, term, chambers):
...
+            if chamber_name == 'Senate Standing':
+                url = url.replace('stab=01', 'stab=04')
+            if chamber_name == 'House Standing':
+                url = url.replace('stab=01', 'stab=04')
             com_name = a.text
             if com_name is None:
                 continue
+            com_name = com_name.replace("Joint ", "")
+            com_name = com_name.replace("Committee on ", "")
+            com_name = com_name.replace(" Committee", "")
```
mileswwatkins: You may want to guarantee that these are where you think they are (ie, make sure that you're not removing "Joint" from the middle of a name, or similar). For example, make the replacements more precise by using regular expressions:
```
import re
…
# The carrot indicates the start of a line in regular expressions
com_name = re.sub("^Joint ", "", com_name)
…
```
or by adding conditionals:
```
if com_name.startswith("Joint "):
    # The last argument means that only the first instance of "Joint " will be replaced
    com_name = com_name.replace("Joint ", "", 1)
```

linzjax:
fixed

openstates/md/committees.py
```diff
@@ -14,17 +14,24 @@ def scrape(self, term, chambers):
...
             com_name = com_name.strip()
-            if 'Senate' in chamber and 'upper' in chambers:
+            if 'Senate' in chamber_name:
                 chamber = 'upper'
-            elif 'House' in chamber and 'Delegation' not in chamber and 'lower' in chambers:
+            elif 'House' in chamber_name and 'Delegation' not in chamber_name:
                 chamber = 'lower'
-            elif chamber in ('Joint', 'Statutory', 'Special Joint', 'Other'):
+            elif 'Joint' or 'Statutory' or 'Special Joint' or 'Other' in chamber_name:
```
mileswwatkins: This may be at the root of your problem; it'll always resolve to true, since everything between each `or` is its own boolean. `'Joint'` will resolve to `True`, for example, because it's a non-empty string. I've gotten bit by this one before!
```
if 'foo' or 'bar' in 'baz':
   print("This doesn't make sense!")
```

mileswwatkins:
Looks like we're missing membership listings in at least some `joint` committees. For example, look at the `data/md/committees` JSON for Joint Spending Affordability.

openstates/md/committees.py
mileswwatkins: Might also be worth uppercasing the first letter of the committee name after these replacements, in case it was a preposition and used to be lower-case.

linzjax:
Hm just saw this. hang on

linzjax:
Actually I'm going to try and clean this up, hang out.

rshorey:
@linzjax am I good to review this?

linzjax:
Yup! Still feels a bit messy, so if you have suggestions I'm open to them.

rshorey:
it's messy, but so is the site, so we'll go with it. thanks, this is great work!

PR #757, opened 2015-06-08
WV: added 2016 session

rshorey:
2016 has no bills yet. Should we pull this in anyway? should we turn it off?


paultag:
Yeah, we should likely turn the scraper off, is it a moxie job? Is 2015 for sure over?

rshorey:
As discussed offline, I'm pulling this in but changing the scraper to continue to scrape 2015. The 2015 session is over, but they still have interim committee meetings until the next session starts which we want to catch.

PR #761, opened 2015-06-17
IN: fixed packet error crashing, maybe?

rshorey:
I have never reproduced this error locally, but this ought to work?

### DIFF
--- a/openstates/in/apiclient.py
+++ b/openstates/in/apiclient.py
@@ -108,7 +108,10 @@ class ApiClient(object):
         while resp is None and tries <  num_bad_packets_allowed:
             try:
                 resp = self.scraper.get(url, *requests_args, **requests_kwargs)
-            except SysCallError:
+            except SysCallError as e:
+                err, string = e.args
+                if err != 104:
+                    raise
                 tries += 1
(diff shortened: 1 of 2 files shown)


paultag:
Hurm. I think a SyscallError might capture too much. Was that the thing raised (if it was, we might just have to deal with it :) )

rshorey:
File "/usr/lib/python2.7/dist-packages/OpenSSL/SSL.py", line 862, in _raise_ssl_error
    raise SysCallError(errno, errorcode[errno])
OpenSSL.SSL.SysCallError: (104, 'ECONNRESET')

Is there a way to specify the arguments in a python exception? Otherwise SysCallError is what we've got.

is

paultag:
Yeah, try:

```python
try:
    .....
except Thing as e:
    err, string = e.args
    if err != 104:
        raise
    # (error handling here)
```

paultag:
I just checked in the SSL code, it's a plain subclass, looks like that should work.

rshorey:
ok @paultag I think this should take care of it.

paultag:
:+1: thank you!

PR #763, opened 2015-06-17
LA: added ignored org session; fixed committee bad links

rshorey:


paultag:
Cool; looks good at a glace. Do we need to update Moxie first?

rshorey:
negative, we want to ignore that session anyway bc it's an organizational session which we've always ignored in the past.

PR #766, opened 2015-06-22
OH: Correct order of actions

mileswwatkins:
Closing https://sunlight.atlassian.net/browse/DATA-474 and https://sunlightfoundation.supportbee.com/tickets/5399875

### DIFF
--- a/openstates/oh/bills.py
+++ b/openstates/oh/bills.py
@@ -1,5 +1,6 @@
 import os
 import datetime
+from operator import itemgetter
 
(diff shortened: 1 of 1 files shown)


openstates/oh/bills.py
```diff
@@ -1,5 +1,6 @@
 import os
 import datetime
+from operator import itemgetter
```
paultag: Hah, I didn't know about this, I usually write `lambda x: x['date']`, but this is fun. Yay `operator`!

PR #768, opened 2015-06-24
AL: Ignore the 2016 session for now

mileswwatkins:
Wait on @jamesturk's second opinion before pulling this in, but for now 2016's `do postback` doesn't function properly and so that session has to be ignored.


mileswwatkins:
Neo-sloss @paultag, I'm fine with this code if you want to pull it in.

paultag:
:+1:

PR #772, opened 2015-06-25
ND: Remove titles from legislators names

mileswwatkins:


mileswwatkins:
We'll also need to MOM every single one of these. *sigh*

rshorey:
mom is the worst.

PR #774, opened 2015-06-25
CA: Add 2015 special session

mileswwatkins:
Don't pull this in yet! @rshorey is going to run it locally on her machine; I'm not sure that the `download.py` and then `billy-update ca --bills` are working properly for the new session.

### DIFF
--- a/openstates/ca/__init__.py
+++ b/openstates/ca/__init__.py
@@ -46,7 +46,7 @@ metadata = dict(
          # 'start_date': datetime.date(2013, ??, ?),
          },
          {'name': '20152016',
-         'sessions': ['20152016',],
+         'sessions': ['20152016', '20152016 Special Session 1'],
          'start_year': 2015, 'end_year': 2016,
          },
         ],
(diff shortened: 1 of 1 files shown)


rshorey:
there are no bills in the 2016 session. should we ignore it for now? put it in but run with --session=2015?

mileswwatkins:
Hmm, yeah. I'm confused about this in general, since if you run with `--session "20132014 Special Session 1"` you also get the "no bills returned" error, even though that appears to be a legitimate `session` in the `__init__.py` and has 10 bills in the production database. @paultag, what am I missing?

paultag:
Hurm. Cache?

mileswwatkins:
Cali's the one where all the bills come from the MySQL database dump, so maybe it just contains bills for the current session.

mileswwatkins:
@rshorey, I keep having timeout issues updating the California DB on my machine. Could you take this PR over from here? Should be pretty basic; can't believe it's been this much of a pain with my MySQL setup =/

mileswwatkins:
There _are_ extraordinary session bills; they just have weird names (like `SBX1 4` for "Senate Bill 4 from Extraordinary Session 1"). Not sure if this clarifies anything. Eg:
- http://www.legislature.ca.gov/cgi-bin/port-postquery?bill_number=sbx1_2&sess=CUR&house=B&author=huff_%3Chuff%3E
- http://www.legislature.ca.gov/cgi-bin/port-postquery?bill_number=sbx2_5&sess=CUR&house=B&author=leno_%3Cleno%3E
- http://www.legislature.ca.gov/cgi-bin/port-postquery?bill_number=abx1_2&sess=CUR&house=B&author=perea_%3Cperea%3E

rshorey:
replaced by #795

PR #778, opened 2015-07-06
MD: Stop truncating certain email addresses

mileswwatkins:


rshorey:
There are some incorrect e-mails stuck in top-level fields, so once this has run, we'll need to clear those out, too.

mileswwatkins:
Cleared bad emails.

PR #779, opened 2015-07-06
ND: Add 2017 session, commented out

mileswwatkins:


rshorey:
@mileswwatkins did you file a bug to remove this from ignored someday?

PR #782, opened 2015-07-08
OR: added 2015-2016 interim to ignored

rshorey:
we seem to be ignoring past years' interim sessions, so I put this one in ignored, too.


mileswwatkins:
Yeah, doesn't look like past Interims have had any content, eg:
https://olis.leg.state.or.us/liz/2013I1/Measures/list/

PR #785, opened 2015-07-14
AL: Add currently-ongoing special session

mileswwatkins:


mileswwatkins:
Spot-checked some of the data, and it looks like everything's copacetic.

PR #786, opened 2015-07-15
Scripts: Fix preload-mongo for OSX systems

mileswwatkins:


mileswwatkins:
The issue was that `mktemp` does exist in OSX, but its syntax doesn't line up with Linux's.

PR #787, opened 2015-07-15
NY: Remove spaces from party names, and disallow null parties

mileswwatkins:


rshorey:
what happens if party is empty string? does this fail in billy validations?

mileswwatkins:
Double-checking now.

paultag:
blank *should* fail

paultag:
Yeah - https://github.com/sunlightlabs/billy/blob/master/billy/schemas/person.json#L20 - blank isn't set to `True`

mileswwatkins:
Confirmed!
`14:21:37 WARNING billy: Failed to validate field 'roles' list schema: Value '' for field 'party' cannot be blank'`

PR #790, opened 2015-07-20
CO: fix misparsing of house votes on senate bills

rshorey:
we were not detecting senate bills as new bills. now we are.

doing some research into whether we'll need to hand-delete the bad bills.

### DIFF
--- a/openstates/co/votes.py
+++ b/openstates/co/votes.py
@@ -69,7 +69,7 @@ class COVoteScraper(VoteScraper, LXMLMixin):
                         continue
 
                 found = re.findall(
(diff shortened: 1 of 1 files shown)


rshorey:
tested locally, correctly imports votes. We'll need to delete the bad votes by hand, though, and they're not going to be easy to find...

rshorey:
sounds terrifying, but probably a good call. let's make sure we've got them backed up first.

mileswwatkins:
We can test it locally and with the UI over a couple days to make sure. Agreed it's scary, haha. :ghost:

rshorey:
OK, I'm pulling these changes in. Timball's looking at how often votes are accessed via ID and we can decide how to do the deletes based on that.

PR #791, opened 2015-07-20
KY: getting budget subcommittees for both chambers

rshorey:
this maybe isn't the purest way to do this, but it works.

### DIFF
--- a/openstates/ky/committees.py
+++ b/openstates/ky/committees.py
@@ -66,6 +66,8 @@ class KYCommitteeScraper(CommitteeScraper):
 
         if comm['members']:
             self.save_committee(comm)
+        else:
+            self.logger.warning("Empty committee, skipping.")
         
         #deal with subcommittees
         if parent_comm is None:
--- a/openstates/ky/committees.py
+++ b/openstates/ky/committees.py
@@ -12,13 +12,11 @@ class KYCommitteeScraper(CommitteeScraper):
     def scrape(self, chamber, term):
 
         if chamber == 'upper':
-            urls = ["http://www.lrc.ky.gov/committee/standing_senate.htm",
-                    "http://www.lrc.ky.gov/committee/standing/A&R(S)/home.htm"]
+            urls = ["http://www.lrc.ky.gov/committee/standing_senate.htm"]
             # also invoke joint scraper
             self.scrape('joint', term)
         elif chamber == 'lower':
-            urls = ["http://www.lrc.ky.gov/committee/standing_house.htm",
-                    "http://www.lrc.ky.gov/committee/standing/A&R(H)/home.htm"]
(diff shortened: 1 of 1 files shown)


rshorey:
don't pull this in yet, just found a duplicated committee

rshorey:
OK @mileswwatkins this is ready to go, I think probably worth a quick review when you get around to it since it ended up being more than a one-line change, but no hurry at all

openstates/ky/committees.py
```diff
@@ -41,14 +41,43 @@ def scrape(self, chamber, term):
...
-                    self.save_committee(comm)
+                self.scrape_committee(chamber, link)
+                
+
+    def scrape_committee(self, chamber, link, parent_comm=None):
+        home_link = link.attrib["href"]
+        name = re.sub(r'\s+\((H|S)\)$', '', link.text).strip().title()
+        name = name.replace(".", "")
+        if "Subcommittee " in name:
+            subcomm = name.split("Subcommittee")[1]
```
mileswwatkins: Looks like there're one or two subcommittees that don't _start_ with `Subcommittee`, such as the `joint` `"Energy Special Subcommittee"`. And `"Administration Regulation Review Sub"` is truncated, seemingly for a similar reason. `"Tobacco Settlement Agreement Fund Oversight Committee"` also needs that word stripped from the end.

openstates/ky/committees.py
```diff
@@ -41,14 +41,43 @@ def scrape(self, chamber, term):
...
+        else:
+            comm = Committee(chamber, name)
+        comm.add_source(home_link)
+        comm_url = home_link.replace(
+            'home.htm', 'members.htm')
+        self.scrape_members(comm, comm_url)
+        
+
+        if comm['members']:
+            self.save_committee(comm)
```
mileswwatkins: Otherwise raise a warning or an error? Or are empty committees commonplace?

openstates/ky/committees.py
rshorey: I'm removing the subcommittee part, but it looks like those aren't actual real subcommittees - they don't seem to report to anyone. So Sub. or Subcommittee or Comm. or whatever are gone, but they aren't listed as anyone's subcommittee. I think at least one of them reports to an executive department instead. Seem reasonable?

mileswwatkins: Huh! Crazy. Well, we mirror the LIS, and if the LIS says they don't explicitly report to anyone then I think you're right that they're a top-level `committee` :D

rshorey:
now there's a warning for empty committees. it looks like those aren't actually empty committees but rather 1) committees filled with citizens, not legislators, so we can't classify them in openstates and 2) links that are otherwise indistinguishable from committee links but aren't actually committees, but the warning will help in the future. thanks!

PR #793, opened 2015-07-23
CT: Handle now-header-less legislator spreadsheet, and capture district offices

mileswwatkins:


rshorey:
looks good, ran fine locally, merging!

PR #794, opened 2015-07-24
AR: inserted David Johnson exception

linzjax:
He didn't have a party tag next to his name. Created an exception that would break if he added one later.

### DIFF
--- a/openstates/ar/legislators.py
+++ b/openstates/ar/legislators.py
@@ -53,7 +53,6 @@ class ARLegislatorScraper(LegislatorScraper):
 
         party = name_and_party[-1]
 
-        # David Johnson special casing:
         if ' '.join(name_and_party[1:]) == 'David Johnson' and "(" not in party:
             full_name = ' '.join(name_and_party[1:])
             party = 'Democratic'
--- a/openstates/ar/legislators.py
+++ b/openstates/ar/legislators.py
@@ -5,6 +5,7 @@ from billy.scrape import NoDataForPeriod
 from billy.scrape.legislators import LegislatorScraper, Legislator
 
 import lxml.html
+_Remove_special_casing = True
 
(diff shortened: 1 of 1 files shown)


openstates/ar/legislators.py
```diff
@@ -48,7 +48,13 @@ def scrape_member(self, chamber, term, member_url):
 
         party = name_and_party[-1]
 
-        if party == '(R)':
+        if ' '.join(name_and_party[1:]) == 'David Johnson' and "(" not in party:
+                full_name = ' '.join(name_and_party[1:])
+                party = 'Democratic'
+        elif ' '.join(name_and_party[1:]) == 'David Johnson' and "(" in party:
```
mileswwatkins: If a party _does_ come back, then `' '.join(name_and_party[1:]) == 'David Johnson'` will be false, right?

mileswwatkins:
Also, can you add code to make sure that an error will be thrown if David Johnson is no longer in the legislature? This way the obsolete code can be thrown out when it's no longer needed.

mileswwatkins:
Otherwise, looks great to me!

linzjax:
Alright, seems to be working.

openstates/ar/legislators.py
```diff
@@ -48,7 +53,13 @@ def scrape_member(self, chamber, term, member_url):
 
         party = name_and_party[-1]
 
-        if party == '(R)':
+        if ' '.join(name_and_party[1:]) == 'David Johnson' and "(" not in party:
+            full_name = ' '.join(name_and_party[1:])
+            party = 'Democratic'
+            global _Remove_special_casing
```
mileswwatkins: You don't need to declare a `global` variable; it already exists since you declared it on line 8.

PR #795, opened 2015-07-24
CA: fix special sessions

rshorey:
this replaces @mileswwatkins' PR, which I will close.


mileswwatkins:
And the bill JSON looks good? If so, pull 'er in!

PR #796, opened 2015-07-24
CO: now catches jpgs and jpegs for leg image links

rshorey:


mileswwatkins:
It's so weird that they only use GIFs for nav images! I think a bunch of chained `try`ing could get muddled, could you turn this into a single `xpath` with a few `or` conditions?

rshorey:
ok, switched to more xpath, less try/except

mileswwatkins:
:pager: :eggplant:

PR #797, opened 2015-07-26
Add and update NC feeds

JessePresnell:
Removed and replaced broken Talking About Politics link

Added several state specific political feeds.


mileswwatkins:
Thanks, Jesse!

PR #799, opened 2015-07-27
VT: Correct legislator emails, and add their state office

mileswwatkins:
@linzjax, want to code review this to get some practice?


rshorey:
looks good, passed quick spot check!

PR #800, opened 2015-07-28
NM: allowed for capitol phone if no office phone

linzjax:
So, couple of questions:
1- do we care about adding area code to phone numbers? (Capital phones are only 7 digits long)
2- Is there a reason that home phone is in 'optional' as well as 'properties'?

screen shot 2015-07-28 at 11 18 36 am

if not is there a reason to keep this:
if found is False and key not in optional:
                 self.warning('bad legislator page %s missing %s' %
                              (url, id_))
                 return

### DIFF
--- a/openstates/nm/legislators.py
+++ b/openstates/nm/legislators.py
@@ -44,7 +44,7 @@ class NMLegislatorScraper(LegislatorScraper):
             # decline to state = independent
             party = 'Independent'
         else:
-            raise Exception("unknown party encountered")
+            raise AssertionError("unknown party encountered")
 
         photo_url = doc.xpath('//img[@id="ctl00_mainCopy_formViewLegislator_imgLegislator"]/@src')[0]
 
@@ -52,10 +52,10 @@ class NMLegislatorScraper(LegislatorScraper):
         if email:
             email = email[0]
         else:
-            print "no email"
+            self.warning('no email')
             email = None
 
-        district_address = ", ".join(doc.xpath(xpath.format('lblAddress')))
+        district_address = "\n".join(doc.xpath(xpath.format('lblAddress')))
 
         district_phone = doc.xpath(xpath.format('lblHomePhone'))
         # if they don't have a home phone, check for an office phone
@@ -68,10 +68,10 @@ class NMLegislatorScraper(LegislatorScraper):
 
         capitol_address = doc.xpath(xpath.format('lblCapitolRoom'))
         if capitol_address:
-            capitol_address = 'Room ' + capitol_address[0] + ' State Capitol, Santa Fe, NM 87501'
+            capitol_address = 'Room ' + capitol_address[0] + ' State Capitol\nSanta Fe, NM 87501'
         else:
             capitol_address = None
-        
+
         capitol_phone = doc.xpath(xpath.format('lblCapitolPhone'))
         if capitol_phone:
             capitol_phone = "(505) " + capitol_phone[0]
@@ -95,12 +95,10 @@ class NMLegislatorScraper(LegislatorScraper):
                 role = 'interim ' + role.lower()
             else:
                 role = role.lower()
+            if ' Committee' in committee:
+                committee = committee.replace(" Committee", '')
+            if ' Subcommittee' in committee:
+                committee = committee.replace(' Subcommittee', '')
             leg.add_role('committee member', term, committee=committee, position=role, chamber=chamber)
-
-        # Already have the photo url.
-        try:
-            del leg['image_url']
-        except KeyError:
-            pass
 
         self.save_legislator(leg)
--- a/openstates/nm/legislators.py
+++ b/openstates/nm/legislators.py
@@ -61,18 +61,19 @@ class NMLegislatorScraper(LegislatorScraper):
                     pass
                 if val:
                     properties[key] = val.strip()
-                else:
-                    properties[key] = None
 
             if found is False and key not in optional:
                 self.warning('bad legislator page %s missing %s' %
                              (url, id_))
                 return
+            elif found is False:
+                properties[key] = None
+            elif properties[key] == ["lblCapitolPhone", "lblOfficePhone"]:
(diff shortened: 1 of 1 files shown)


mileswwatkins:
(We had a conversation in meatspace about this.)

openstates/nm/legislators.py
```diff
@@ -28,6 +28,9 @@ def scrape_legislator(self, chamber, term, url):
         doc = lxml.html.fromstring(html)
         doc.make_links_absolute(url)
 
+<<<<<<< HEAD
```
mileswwatkins: Looks like you have some Git cruft in here from `merge`ing with `master`.

linzjax: I was wondering what that was. How do i get rid of it?

mileswwatkins: You can just go in and edit the text. When you merge a file from two different branches that've both been altered from the original, Git does this so that you can see which part of the change was contributed from each.

linzjax:
Ok, I think that did it.

mileswwatkins:
I'll check this out tomorrow.

openstates/nm/legislators.py
```diff
@@ -28,101 +28,74 @@ def scrape_legislator(self, chamber, term, url):
...
-        leg = Legislator(**properties)
-        leg.add_source(url)
+        photo_url = doc.xpath('//img[@id="ctl00_mainCopy_formViewLegislator_imgLegislator"]/@src')[0]
 
+        email = doc.xpath('//a[@id="ctl00_mainCopy_formViewLegislator_linkEmail"]/text()')
         if email:
-            properties['email'] = email.strip()
+            email = email[0]
         else:
             print "no email"
```
mileswwatkins: We have a `self.warning()` function for worrying debug `print`s like these; try out `self.warning("no email")`.

mileswwatkins: It causes it to print in a different color in a local run.

openstates/nm/legislators.py
```diff
@@ -28,101 +28,74 @@ def scrape_legislator(self, chamber, term, url):
...
             email = None
 
-        leg.add_office('district', 'District Address', address=address,
-                       phone=phone, email=email)
+        district_address = ", ".join(doc.xpath(xpath.format('lblAddress')))
```
mileswwatkins: Separate the street address from the city-state-ZIP using a `newline` instead of a comma.

openstates/nm/legislators.py
```diff
@@ -28,101 +28,74 @@ def scrape_legislator(self, chamber, term, url):
...
+        if not district_phone:
+            district_phone = doc.xpath(xpath.format('lblOfficePhone'))
+        if district_phone:
+            district_phone = district_phone[0]
+        else:
+            district_phone = None
+
+        capitol_address = doc.xpath(xpath.format('lblCapitolRoom'))
+        if capitol_address:
+            capitol_address = 'Room ' + capitol_address[0] + ' State Capitol, Santa Fe, NM 87501'
```
mileswwatkins: Make sure to add the `newline` in between address lines here, too.

openstates/nm/legislators.py
linzjax: Ok, do I need to format the other address the same way?

openstates/nm/legislators.py
```diff
@@ -28,101 +28,74 @@ def scrape_legislator(self, chamber, term, url):
...
             committee = committee.title()
             if 'Interim' in note:
                 role = 'interim ' + role.lower()
             else:
                 role = role.lower()
-            leg.add_role('committee member', term, committee=committee,
-                          position=role, chamber=chamber)
+            leg.add_role('committee member', term, committee=committee, position=role, chamber=chamber)
 
         # Already have the photo url.
```
mileswwatkins: AFAICT, this bit about `image_url` is cruft from the old system; clear it out.

openstates/nm/legislators.py
```diff
@@ -28,101 +28,74 @@ def scrape_legislator(self, chamber, term, url):
...
+        leg.add_office('capitol', 'Capitol Office', address=capitol_address,
+                        phone=capitol_phone, email=email)
 
         # committees
         # skip first header row
         for row in doc.xpath('//table[@id="ctl00_mainCopy_gridViewCommittees"]/tr')[1:]:
-            role, committee, note = [x.text_content()
-                                     for x in row.xpath('td')]
+            role, committee, note = [x.text_content() for x in row.xpath('td')]
             committee = committee.title()
```
mileswwatkins: This wasn't being done before, but could you make sure to standardize the committee names by removing "Committee" and "Subcommittee" from them? Also, if there's a way to figure out which `committee` a `subcommittee` is part of, that would be great, too!

openstates/nm/legislators.py
```diff
@@ -28,101 +28,74 @@ def scrape_legislator(self, chamber, term, url):
...
-        
-
-        properties['url'] = url
-
-        properties['chamber'] = chamber
-        properties['term'] = term
+        xpath = '//span[@id="ctl00_mainCopy_formViewLegislator_{}"]/text()'
 
-        full_name, party = properties['header'].rsplit("-", 1)
+        district = doc.xpath('//a[@id="ctl00_mainCopy_formViewLegislator_linkDistrict"]/text()')[0]
```
mileswwatkins: Can't you use your `xpath` variable here?

openstates/nm/legislators.py
mileswwatkins: :+1:

openstates/nm/legislators.py
linzjax: Nope. It's an 'a' not a 'span'

openstates/nm/legislators.py
linzjax: is that /n or //n?

mileswwatkins: `\n`

PR #804, opened 2015-08-03
CT: Update legislator spreadsheet parsing, and add checks

mileswwatkins:


mileswwatkins:
They added column headers back. How fickle life can be.

PR #805, opened 2015-08-03
IL: Skip over missing bill pages

mileswwatkins:


mileswwatkins:
Only one that I noticed so far. Made sure it was a `500` error, not a `4xx` or the like.

PR #837, opened 2015-09-11
CT: attempt to fix ssl problem that only occurs on prod

rshorey:
yay guess and check.

@mileswwatkins can you check this one to make sure it doesn't look like a terrible idea please?


mileswwatkins:
Let's try!

PR #840, opened 2015-09-16
Fixed parsing dates and times from the KY legislative calendar.

loandy:


mileswwatkins:
Way to get on the board, @loandy!

PR #841, opened 2015-09-22
ME: Added cases for skipping retrieval of inaccessible bill document links.

loandy:


openstates/me/bills.py
```diff
@@ -194,21 +194,31 @@ def scrape_bill(self, bill):
...
-                vdoc.make_links_absolute(ver_url)
-                # various versions: billtexts, billdocs, billpdfs
-                vurl = vdoc.xpath('//a[contains(@href, "billtexts/")]/@href')
-                if vurl:
-                    bill.add_version('Initial Version', vurl[0],
-                                     mimetype='text/html')
+
+                # Check whether the bill text is missing.
+                bill_text_missing = vdoc.xpath('string(//div[@id = "sec0"])').strip()
+                match = re.search('Cannot find requested paper', bill_text_missing)
```
mileswwatkins: Open States generally avoids doing regex-parsing of HTML; could this be switched to Xpath instead? Maybe something like:
`vdoc.xpath('//h1[contains(@text, "Cannot find requested paper")')`?

openstates/me/bills.py
```diff
@@ -194,21 +194,31 @@ def scrape_bill(self, bill):
         if srow:
             bill['subjects'] = [s.strip() for s in srow if s.strip()]
 
+        # Attempt to find link to bill text/documents.
         ver_link = page.xpath("//a[contains(@href, 'display_ps.asp')]")[0]
         ver_url = ver_link.get('href')
+
         try:
             ver_html = self.get(ver_url, retry_on_404=True).text
```
mileswwatkins: @loandy, not necessary to change here, but check out the `lxmlize()` function, a utility that's really really useful in our scraping code:
https://github.com/sunlightlabs/openstates/blob/master/openstates/utils/lxmlize.py

openstates/me/bills.py
```diff
@@ -194,21 +194,31 @@ def scrape_bill(self, bill):
...
+                match = re.search('Cannot find requested paper', bill_text_missing)
+
+                if not match:
+                    vdoc.make_links_absolute(ver_url)
+
+                    # various versions: billtexts, billdocs, billpdfs
+                    vurl = vdoc.xpath('//a[contains(@href, "billtexts/")]/@href')
+
+                    if vurl:
+                        bill.add_version('Initial Version', vurl[0],
```
mileswwatkins: Not a fix, but an enhancement: could we also scrape the `*.rtf` links to bill text? Seems like user-friendly functionality that's easy to add.

mileswwatkins:
Otherwise, looks great, @loandy! Just those three comments.

openstates/me/bills.py
loandy: Thanks, I'll keep that in mind if I run into these cases again.  Also, if you think it's better practice, feel free to edit it in before merging.

mileswwatkins: ```Thanks, I'll keep that in mind when I run into these cases.```
FTFY ;)

openstates/me/bills.py
loandy: This should be addressed by my latest commit.  I removed the regex and managed to make it work strictly with xpath.

openstates/me/bills.py
loandy: I made this change on a separate branch, which I'll make a pull request for in just a moment.

loandy: Merged those changes into the current pull request branch.

mileswwatkins:
Thanks @loandy!

PR #842, opened 2015-09-22
Added additional processing for ... bill text documents.

loandy:


loandy:
Sorry if this makes life more difficult Miles.  Thought it would be better to keep this functionality separate.

PR #844, opened 2015-09-24
Added handling missing office phone numbers for WV legislators.

loandy:
I also removed an extraneous e-mail parsing statement that by all means looks to be useless.  I tested the scraper and it's still pulling e-mails properly.


mileswwatkins:
Looks great, @loandy!

As long as you're in WV `legislators.py` and fixing contact info: these members should have a `district` office if they have _at least one of_ a home phone, home address, or business phone. Right now we're not even grabbing home address, and're only looking for business phone.

loandy:
I'll take a look into it.

loandy:
The latest commit should deal with district offices now.

mileswwatkins:
Perfecto :D

PR #849, opened 2015-10-02
AL: Populated first/last name for legislators

drobertsus:
Simple addition after investigating the JIRA DATA-510 issue. Not sure how first/last names are currently being populated on API but I'm going on the assumption that it is using fixed data. Feedback is appreciated - I was just looking for an easy issue for a first contribution.


drobertsus:
Added an solution to  DATA-512. It looks like this pull request now includes both changes because no branching.

loandy:
Hey @drobertsus, thanks for taking a look.  For DATA-510, from what I know, the issue wasn't caused by the code, but resulted from the restrictions we have configured in production - the legislator's name should be updated properly now.

For DATA-512, I took a look at your commit and it looks good for the most part.  If you could remove the `import sys` statement on line 4 and fix the comment error on line 68, I'll go ahead and pull it in.

PR #903, opened 2015-12-22
Add Louisiana 2016 Regular Session to scraper

showerst:
Louisiana recently added a 2016 session, and scraper was failing with a missing session error.

I've tried adding it myself, it seems to be working now, but this is my first time trying to contribute to this project, so please let me know if I've done anything wrong.


showerst:
Actually this seems to be wrongly pulling in a bunch of 2015 bills and marking them as 2016, so i'm going to kill this PR until I figure out why.

PR #919, opened 2016-01-11
modded if statement that checks to see what stage bills are in, house…

dmc2015:
… or senate, and when they are passed. Now the 2013-2014 session bill scrapper should run without error and jira issue DATA-362 can be closed


loandy:
Checks out.  Thanks.

PR #934, opened 2016-01-15
hotfix for mo senate bill text

Skekacs93:
Missouri change the structure of the listing of their bill text links.


loandy:
Hello again Steve,

Would you mind submitting each change set in a separate PR as well?

PR #936, opened 2016-01-19
Added 2016 session info for Idaho

showerst:
Added 2016 Session information for Idaho.

Note: the display name for the legislature seems out of order from what's already in the code, but I went with the official source at https://www.legislature.idaho.gov/sessioninfo/2016/Journals/journals.htm .

This is my first PR for this repo, so please let me know if I should change how I do it in the future.

### DIFF
--- a/openstates/id/__init__.py
+++ b/openstates/id/__init__.py
@@ -74,7 +74,7 @@ metadata = {
             'name': '2015-2016',
             'start_year': 2015,
             'end_year': 2016,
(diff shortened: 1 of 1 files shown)


loandy:
Whoops.  Posted this in the wrong box on accident:

Looks like this checks out.  Thanks for the update, and also for catching the issue with the metadata being wrong.  I'll be fixing that shortly.

PR #944, opened 2016-01-20
added the right session number for the 2015 special session

dmc2015:
Run this to remove the duplicate special sessions
```
js
db.bills.remove({ 'state': 'va', 'session': '2015specialI'})
```

Run this to get the right bills after committing the fix
```
bash
billy-update --session=2015specialI --bills --fast va
```


loandy:
Thanks, Don!

PR #945, opened 2016-01-21
added another edge case for future bill actions

Skekacs93:
'Placed on calendar...' for future events was breaking the bill scraper. Fix adds search to handle this.


loandy:
I think the action in question, which I believe is the "placed on second reading" action regarding WA bill HB 5255, must have been improperly modified to be dated to the current year - our own production data has this action on the same date in 2015.  The "scheduled on" actions should be the only ones that take place in the future.

I'm going to close this for now and see about contacting someone about this issue upstream.

PR #946, opened 2016-01-21
allow for no address for one legislator to allow for rest of informat…

Skekacs93:
…ion to update (same case as RI)


loandy:
Looks good.  Thanks.

PR #952, opened 2016-01-25
fixing issue with scraper for anything outside of a normal session

dmc2015:
Pretty concerned the init for la. 
Dates are missing from the session or are incorrect, I am wondering how this was working previously.
When no date is set it, ’start_date’ - ‘end_date’, it seems to be pulling 2016 data. On my local version each session has 28 bills.

image

I am fairly sure the 2011 Special session fix is good though.

Regarding the other session specific scrapes:
We are going to a URL for scraping and it is stuck on 2016, it doesn’t change based on session selected. I am guessing the production environment already has historical data scraped prior to them changing the URL on the site to being specific to 2016. So this secondary issue is only impacting local versions of OpenStates.

On line 15 of la/bills.py a URL var is set, "http://www.legis.la.gov/Legis/BillSearchListQ.aspx?r=%s1*" . On line 44 we format the string and it goes here

http://www.legis.la.gov/Legis/BillSearchListQ.aspx?r=SB1*
or 
http://www.legis.la.gov/Legis/BillSearchListQ.aspx?r=HB1*

That link is always 2016. It is not directly related to my fix but we should keep it in mind and never remove the la collection. I also I want to make sure that the issue can be replicated.

### DIFF
--- a/openstates/la/bills.py
+++ b/openstates/la/bills.py
@@ -239,6 +239,12 @@ class LABillScraper(BillScraper, LXMLMixin):
             date, chamber, page, text = [x.text for x in action.xpath(".//td")]
             date += "/%s" % (session)  # Session is April --> June. Prefiles
             # look like they're in January at earliest.
+
+            #required to remove the special session extended name, this will remove any characters, including and following a space
+            if (re.search('\s(.*)', date) is not None):
+                m = re.search('\s(.*)', date)
+                date = date[:m.start()]
+
             date = dt.datetime.strptime(date, "%m/%d/%Y")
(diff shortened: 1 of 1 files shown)


loandy:
I ended up creating a more robust fix for this issue.

dmc2015:
@loandy where is the pull request for the solution?

loandy:
You can check my Jan 26, 2016 commits for LA for more, but the critical changes are here: https://github.com/sunlightlabs/openstates/commit/068457442a3cf2c111d67f00e211a2f20aa05fd9

dmc2015:
@loandy thanks

PR #953, opened 2016-01-25
handles slightly changed structure of website

Skekacs93:
IA recently modified the structure of their website, includes several changed to handle this. 
Subject mapping not working -- seems like they got rid of the html page with all bills for each subject and replaced with a pdf. 

Also handles one minor unicode bug with votes.


loandy:
This looks pretty good.  I did figure out the structure to grab the bill version HTML format documents, so I'll be adding that functionality in after I merge.

PR #957, opened 2016-01-26
small syntax fix

Skekacs93:


loandy:
Thanks for catching that.

PR #977, opened 2016-02-04
WA: added regex case for irregular doc name

Skekacs93:


loandy:
This issue seems to have resolved on its own and it seems the change breaks the scraper whenever I've tested it.  Also, whitespace error.  I'm closing this for now.

PR #980, opened 2016-02-07
WA: Added legislators' website URLs

richard-ctcl:


loandy:
Hi Richard,

Are there cases where you're finding legislators missing their website URLs?  From the API calls that I've checked, it looks like the legislators are returning this information.

richard-ctcl:
I also see that the API output shows the `url` field present and populated. I'm not seeing anything in the scraper that would account for that. From my own testing:

Prepare test environment:
```javascript
> db.legislators.remove({ state: 'wa' })
WriteResult({ "nRemoved" : 146 })
```

Following pre-patch run of billy-update:
```javascript
> db.legislators.find({ state: 'wa' }).count()
146
> db.legislators.find({ state: 'wa', url: { $exists: true }}).count()
0
```

Post-patch:
```javascript
> db.legislators.find({ state: 'wa', url: { $exists: true }}).count()
146
```

loandy:
Odd.  I'll look into it some more from my end.

loandy:
Tested.  Looks good.  After doing some investigation, I think that the best explanation is that the legislator URLs were mostly retained due to low turnover, as their scraping was eliminated in a rewrite from last year.

PR #989, opened 2016-02-09
OH: Generate legislators' email addresses

richard-ctcl:
Ohio representatives and senators do not have their email addresses listed on their official websites, but the addresses can be reliably generated based on the district number. See, e.g.:

* House: http://www.semasan.com/page.asp?content=aa_2014OH2&g=SEMAGA
* Senate: http://www.oh-municipalities.com/publication/index.php?i=183210&m=&l=&p=122&pre=&ver=html


loandy:
Ran a couple of different e-mails through various e-mail validators, which all say different things, but this generally seems to check out.  If you can resubmit this without the print statement on 162, I'll merge it in.

PR #996, opened 2016-02-10
Added new kansas introduction action code

showerst:
KS added a new action code, so right now we're getting the warning:

15:40:55 WARNING billy: unknown action code on HB2375: intro_ibc_158 Introduced

This fix allows the action to be parsed.


loandy:
I believe the action code file is obsolete and the bill scraper looks for the action codes in `ksapi`.  I'll be removing the `action_codes` file.  Thanks for pointing out the issue though.  Closing for now.

PR #1005, opened 2016-02-17
Reworked FL Senate Event Scraper for new feed

showerst:
The FL events scraper was broken because the Senate changed their feed. This rewrite makes it work again for the senate.


showerst:
Actually I can add a little better link metadata here, going to close this and reissue.

PR #1011, opened 2016-02-18
WV: minor vote counts fix

Skekacs93:


loandy:
Hey Steve,

Do you have some examples of vote count issues that this addresses?

Skekacs93:
http://www.legis.state.wv.us/legisdocs/2016/RS/votes/senate/02-15-0100.pdf

The vals regex was matching "A plans that are exempt from regulation by Insurance", causing an extra count to be added to other_count.

Skekacs93:
Then splitting the text on '\n\n\n' was breaking http://www.legis.state.wv.us/legisdocs/2014/RS/votes/senate/02-03-0003.pdf.

PR #1012, opened 2016-02-19
Fix IA Events Scraper

showerst:
The Iowa events scraper wasn't saving any objects due to changes in the page HTML. This update fixes that.


loandy:
Cool, thanks.

PR #1017, opened 2016-02-25
NE: added exception for committee sponsors (not included within a tags)

Skekacs93:
ex: http://nebraskalegislature.gov/bills/view_bill.php?DocumentID=28140


loandy:
Hey Steve,

Could you add a colon to the end of the 'Introduced By' phrase in the `split` call on line 76 and then separate the line itself into two lines?  I'd like to keep to [PEP-8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) whenever possible.

loandy:
Cool, thanks.

PR #1018, opened 2016-02-26
OK: Made filter regex case-insensitive.

loandy:
This is because the URLs we filter on are not in a consistent case.

### DIFF
--- a/openstates/ok/bills.py
+++ b/openstates/ok/bills.py
@@ -123,7 +123,7 @@ class OKBillScraper(BillScraper):
             version_url = link.attrib['href']
             name = link.text.strip()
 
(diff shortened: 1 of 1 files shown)


loandy:
Looking at Oklahoma Senate bill [SB 85](http://www.oklegislature.gov/BillInfo.aspx?Bill=SB85&session=1600) for the 2016 Regular Session, the regex was picking up the Senate Conference Committee Report dated 2/18/2016 due to the lowercasing of "reports" in the link: `http://webserver1.lsb.state.ok.us/cf_pdf/2015-16%20COMMITTEE%20reports/Senate/SB85%20SCCR.PDF`.

PR #1019, opened 2016-02-29
modded gov signed rule regex

dmc2015:


loandy:
Looks good, Don.  Thanks!

PR #1024, opened 2016-03-02
Fixed date value out of range error in california db download script 

showerst:
The script that downloads the California bill database breaks now, because there's a date of Feb 29 without a year, which throws a date value out of range error.

This patch adds a special case. The 2020 leap day is on a Saturday, so I just hard coded a Feb 29 check rather than try to infer the date since we should be good for 8 years.


loandy:
It looks like this issue has been resolved by the other changeset that I made: [2d1a32c](https://github.com/sunlightlabs/openstates/commit/2d1a32c42f981cae0dd70faed8c4751b90bfc7e0).  Closing.

PR #1027, opened 2016-03-04
NH 2016

showerst:
Update the NH scraper to add 2016 Session and deal with a few format changes.

### DIFF
--- a/openstates/nh/bills.py
+++ b/openstates/nh/bills.py
@@ -80,7 +80,6 @@ class NHBillScraper(BillScraper):
             bill_id = line[10]
 
             if body == body_code[chamber] and session_yr == session:
-                print "%s %s" % (expanded_bill_id, bill_id)
                 if expanded_bill_id.startswith('CACR'):
                     bill_type = 'constitutional amendment'
                 elif expanded_bill_id.startswith('PET'):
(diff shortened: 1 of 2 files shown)


openstates/nh/bills.py
```diff
@@ -84,6 +84,8 @@ def scrape(self, chamber, session):
                     bill_type = 'constitutional amendment'
                 elif expanded_bill_id.startswith('PET'):
                     bill_type = 'petition'
+                elif expanded_bill_id.startswith('AR') and bill_id.startswith('CACR'):
```
loandy: Do you have an example of a bill where this is not getting tagged properly?

showerst:
If you take that line out, you'll get a key error on the following line:
                    bill_type = bill_type_map[expanded_bill_id.split(' ')[0][1:]]

It's hitting this line in the data (tbllsrs.txt) and choking:

2016|2057|relating to the election of judges. Providing that judges be elected for specific terms.|H|3|0|0|0|16-2057|AR0015|CACR15||CAC|H10|H10|1/6/2016 12:00:00 AM|09|1/7/2016 2:09:59 PM|3/3/2016 12:00:00 AM|1/27/2016 12:00:00 AM|0|H10|H10||||||0|02|H10|1/13/2016 10:30:00 AM|LOB Room 208||||0|0|

I think the issue is that this year they started coding constitutional amendments with "ARDDDD" (where d is a digit), where in the past they'd always coded in the format "CACR DDDD". There are a number of instances in that datafile that follow the same pattern. I looked a few of them up on the NH state site and they're all Constitutional Amendments.

loandy:
Looks great.  Thanks!

PR #1039, opened 2016-03-11
CA Add Event Committees

showerst:
The CA event scraper was failing because it was missing some committees that were added during the recent special sessions in the lookup list.


openstates/ca/events.py
```diff
@@ -137,6 +140,8 @@ def scrape(self, chamber, session):
     62L: u'Senate Budget and Fiscal Review',
     64L: u'Senate Environmental Quality',
     66L: u'Senate Veterans Affairs',
+    67L: u'Senate Transportation and Infrastructure Development',
+    68L: u'Senate Public Health and Developmental Services'
```
loandy: This line needs a comma.

loandy:
Looks good, thanks.

PR #1043, opened 2016-03-16
CA: Fix assembly member address parsing, address type classification

rsimoes:
See any lower-chamber member; e.g., http://openstates.org/api/v1/legislators/?state=ca&chamber=lower&district=22

This patch fixes:
  - "CA" showing up twice in assembly members' addresses
  - District addresses being classified as capitol addresses, and capitol addresses not being included at all


loandy:
Looks good.  Thanks!

PR #1044, opened 2016-03-16
GA: Skip legislators who have vacated

rsimoes:
The production openstates API appears to be manually pruned of vacated members, but the scraper adds them.

### DIFF
--- a/openstates/ga/legislators.py
+++ b/openstates/ga/legislators.py
@@ -50,6 +50,20 @@ class GALegislatorScraper(LegislatorScraper, LXMLMixin):
         for member in members:
             guid = member['Id']
             member_info = backoff(self.sservice.GetMember, guid)
+
+            # Check to see if the member has vacated; skip if so:
+            try:
+                legislative_service = filter(
+                    lambda service: service['Session']['Id'] == sid,
+                    member_info['SessionsInService']['LegislativeService']
+                )[0]
+            except IndexError:
+                raise Exception("Something very bad is going on with the "
+                                "Legislative service")
+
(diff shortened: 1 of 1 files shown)


loandy:
Hey Richard,

Do you have any examples of which legislators seem to be re-added?

rsimoes:
One example is Harry Geisinger, of District 48, who died in May 2015. The GA scraper in its current state pulls his entry from the SOAP API and adds it despite there being a populated DateVacated field. My copy of the fiftystates database ends up with two District 48 entries--the other being for Geisinger's successor, Betty Price. The production openstates API lists only Price: http://openstates.org/api/v1/legislators/?state=ga&district=48&chamber=lower

openstates/ga/legislators.py
```diff
@@ -50,6 +50,20 @@ def scrape_session(self, term, chambers, session):
         for member in members:
             guid = member['Id']
             member_info = backoff(self.sservice.GetMember, guid)
+
+            # Check to see if the member has vacated; skip if so:
+            try:
```
loandy: Hey Richard,

Generally, I'm not opposed to making things more functional, but for the Open States code base, I'd like to avoid these sorts of function compositions with lambdas if possible for a number of reasons.  I would consider something like the following more fitting:
```
legislative_service = next(service for service in
    member_info['SessionsInService']['LegislativeService’]
    if service['Session']['Id'] == sid)
```
Otherwise, everything checks out.
Everything checks out, though.

rsimoes:
Sure thing; I'll prefer list comprehensions going forward.

PR #1047, opened 2016-03-18
FL: Update detection of vacant House seats

rsimoes:
See http://www.myfloridahouse.gov/Sections/Representatives/representatives.aspx#ctl00_ContentPlaceHolder1_divInactiveMemberList for reference

### DIFF
--- a/openstates/fl/legislators.py
+++ b/openstates/fl/legislators.py
@@ -132,9 +132,9 @@ class FLLegislatorScraper(LegislatorScraper):
             link = div.xpath('.//div[@class="rep_style"]/a')[0]
             name = link.text_content().strip()
 
-            if 'Vacant' in name or \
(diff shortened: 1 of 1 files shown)


openstates/fl/legislators.py
```diff
@@ -132,9 +132,9 @@ def scrape_reps(self, term):
             link = div.xpath('.//div[@class="rep_style"]/a')[0]
             name = link.text_content().strip()
 
-            if 'Vacant' in name or \
-                    'Resigned' in name or \
-                    'Pending' in name:
+            term_details = div.xpath(
+                './/div[@class="term_style"]/a')[0].text_content()
```
loandy: The `term_style` divs contain no anchors, but once you remove the trailing `/a` in the xpath, it works.

PR #1052, opened 2016-03-18
Fixed TN Bill scraper to respect pagination

showerst:
TN is only showing 100 bills at time on their listing page, so scrape the index of listing pages, then hit every listing page link.

Also, remove double-space in bill names.

### DIFF
--- a/openstates/tn/bills.py
+++ b/openstates/tn/bills.py
@@ -237,10 +237,11 @@ class TNBillScraper(BillScraper):
 
         if 'B' in bill_id:
             bill_type = 'bill'
-        elif 'JR' in start_abbr:
+        elif 'JR' in bill_id:
             bill_type = 'joint resolution'
-        else:
+        elif 'R' in bill_id:
             bill_type = 'resolution'
+            
 
         primary_chamber = 'lower' if 'H' in bill_id else 'upper'
         # secondary_chamber = 'upper' if primary_chamber == 'lower' else 'lower'
--- a/openstates/tn/bills.py
+++ b/openstates/tn/bills.py
@@ -190,30 +190,28 @@ class TNBillScraper(BillScraper):
     jurisdiction = 'tn'
 
     def scrape(self, term, chambers):
-        #types of bills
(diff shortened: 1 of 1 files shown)


openstates/tn/bills.py
```diff
@@ -229,13 +227,20 @@ def scrape_bill(self, term, bill_url, bill_type):
...
-        bill_id = bill_id.replace('*', '').strip()
+            secondary_bill_id = secondary_bill_id.replace('  ',' ')
+            
+        bill_id = bill_id.replace('*', '').replace('  ',' ').strip()
+
+        if 'B' in bill_id:
+            bill_type = 'bill'
+        elif 'JR' in start_abbr:
+            bill_type = 'joint resolution'
+        else:
```
loandy: May want to turn this `else` case into `elif 'R' in start_abbr`, and reserve the `else` case to emit an exception.  Although the assumption that anything else would be a resolution is largely good, it could potentially capture faulty data, and we would want to explicitly know if that's the case so that the issue can be investigated.

showerst: That's just copy-pasted from the existing code, I just moved it from the landing-page scraper to the bill page scraper.
The exception makes sense, i'll update the code and re-PR shortly.

openstates/tn/bills.py
```diff
@@ -229,13 +227,20 @@ def scrape_bill(self, term, bill_url, bill_type):
...
                 secondary_bill_id = secondary_bill_id.strip()
-
-        bill_id = bill_id.replace('*', '').strip()
+            secondary_bill_id = secondary_bill_id.replace('  ',' ')
+            
+        bill_id = bill_id.replace('*', '').replace('  ',' ').strip()
+
+        if 'B' in bill_id:
+            bill_type = 'bill'
+        elif 'JR' in start_abbr:
```
loandy: I'm not seeing where `start_abbr` is being defined.  Am I missing something?

openstates/tn/bills.py
loandy: Yeah, I noticed that.  Feel free to edit it, or I can do so later.

PR #1053, opened 2016-03-20
KS: Throttle committee scraping when receiving 429s.

cliftonmcintosh:
This is in response to https://sunlight.atlassian.net/browse/DATA-564. Please see my comment on that ticket.

The throttling mechanism is modeled on one in openstates/ny/apiclient.py.

The 429 asks for a pause of 5 seconds. However, I found that using that pause time caused occasional failures, possibly because the request was being made over a connection that had closed before the retry could be made. I divided the wait in half and found that the KS legislature API still let the requests through.

It should also be noted that one of the cache file names is too long for me, and scrapelib/cache.py chokes when trying to write a cache file for one of the committees. This may be particular to the location and name of the directories in my system. I had to comment out the file opening in scrapelib/cache.py to allow the scraping to complete.

Here's the end of the trace that I received:
  File "/home/cmcintosh/.virtualenvs/openstates/local/lib/python2.7/site-packages/scrapelib/cache.py", line 168, in set
    with open(path, 'wb') as f:
IOError: [Errno 36] File name too long: '/home/cmcintosh/code/openstates/cache/www.kslegislature.org,li,api,v7,rev-1,ctte,ctte_sub_department_of_administration_subcommittee_ways_and_means_1,,1355e9ab535edc13abe3a9867b026aee'


loandy:
Hi Clifton,

Billy actually supports request throttling measured in requests per minute through the `-r` or `--rpm` option.  We managed to have a successful scrape after implementing throttling on our runs through the use of this option, but now it isn't recognizing chamber journal file formats, which I'm currently addressing.

cliftonmcintosh:
Thanks. Good to know. I'll close this pull request.

PR #1058, opened 2016-03-21
GA: Skip legislators who have vacated

rsimoes:
With requested modification


loandy:
Cool, thanks for making that change.  Looks good.

PR #1063, opened 2016-03-23
NC: prepare for 2016 Extra Session 2 (starts 3/23)

Skekacs93:


openstates/nc/__init__.py
```diff
@@ -83,8 +83,13 @@
                  'display_name': '2015-2016 Session',
                  '_scraped_name': '2015-2016 Session',
                 },
+        # '2016 Extra Session 2': {'start_date': datetime.date(2016,3,23), 'type': 'special',
```
loandy: Since we're ignoring the session, no need to keep this in.

Skekacs93:
Ok, I was ignoring it for now because the scraper won't run without it, but had the session info in there so we can easily uncomment it once it starts?

loandy:
I debated that, but for now, I'd like to address things as they come and keep the codebase cleaner.  Also, it looks like the session key format used in the past is something like `2016E2`.  Not sure why all of those commented terms exist, but I'm inclined to remove them eventually.

Skekacs93:
Alright sounds good.

loandy:
Thanks!

PR #1064, opened 2016-03-24
AL: Fixed Bill URLs in case of Special Sessions

gilhooleyd:
This is related to bug [[Data-541]](https://sunlight.atlassian.net/browse/DATA-541).

### DIFF
--- a/openstates/al/bills.py
+++ b/openstates/al/bills.py
@@ -235,10 +235,19 @@ class ALBillScraper(BillScraper):
                 title = "[No title given by state]"
             bill['title'] = title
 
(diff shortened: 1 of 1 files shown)


loandy:
Hey David,

I've been looking into this issue, and I think I may have found a better (albeit slightly more involved) way to solve it, but I'm currently putting it through testing.  I think that the session keys in the metadata should probably be reworked for consistency's sake, and resolving the issues this causes in existing databases is a straightforward problem to solve.

loandy:
This issue should be resolved now.  Closing.

PR #1065, opened 2016-03-24
MS: add exception for legislator with inadequate profile

Skekacs93:


openstates/ms/legislators.py
```diff
@@ -101,6 +101,9 @@ def scrape_details(self, chamber, term, leg_name, leg_link, role):
             elif leg_name in ('James W. Mathis'):
                 assert not party, "Remove special-casing for this Republican without a listed party: {}".format(leg_name)
                 party = 'Republican'
+            elif leg_name in ('J. Walter Michel'):
```
loandy: Hey Steve,

It looks like this legislator is a Republican, so you can add his name to the `elif` case above for Republicans.

Skekacs93:
yup!

PR #1068, opened 2016-03-26
NM: add committee scraper, remove committee roles from legislators scraper

cliftonmcintosh:
This is for [DATA-499](https://sunlight.atlassian.net/projects/DATA/issues/DATA-499).

Data is being pulled from the NM legislature's [new site](http://www.nmlegis.gov:8080/). Specifically, http://www.nmlegis.gov:8080/Committee/Default is the root from which committee scraping starts.

This version of the scraper maintains the use of "interim member," "interim co-chair," etc. that the legislator scraper was assigning when it was used to get committee information. It seems that in NM's case "interim" does not seem to mean "temporary until something permanent comes along." "Interim" seems to mean committees that meet when the legislature is not in session. At least that's what I gathered [by Googling](https://www.google.com/webhp?#q=new%20mexico%20legislature%20interim%20committees). If that is true, perhaps "chair" is a more accurate role than "interim chair." If this is something that should be changed, let me know, and I can remove the "interim" part of the role.

The committee scraper also follows the legislators' scraper's policy of removing "Committee" and "Subcommittee" from the name of each committee. Is that the expected practice for openstates.org data?

Note that most interim committees are joint legislative committees, and this was not being captured when the committee information was derived from the legislator scraper.

### DIFF
--- a/openstates/nm/committees.py
+++ b/openstates/nm/committees.py
@@ -1,0 +1,104 @@
+from billy.scrape.committees import CommitteeScraper, Committee
+from openstates.utils import LXMLMixin
+
+base_url = 'http://www.nmlegis.gov:8080/Committee/'
+
+
+class NMCommitteeScraper(CommitteeScraper, LXMLMixin):
+    jurisdiction = 'nm'
+
+    def scrape(self, chamber, term):
+        self.validate_term(term)
+
+        # Xpath query string format for legislative chamber committee urls
+        base_xpath = '//table[@id="MainContent_gridView{0}Committees"]' \
+                     '//a[contains(@id, "MainContent_gridView{1}Committees_link{2}Committee")]/@href'
+
+        if chamber == 'upper':
+            url = '{}Senate_Standing'.format(base_url)
+            chamber_xpath = base_xpath.format('Senate', 'Senate', 'Senate')
+
+            # Most interim committees are joint. Use 'joint' chamber as the way
+            # to collect Interim committee data
+            self.scrape('joint', term)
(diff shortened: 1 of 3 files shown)


openstates/nm/committees.py
```diff
@@ -0,0 +1,104 @@
...
+
+        elif chamber == 'lower':
+            url = '{}House_Standing'.format(base_url)
+            chamber_xpath = base_xpath.format('House', 'House', 'House')
+
+        elif chamber == 'joint':
+            url = '{}Interim'.format(base_url)
+            chamber_xpath = base_xpath.format('', '', '')
+
+        self.debug('Scraping NM {0} chamber term {1} for committees listed at {2}'.format(chamber, term, url))
```
loandy: Hey Clifton,

Generally, we try to leave debugging statements out of the source.

loandy:
Hey Clifton,

In the interest of consistency, I would prefer to maintain a 79-char limit as per [PEP-8](https://www.python.org/dev/peps/pep-0008/#id17) for Open States (I should probably make that better known somewhere).  I also made a comment about removing a line, but other than these simple issues, it looks good to go.  Great work!

cliftonmcintosh:
Andy,

Thanks. I'll be happy to make the changes.

cliftonmcintosh:
@loandy,

I've made updates to address the line lengths and to remove the debug statement. Please let me know if there are other changes that would be useful.

loandy:
Great!  Thanks for making those changes.

PR #1072, opened 2016-03-28
ND: Preserve vertical whitespace in addresses

rsimoes:
Street names in North Dakota appear rife with directional suffixes. This is a tiny patch to prevent ambiguous addresses such as "752 51st Street South Fargo, ND" (the city of Fargo and not South Fargo) and "1822 Brentwood Court West Fargo, ND" (the city is West Fargo and not Fargo).


loandy:
Sounds fair.  Thanks!

PR #1075, opened 2016-03-30
MO: Prevent legislators from prior sessions from being added, other adjustments

rsimoes:
- Prevent addition of former legislators by referencing only current rosters
- Rewrite of senate scraper due to new layout of senator pages
- Ensure email addresses are included in house members' office details
- Sundry, small adjustments for PEP8 conformance and clarity of variable names

In case of mid-term legislator changes, both members were being added:
- http://www.house.mo.gov/member.aspx?year=2015&district=029
- http://www.house.mo.gov/member.aspx?year=2016&district=029

It turns out that the MO House site will bring up the current legislator if the year is excluded:
- http://www.house.mo.gov/member.aspx?district=029

### DIFF
--- a/openstates/mo/legislators.py
+++ b/openstates/mo/legislators.py
@@ -1,30 +1,31 @@
 import lxml.html
-import datetime
+import re
 from billy.scrape.legislators import LegislatorScraper, Legislator
+
 
 class MOLegislatorScraper(LegislatorScraper):
 
     jurisdiction = 'mo'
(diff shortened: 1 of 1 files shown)


fgregg:
@katduffy any updates on this PR?

loandy:
Hey Forest, I keep meaning to revisit this particular changeset.  Since Missouri keeps their legislators separated by year, I thought it would be nice to be able to retain the ability to capture non-current sessions.  Also, when I tested this, it appeared that the member list when specifying the year was more up to date than the "current" listing without.

I wanted to review and see if I could get the meat of the changes here while preserving the functionality already written.

PR #1081, opened 2016-04-07
NM: reference NM legislature's new site in bill scraping

cliftonmcintosh:
Sets the "source" for each bill to the location of that bill on the new site. (For example, like http://www.nmlegis.gov:8080/Legislation/Legislation?chamber=H&legType=B&legNo=1&year=16)

Main data is still retrieved from the same ftp location.
Documents (like PDFs for votes) are still retrieved from the same locations. (Fior example, like "http://www.nmlegis.gov/Sessions/16%20Regular/firs/")

### DIFF
--- a/openstates/nm/__init__.py
+++ b/openstates/nm/__init__.py
@@ -108,8 +108,8 @@ metadata = {
 
 
 def session_list():
-    return url_xpath('http://www.nmlegis.gov:8080/',
-        '//select[@name="ctl00$MainContent$ddlSessions"]/option/text()')
+    return url_xpath('http://www.nmlegis.gov/lcs/keyword.aspx',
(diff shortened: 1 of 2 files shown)


loandy:
The changes to session scraping in `__init__.py`, although related, should probably be separate from the changes to `bills.py`.  This case is simple enough, but for the future, it helps comprehension and testing changesets in isolation to submit changesets which affect different functionality in separate pull requests, which I'm totally fine with.

I'm still getting used to managing PRs from a wider community, so please bear with me.

cliftonmcintosh:
@loandy,

Makes sense. I will remove that from the PR.

loandy:
Great, thanks!  Looks good to me.

PR #1085, opened 2016-04-08
AZ: fix for missing actions

Skekacs93:
A good amount of bill actions have been being missed because the scraper only looks for actions within the table with class='ContentAreaBackground'. This adds the actions that are outside this table. Example: http://www.azleg.gov//FormatDocument.asp?inDoc=/legtext/52Leg/2r/bills/sb1010o.asp&Session_ID=115.


loandy:
Great work, looks good.  Thanks!

PR #1086, opened 2016-04-08
RI: added exception to event time

Skekacs93:
Covers "TIME:	Immediately after the full Finance committee meeting scheduled for the Rise" from http://status.rilin.state.ri.us/documents/agenda-11801.aspx.


loandy:
Hey Steve,

I think I've found a better way to distill the "all day" checking based on the cases that we've seen.  We can revisit if a new case pops up that breaks it, but I'll close this for now.

PR #1091, opened 2016-04-11
KY: Expanded skipping for actions that could not be parsed.

loandy:


loandy:
This patch is kind of a stopgap measure.  Any actions that have elements embedded in the text (veto actions in particular) are troublesome given the markup structure of the content and cannot currently be parsed without jumping through some rather stupid hoops.

PR #1093, opened 2016-04-14
AR: add 2016 Fiscal Session (started 4/13)

Skekacs93:


openstates/ar/__init__.py
```diff
@@ -116,11 +116,18 @@
             'display_name': '2016 Second Extraordinary Session',
             '_scraped_name': 'Second Extraordinary Session, 2016',
             'slug': '2016S2',
-        }
+        },
+        '2016F': {
+            'type': 'special',
+            'start_date': datetime.date(2016, 04, 13),
+            'end_date': None,
```
loandy: You can omit the `end_date` field if it's indeterminate.

PR #1095, opened 2016-04-15
NY: patch for senate votes

Skekacs93:
Senate Votes haven't been being pulled in recently, this adds a senate vote builder (assembly site only posts assembly votes and senate site only posts senate votes... annoyingly).


loandy:
This issue should be addressed with the latest refactoring.  This is also quite a massive changeset.  Closing.

PR #1100, opened 2016-04-26
don't try to add blank actions to bills in ky

showerst:
KY HB15 has an action with a blank in between two semicolons - "; ;". This is being interpreted as blank actions and blowing up the scraper.

http://www.lrc.ky.gov/record/16RS/HB15.htm

This hotfix ignores blank actions to avoid the validation error and allow the scraper to continue.


loandy:
Looks good.  Thanks!

PR #1103, opened 2016-04-30
Handle failure when disable_warnings is not available

kamyarrasta:
Fail gracefully if disable_warnings not available.
See [this](https://github.com/jkbrzt/httpie/issues/418)


loandy:
Thanks for this.

PR #1104, opened 2016-05-01
NM: remove port 8080 from addresses.

cliftonmcintosh:
The new NM legislature site is live and now works without port 8080 in the addresses. This removes `:8080` from all addresses.


loandy:
Checks out.  Thanks!

PR #1116, opened 2016-05-11
VA: fixed votes with abstentions not being parsed

danielmroberts:
Votes that include abstention values are being skipped. I have updated the regular expression to include an allowance for an abstention count in the vote name.


loandy:
Thanks for this.  I wrote a few additions which supersede this changeset.  Closing.

PR #1124, opened 2016-05-19
Added 2016 WV special session

showerst:
Added 2016 WV special session and bill scraper changes to accommodate future special sessions.


loandy:
Looks to work alright.  Thanks!

PR #1133, opened 2016-07-05
Ca vetoes

showerst:
CA Sometimes uses 'Vetoed by the governor' in addition to 'Vetoed by Governor' so we were not classifying that action as governor:vetoed.

EX: http://www.leginfo.ca.gov/pub/15-16/bill/sen/sb_0851-0900/sb_894_bill_20160701_history.html


loandy:
Checks out.  Thanks!

PR #1145, opened 2016-08-08
Added FL 2017 session

showerst:
FL has started prefiling and has bills up on the site, so add 2017 session metadata.

### DIFF
--- a/openstates/fl/__init__.py
+++ b/openstates/fl/__init__.py
@@ -34,7 +34,7 @@ metadata = {
             'name': '2015-2016',
             'start_year': 2015,
             'end_year': 2016,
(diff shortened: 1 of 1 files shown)


openstates/fl/__init__.py
```diff
@@ -34,7 +34,7 @@
             'name': '2015-2016',
             'start_year': 2015,
             'end_year': 2016,
-            'sessions': ['2015', '2015A', '2015B', '2015C', '2016'],
+            'sessions': ['2015', '2015A', '2015B', '2015C', '2016','2017'],
```
loandy: Probably should create a new term entry for 2017-2018 for the 2017 session.

mileswwatkins:
Covered in https://github.com/openstates/openstates/pull/1171

PR #1146, opened 2016-08-09
KY: Added code to scrape prefiled bills if within prefile timespan

showerst:
KY has a seperate set of listing pages for prefiled bills. 

This patch adds a new session metadata attribute 'prefile_start_date', and if the date falls between prefile start and session start, it checks the prefile page instead of the standard pages, which won't exist yet.

### DIFF
--- a/openstates/ky/__init__.py
+++ b/openstates/ky/__init__.py
@@ -118,8 +118,8 @@ metadata = {
         },
         '2017RS': {
             'type': 'primary',
-            'start_date': datetime.date(2017, 1, 5),
-            'end_date': datetime.date(2017, 4, 12),
+            'start_date': datetime.date(2017, 1, 3),
+            'end_date': datetime.date(2017, 3, 30),
             'prefile_start_date': datetime.date(2016, 8, 1),
             'display_name': '2017 Regular Session',
             '_scraped_name': '2017 Regular Session',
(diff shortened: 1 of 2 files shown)


loandy:
Meant to leave a comment about this earlier.  I've thought about this kind of approach before, but I'm thinking that maybe an explicit flag in the metadata might be a better approach than inferring prefile status based on the date.

showerst:
Makes sense; so something like ['session_details]['session']['prefiling'] = True ? 

I'm a little concerned about the maintenance burden here, since we're then responsible for flipping that flag (and users need to git pull) same-day as the changeover to keep the data totally accurate, as opposed to the date approach, but on the other hand we avoid timezone hell and states that might be a little off on their dates. Thoughts?

loandy:
Yeah, there is a slight maintenance burden, but I think it's simpler to debug if something goes awry than the date approach.  I would tend to lean toward less magic, if possible.

jamesturk:
since this one is fairly out of date and we aren't in a prefile period anyway I'm going to wait until KY is ported to pupa to bring in prefiles

I'll leave the issue open so we don't lose track of the branch

as discussed in #1205, we'll add a prefiles flag to the pupa scraper, so there won't need to be a git pull to toggle prefiles but it is possible to turn off the prefile check by changing the task runner.  the idea of keeping date-based windowing on the prefiles is a good idea as well

jamesturk:
KY pupa has landed so this is unblocked if someone wants to tackle it

PR #1147, opened 2016-08-16
Al 2017

showerst:
Updated Alabama code for website changes, and to add 2016 special and 2017 sessions.

### DIFF
--- a/openstates/al/__init__.py
+++ b/openstates/al/__init__.py
@@ -24,7 +24,7 @@ metadata = {
             'name': '2015-2018',
             'start_year': 2015,
             'end_year': 2018,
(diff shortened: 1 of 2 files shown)


mileswwatkins:
@showerst, does this contain all of the AL changes that you made? Or have there been more since (or in a different `showerst/openstates` branch)?

showerst:
@mileswwatkins this is an older commit, but I don't see any updates on my fork, and I just ran it for 2017 and it seems to be completing fine, so should be good.

mileswwatkins:
Gotcha, yeah, wanted to see if I could start reviewing this right away. Will do!

mileswwatkins:
For me, this code seems to choke on the vacant seats in House Districts 58 and 67. Bills module looks solid, though! Can you add special-casing so that Legislator scraping works, then I'll review that output and merge this in?

PR #1149, opened 2016-09-14
Added support for special sessions in TN

showerst:
Modified the scraper to handle special sessions, and added the 2016 specials.


mileswwatkins:
Subsumed by https://github.com/openstates/openstates/pull/1192

PR #1150, opened 2016-09-15
UT Special

showerst:
Added 2 special sessions in UT from earlier this year.


mileswwatkins:
Subsumed by https://github.com/openstates/openstates/pull/1193

PR #1151, opened 2016-09-18
SD: change URLs to use www.sdlegislature.gov

cliftonmcintosh:
Requests to `legis.sd.gov` are re-directing to `www.sdlegislature.gov` and scraping will not complete. This is to change the URLs to the new one. Also, some div IDs have changed on the site, and scraping was not working because of those. I have corrected those as well.


mileswwatkins:
Thanks @cliftonmcintosh!

PR #1165, opened 2016-12-03
[WIP] Update SD for 2017 session

mileswwatkins:
Won't run while the list of 2017 bills is empty, needs to have the scraping of those confirmed to work.


mileswwatkins:
Scraping 2016's session bills fine, for now we'll assume it's fine with 2017's. Other data ingesting fine, AFAICT.

PR #1169, opened 2016-12-09
Added "Getting Started" instructions to the README

mshenfield:
This includes information on how to
* build and run the Docker development environment
* run any tests (just `il` for now)

It also explains how the project relates to `billy`, and that the Dockerfile entry-point  is the `billy-update` command.


jamesturk:
thanks for this Max! 

as an aside: we'll probably want to tweak the notes to not encourage people to run the tests here (the real tests are in the billy repo, we encourage scrapers to be written to "break early & often" as we'd much rather know that a scraper is broken than have it fail silently. over the years we've determined this doesn't play nicely w/ most people's test strategies, which is why the true testing is at the framework level and not here)

PR #1171, opened 2016-12-09
Add 2017 session metadata and late 2016 special sessions

showerst:
These are the late 2016/2017 sessions for states that only need an __init__.py change.


openstates/pa/__init__.py
```diff
@@ -74,6 +74,96 @@
...
+		'1821-1822 Regular Session',
+		'1819-1820 Regular Session',
+		'1817-1818 Regular Session',
+		'1815-1816 Regular Session',
+		'1813-1814 Regular Session',
+		'1811-1812 Regular Session',
+		'1809-1810 Regular Session',
+		'1807-1808 Regular Session',
+		'1805-1806 Regular Session',
+		'1803-1804 Regular Session',
```
jamesturk: this is nuts

mileswwatkins:
Linking this to the source ticket, for reference

https://github.com/openstates/openstates/issues/1170#issuecomment-266019493

Issue #1170, opened 2016-12-09
Create by-state PRs for substantially-rewritten states


mileswwatkins:
> Off the top of my head AL and MA have had the most substantial work. Many of the rest are just init.py updates.

@showerst, could you create by-state PRs to this repo for the few states that required substantial rewrites for 2017? Instead of handling a monolithic PR from `showerst/opentstates` to `openstates/openstates`, it'd be nice for @jamesturk and I to merge them in on a chunked basis.

The simplest way I can think of would be to copy-paste a state's directory from `showerst/openstates` into `openstates/openstates`, and branch and submit as a PR, but there might be a more git-y way to do this.

jamesturk:
If a state is only __init__.py changes feel free to put those PRs together
on a single branch, but yeah separate PRs would be super helpful for states
that had rewrites/etc.

mileswwatkins:
I'll take care of this tonight.

mileswwatkins:
Phew. I haven't code-reviewed these, just created the PRs. I'll work on review/merging as I get time this week.

@showerst mentioned that he'd like to work on this repo rather than his fork once this is all sync'd up. So let's make it happen soon!

PR #1184, opened 2016-12-19
KY: Update KY for 2017 with prefiled bills

mileswwatkins:

Includes the `prefiles` feature, from Tim's branch of Billy; we may not want this bit.


mileswwatkins:
Argh, dup of https://github.com/openstates/openstates/pull/1146

PR #1187, opened 2016-12-19
NJ: Skip errors adding bills

mileswwatkins:


mileswwatkins:
Doesn't look like this is necessary to run given the current set of prefiled bills, but adding it in anyway. Going to throw in a `self.warning()` for good measure.

PR #1189, opened 2016-12-19
NY: Remove reliance on Mongo

mileswwatkins:


showerst:
Just to explain this -- By default the NY scraper relies on the term_for_session helper function in billy, which assumes a mongodb backend. This inlines that function in a way that will work if the scraper is run independently.

PR #1190, opened 2016-12-19
SC: Add prefiled bills to scraping

mileswwatkins:


openstates/sc/bills.py
```diff
@@ -252,7 +254,7 @@ def scrape_details(self, bill_detail_url, session, chamber, bill_id):
     def scrape(self, chamber, session):
         # start with subjects
         session_code = self.metadata['session_details'][session]['_code']
-        self.scrape_subjects(session_code)
+        #self.scrape_subjects(session_code)
```
showerst: I commented this out in my branch because it's not needed for prefiles and slows down the scrape, it should be restored to fix previous-session scrapes.

PR #1192, opened 2016-12-19
TN: Allow special-session bill scraping

mileswwatkins:


mileswwatkins:
Hmm, odd that there are no bills listed on [the special session index](http://wapp.capitol.tn.gov/apps/indexes/SPSession1.aspx), for either 2015's or 2016's. But I see where you're coming from! And frankly there were probably bills there before when you wrote it, and the LIS just removed them?

showerst:
That prefile Index page was populated when I wrote the code; I guess they remove the bills later. That's weird though since the session hasn't started yet.

PR #1204, opened 2016-12-23
[WIP] NH: First pass at 2017 scraper

showerst:
NH: 2017 Scraper Update

WARNING: There's some low quality code in here (mostly a cut and paste from the legacy scraper, as they've returned to a similar system going forward), but I have to catch a plane so I figured it's better than nothing. Not all the bill URLs work, I'll email NH next week and figure out why some bill urls use the LSR and some don't.

If nobody else gets to it I'll refactor/clean it up in January.

The new NH files come from http://gencourt.state.nh.us/dynamicdatafiles/ instead of a ZIP file.

### DIFF
--- a/openstates/nh/__init__.py
+++ b/openstates/nh/__init__.py
@@ -20,7 +20,9 @@ metadata = {
         {'name': '2013-2014', 'sessions': ['2013', '2014'],
          'start_year': 2013, 'end_year': 2014},
         {'name': '2015-2016', 'sessions': ['2015', '2016'],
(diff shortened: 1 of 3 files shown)


jamesturk:
do we know if the legislator scraper should be working?  getting DB errors when i try to run

showerst:
@jamesturk I just did bills, but members are in a csv at http://gencourt.state.nh.us/dynamicdatafiles/ so it should be an easy one. I'll do it this week if nobody beats me to it.

PR #1219, opened 2016-12-30
[WIP] IN: Add 2017 session and temporarily stop version scraping

showerst:
Scraping votes/versions is causing the scraper to hang, maybe because the session is missing in the openstates api? 

This adds 2017 to the metadata and comments out the function that breaks the bill scraper.

### DIFF
--- a/openstates/in/__init__.py
+++ b/openstates/in/__init__.py
@@ -26,6 +26,8 @@ metadata = dict(
          'end_year': 2014, 'sessions': ['2013', '2014']},
         {'name': '2015-2016', 'start_year': 2015,
          'end_year': 2016, 'sessions': ['2015', '2016']},
+        {'name': '2017-2017', 'start_year': 2017,
+         'end_year': 2018, 'sessions': ['2017']},
         ],
     session_details={
         '2009': {'display_name': '2009 Regular Session',
@@ -45,6 +47,11 @@ metadata = dict(
             '_scraped_name': 'First Regular Session 119th General Assembly (2015)'},
         '2016': {'display_name': '2016 Regular Session',
             '_scraped_name': 'Second Regular Session 119th General Assembly (2016)'},
+       '2017': {'display_name': '2017 Regular Session',
+            '_scraped_name': 'First Regular Session 120th General Assembly (2017)',
+            'start_date': datetime.date(2017, 1, 9),
+            'end_date': datetime.date(2017, 4, 29)
+            },
         },
     feature_flags=['subjects', 'capitol_maps', 'influenceexplorer'],
     capitol_maps=[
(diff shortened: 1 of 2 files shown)


mileswwatkins:
> the session is missing in the openstates api

Yeah, since the IN documents are behind their API key, we created a scraper to re-host them ourselves (so we can easily process full-text, and offer document links on the API and website). See https://github.com/openstates/indiana-docs

@jamesturk, what do you think about this practice going forward? I'm fine with it. Maybe we could get @rshorey to take on bringing her scraper system back up? ;)

jamesturk:
Yeah IIRC this was running on a free Heroku instance and should be easy to
bring back up on our OpenStates free heroku we have for Slack invites.  I
didn't realize it was used in scraping, but I can probably try a heroku
deploy tonight or tomorrow if IN is otherwise ready

mileswwatkins:
Yeah, IN's essentially ready otherwise. Give the Heroku a shot!

jamesturk:
unfortunately Heroku won't let me use the old URL, so I just switched it to http://in-proxy.openstates.org/

I think it is working, let me know if anything seems off

rshorey:
I don't recall the doc part being used in scraping, but it's been a while.

@showerst, in general Indiana's site is really fragile. You have to use a combination of the API and the site, and the site is *really* easy to crash. I'm in the process of getting things set up on a new machine but if you want me to take on Indiana for 2017, I'm happy to do that. If so, is this branch in a good place for a handoff?

mileswwatkins:
@rshorey, you're cleared to go on this branch, AFAICT. I sent you the Indiana API key on Slack, too.

rshorey:
I uncommented the version scraping line.

But I encountered another issue, and I can't remember how we feel about this sort of thing. IN posted an inaccurate vote count [here](http://in-proxy.openstates.org/2017/rollcalls/MTIwLzIwMTcvY2FiaW5ldC9yZXNvbHV0aW9ucy9ob3VzZS9jb25jdXJyZW50L0hDMDAwMS9yb2xsY2FsbC9IQzAwMDEuMi5wZGYuaHRtbA==) (note that there is no one listed as not voting despite the claim that 1 person did not vote) which currently throws this error:

`AssertionError: Other vote counts (3) don't match count of actual votes (2)`

Is this the kind of thing that should be a warning instead of an error? Or should I manually skip that vote or something? I've been out of this game for a little while, sorry y'all.

jamesturk:
Hey @rshorey - yeah I think it'd be OK to switch those to warnings for now, maybe add a link to the actual bill that's broken in the comment so we can re-evaluate making those strict again.

Many states unfortunately have weird vote counts like this, looks like the IN scraper was trying to be proactive but not a whole lot else can be done besides relaxing the constraint here.

rshorey:
switched error to warning. @showerst does this all look OK to you?

showerst:
@rshorey looks good to me!

jamesturk:
👍

PR #1221, opened 2016-12-31
Use `Scraper.get` in `LXMLMixin.lxmlize`

### LINKED ISSUE #1214: Make `LXMLMixin.lxmlize` use scraper's `session`, rather than a raw `request.get()`
Otherwise it slows down `--fast` scrapes because it doesn't use cached responses.

mattjmcnaughton:
Fix #1214 

Convert the `LXMLMixin.lxmlize` method to use the `get` method defined
by the subclass of `billy.Scraper` for which `LXMLMixin` will serve as a
mixin, instead of `requests.get`. Using the scrapers `get` method uses
cached responses during `--fast` scrapes, which saves time.

Signed-off-by: mattjmcnaughton <mattjmcnaughton@gmail.com>

### DIFF
--- a/openstates/utils/lxmlize.py
+++ b/openstates/utils/lxmlize.py
@@ -14,11 +14,13 @@ class LXMLMixin(object):
             Element: Document node representing the page.
         """
         try:
-            response = requests.get(url)
+            # This class is always mixed into subclasses of `billy.Scraper`,
+            # which have a `get` method defined.
+            response = self.get(url)
         except requests.exceptions.SSLError:
             self.warning('`self.lxmlize()` failed due to SSL error, trying'\
-                'an unverified `requests.get()`')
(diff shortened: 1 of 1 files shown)


mattjmcnaughton:
First time working on the project, so any feedback would be appreciated. Thanks!

mileswwatkins:
Thanks @mattjmcnaughton! Will be able to give substantial feedback on larger PRs, but this looks good to me.

@jamesturk, can you think of any places where switching from `requests.get()` to `Scrapelib.session.get()` could break a scraper? I tried this for a couple currently-working states that use `lxmlize`, and didn't notice any nasty side-effects.

jamesturk:
I can't think of a good reason why it'd break unless something more serious was wrong, I'm good w/ this.

mileswwatkins:
Here we go!

PR #1227, opened 2017-01-02
NY: Update for 2017 session, allow missing-party members

mileswwatkins:


mileswwatkins:
@jamesturk, is `party` one of the keys that Billy uses when de-duping from new scrape to existing Mongo legislators? Might be a problem for NY in the next month.

PR #1230, opened 2017-01-04
Tim/wi/2017

showerst:
WI: Add the 2017 Session, and prevent the scraper from choking on missing vote pages.


mileswwatkins:
Breaking for me when parsing this House vote page: http://docs.legis.wisconsin.gov/2017/related/votes/assembly/av0001

Looks like they changed around the structure of those vote pages since last time. Even more inscrutable to scrape :|

mileswwatkins:
@showerst, looks like the committees pulled are all from 2015, though.

PR #1231, opened 2017-01-04
VT: Add filed bills to list of fetched bills

showerst:
We weren't grabbing all the VT bills because there's a stage before 'introduced' in their system.


mileswwatkins:
This guyyyyyy. What a catch.

PR #1232, opened 2017-01-04
MS: Add 20162E and 2017 sessions

showerst:
Adding 2017 regular and 2016 2nd Extraordinary sessions in MS.


showerst:
@jamesturk this is ready to go, do note that I added two sessions, one of which is passed.

PR #1265, opened 2017-01-17
Add docker-compose config.

jmcarp:
So that devs don't have to run a separate local mongo, and to make the quickstart Just Work.


jamesturk:
thanks, this looks great, definitely like not having to have people explicitly link in their local dir 👍

PR #1278, opened 2017-01-20
MD: Skip votes with no breakdown

showerst:
The MD Bill scraper was choking due to an action that looked like a vote, but wasn't.

### DIFF
--- a/openstates/md/bills.py
+++ b/openstates/md/bills.py
@@ -276,63 +276,65 @@ class MDBillScraper(BillScraper):
             chamber = 'lower'
 
         motion, unused_date = action_text.rsplit(' - ', 1)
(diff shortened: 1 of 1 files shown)


mileswwatkins:
Hmm, it looks like that action is actually a vote, right? http://mgaleg.maryland.gov/webmga/frmMain.aspx?pid=flrvotepage&tab=subject3&id=sb0028,s-0015&stab=02&ys=2017RS

The only problem was that the motion text didn't include a vote count. Could you check the logic in my alternate solution, which would capture this second-reading vote? https://github.com/openstates/openstates/pull/1281

mileswwatkins:
Superseded by https://github.com/openstates/openstates/pull/1281

PR #1297, opened 2017-01-24
Added apparent_encoding call to fixed unicode issue caused by a empty…

jonConerly:
… encoder string sent by the server


jamesturk:
thanks!

Issue #1269, opened 2017-01-18
MO: Retire 98th-session senators


Wizofgoz:
Example:
Paul LeVota is shown in the list:
https://openstates.org/api/v1//legislators/?state=mo&active=true&apikey=*******

and not in this list
http://www.senate.mo.gov/2017-senator-listing-2/

ladonkey7:
We checked in and heard that the MO upper chamber was up to date, but we're still seeing discrepancies in several districts – 4, 15, 19, 21, 23. For example, Eigel is now the senator for district 23 and Dempsey is still returned

jamesturk:
Confirmed that MO changed their Senator listing but left the old page up, we'll need to change the scraper to use the above http://www.senate.mo.gov/2017-senator-listing-2/

mshenfield:
This should be easier - http://www.senate.mo.gov/senators-listing/. It doesn't include Paul LeVota, and Eigel is 23, so I think it is up to date.

mileswwatkins:
The PR has been merged, only thing left is to remove the incorrect senators from the production database.

PR #1305, opened 2017-01-28
MO: Use new senate listings page

mshenfield:
Switch the scraper for the MO senate from http://www.senate.mo.gov/CurrentRoster.htm to http://www.senate.mo.gov/senators-listing/.

The relative position of the table on the new listing page is differentt, but the columns of the table, and the structure of the content within each column is exactly the same.

`if rowcount < 2:` has been changed to `if rowcount <= 2:`. This line of code, as indicated by the comments, is meant to skip the first two rows of the table, both of which are headers. But using strictly `<`, the second row has `rowcount=2` and is not skipped. I don't know if/how this functioned in the original scrape, but have confirmed that with this change, the first row "Dan Brown" is scraped correctly, and without the scrape fails.

Verifying the output, I noticed that the photo_url was being returned as http://www.senate.mo.gov/17web/wp-content/uploads/2015/11/301px-Seal_of_the_Senate_of_Missouri_svg-e1448041814274.png, which links to the state seal.  I've updated the photo_url selector to be more specific, and it now correctly returns the member photo url.

### DIFF
--- a/openstates/mo/legislators.py
+++ b/openstates/mo/legislators.py
@@ -10,7 +10,9 @@ class MOLegislatorScraper(LegislatorScraper):
                             'Jefferson City, MO 65101')
     # senators_url = 'http://www.senate.mo.gov/{}info/senalpha.htm'
     # ^^^^^^^^^^^ pre-2013 URL. Keep if we need to scrape old pages.
-    _senators_url = 'http://www.senate.mo.gov/CurrentRoster.htm'
+    # _senators_url = 'http://www.senate.mo.gov/CurrentRoster.htm'
+    # ^^^^^^^^^^^ pre-2017 URL. Keep if we need to scrape the old pages.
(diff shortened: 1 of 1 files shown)


mshenfield:
This should close #1269. For comparison, here is a diff between the output of the old scraper and the new scraper for Dan Brown.

```
17c17
<         "url": "http://www.senate.mo.gov/CurrentRoster.htm"
---
>         "url": "http://www.senate.mo.gov/senators-listing/"
31c31
<     "photo_url": "http://www.senate.mo.gov/17web/wp-content/uploads/2015/11/301px-Seal_of_the_Senate_of_Missouri_svg-e1448041814274.png",
---
>     "photo_url": "http://www.senate.mo.gov/17web/wp-content/uploads/2015/11/Brown-200x300.jpg",
```

mileswwatkins:
Thanks, @mshenfield!

Issue #1276, opened 2017-01-20
NM: Fix vote PDF parsing


mileswwatkins:
Right now, it appears that NM Senate vote PDFs aren't parsed correctly; [I've turned that off temporarily](https://github.com/openstates/openstates/pull/1188/commits/1207f3020f771baa35520c635cc507918e98c15b). And, FWIW, House vote PDFs only get their totals parsed, not their voting members.

jonConerly:
I think I have a solution for this one that I can put a PR in for if you are not working on it already.

mileswwatkins:
Fire away, @jonConerly!

jonConerly:
I put in the PR for my change, this ended up being a can of worms.  I believe my method is less likely to break over time compared to the old method of pdf to text conversion; one primary reason for this is that it appears that the vote tally is being filled in by hand base on some metadata.  The plotted grid parser was also able to start parsing the house votes since their vote sheet is similar in concept, so I added that method in as well.

Added a tests folder as well to spot any future issues, any new pdfs added to the tests/testdate folder will be analyzed automatically if named correctly(*_senate.py and *_house.pdf).  nosetests was able to find and run the tests successfully.  My PR did not delete the old tests for the vote parsing regex so that might need to be cleaned up depending on what parts of this PR are accepted.

jamesturk:
ran locally and verified things looked good, sorry for the delay on this one

PR #1308, opened 2017-01-29
NM: Rewrote vote parser #1276

jonConerly:
Fixed the senate parser by moving away from the old method of pdf to text method; this was having problems with the differently formatted vote roles depending on the year/chamber.  Moved to a plotted x.y system to match votes.

Also added house vote parsing and tests


jonConerly:
Accidentally left 3 lines of debug code in the original PR, 1 line that would have caused a bug and 2 lines of comment coded debug statements.

jamesturk:
does this fully replace the vote parsing in bills.py?  if so we should probably remove that as well so we aren't double scraping votes

jonConerly:
I did find a case where NM had a bad link to a vote URL, handling that error case now.

This does replace the vote parsing, I don't follow the double scrape comment though.  I didn't see where the pdfs were being scraped twice, could be missing it though.

PR #1322, opened 2017-02-02
IN nits

jmcarp:
I was looking at a few outstanding issues and noticed some nits, fixed here:
* Pass API keys through from host environment
* Raise appropriate error on failed request


jamesturk:
looks good, thanks!

Issue #1325, opened 2017-02-02
TX: Committee list and membership is scraping from the old session's URL?


grgcombs:
State: TX

The committees and the committee membership pages are being scraped from URLs for the 84th session instead of URLs for the current 85th session.  At least this is what the committees.sources.url query parameters seem to indicate...

For each committee's membership, the scraper is currently looking at URLs like this (notice the LegSess parameter is **84R**, meaning 84th Regular Session):

    http://www.capitol.state.tx.us/Committees/MembershipCmte.aspx?LegSess=84R&CmteCode=C500

It should be scraping membership listings for the current/85th session using URLs like this (LegSess=85R):

    http://www.capitol.state.tx.us/Committees/MembershipCmte.aspx?LegSess=85R&CmteCode=C500

Note that at the time of this writing, the House still hasn't published any committee memberships -- those will probably come out sometime in the next month or so.  However, the Senate's committee memberships are updated and published.

jonConerly:
Are you running this from the production scrapers or have local changes not in master by chance?  My local branch is pulling 85R for upper(works) and lower(exception):
18:38:50 DEBUG requests.packages.urllib3.connectionpool: "GET /Committees/MembershipCmte.aspx?LegSess=85R&CmteCode=C020 HTTP/1.1" 200 9581

Scraping the lower committee membership pages throws an AttributeError exception because the committee scraper does not check to see if the membership table exists before trying to read it and/or lxmlize.get_nodes does not check the validity of the object it receives before trying to use the xpath attribute.  I could see a local change being made to keep the scraper from crashing out on production, although this looks like more of a bug that the scraper can't handle a blank membership page.

If you aren't using a prod copy, I don't see how the version could be set to 84 in code.  The committees' list page is hard coded to:  http://www.capitol.state.tx.us/Committees/CommitteesMbrs.aspx?Chamber=H which defaults the to the 85th session; all the committee membership links are scraped from the page without any metadata substitution.

grgcombs:
I wasn't running it myself, just seeing the issue in the results from the API at https://openstates.org/api/v1

showerst:
@jonConerly just confirming your comment that the scraper is fine. 

@jamesturk how have you dealt with source urls changing in the past, did you just blow that URL away when you rescrape, or leave it and if the state page changes append the new one?

jamesturk:
[...]

PR #1329, opened 2017-02-04
Fix TX lower committee exception #1325

jonConerly:
#1325  There are two commits in this pull request, both can independently fix the lower committee scraper's exception.  Pick one or both depending on how broad you feel the fix for this should go.

The commit to committees.py will check if the members page does not contain the members table and skip the member enumeration routine, this avoids calling get_nodes with a None type object and contains the change to the TX module.

The other commit lxmlize addresses another bug which was actually throwing the exception for this issue.  lxmlize.get_nodes does not check the base_node for sanity before trying to access its attributes.  Since the function is suppose to return a list of elements, I am returning an empty list instead of throwing an exception(as well as logging a warning).  Looking around the other state modules, most things are not checking the validity of what is being passed into get_nodes.  This fix will either help prevent future crashes or make data collection errors less obvious depending on how strictly the output of Billy is being monitored.

### DIFF
--- a/openstates/tx/committees.py
+++ b/openstates/tx/committees.py
@@ -33,32 +33,36 @@ class TXCommitteeScraper(CommitteeScraper, LXMLMixin):
             # Capture table with committee membership data.
             details_table = self.get_node(committee_page,
                 '//div[@id="content"]//table[2]')
-            # Skip the first row because it currently contains only headers.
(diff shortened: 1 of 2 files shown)


jamesturk:
Thanks, pulled in the first commit.

I think other scrapers are likely to silently fail if we protect them from bad HTML, and since we much prefer a loud crash to a silent failure I'm going w/ just the first half of this.

Issue #1330, opened 2017-02-04
OR: Suffixes field for ORL000097 contains trailing garbage


ronisaacson:
State: OR

Please see https://openstates.org/api/v1/legislators/ORL000097/. The suffixes field includes a couple of extraneous non-printing characters at the end. The full_name field is also affected. Here's a bit of the hex dump:

```
00000590: 3a 20 22 63 61 70 69 74  6f 6c 22 2c 20 22 65 6d | : "capit ol", "em
000005a0: 61 69 6c 22 3a 20 22 53  65 6e 2e 48 65 72 6d 61 | ail": "S en.Herma
000005b0: 6e 42 61 65 72 74 73 63  68 69 67 65 72 40 6f 72 | nBaertsc higer@or
000005c0: 65 67 6f 6e 6c 65 67 69  73 6c 61 74 75 72 65 2e | egonlegi slature.
000005d0: 67 6f 76 22 7d 5d 2c 20  22 73 75 66 66 69 78 65 | gov"}],  "suffixe
000005e0: 73 22 3a 20 22 4a 72 2e  e2 80 8b 22 7d          | s": "Jr. ..."}

```

Notice the extra characters after Jr. on the last line (e2 80 8b).

jamesturk:
will manually clean up after merge

PR #1335, opened 2017-02-05
Fix unicode whitespace character removal method for OR legislators

jonConerly:
#1330  The method that was attempting to remove non-printable characters did not handle the case where there non-printable character was directly attached to legislator's name.  Inspecting the names per character now instead of by space delimited words.

### DIFF
--- a/openstates/or/legislators.py
+++ b/openstates/or/legislators.py
@@ -1,6 +1,7 @@
+import re
+import string
 from billy.scrape.legislators import LegislatorScraper, Legislator
 from openstates.utils import LXMLMixin
-import re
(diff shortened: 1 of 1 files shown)


jmcarp:
Is it safe to assume that there won't be any printable characters that aren't in `string.printable`?

```
In [8]: 'ö' in string.printable
Out[8]: False
```

jonConerly:
...no it is not safe to assume that, good catch!  The page is returning a utf-8 encoding which is in conflict with this change as string.printable appears to only contain ascii-128.  Closing this PR for now.

Issue #1248, opened 2017-01-10
NH: needs committees & photos


jamesturk:
disabled these for 2017

jmcarp:
Is the idea to pull from http://gencourt.state.nh.us/dynamicdatafiles/ instead of whatever lives at 216.177.20.245?

showerst:
Yes, I think we're treating that SQL database as deprecated.

jamesturk:
confirmed comms look good in DB, waiting on #1374

PR #1336, opened 2017-02-06
Parse NH committees from dynamic data.

jmcarp:
Starting on #1248.

* [x] Parse committees from http://gencourt.state.nh.us/dynamicdatafiles/
* [x] Parse committee membership from committee pages

Like @showerst said, it looks like we'll need to parse membership from committee pages. Will take a look at that this week.

### DIFF
--- a/openstates/nh/bills.py
+++ b/openstates/nh/bills.py
@@ -1,20 +1,21 @@
-import os
 import re
-import zipfile
 import datetime as dt
 
 from billy.scrape.bills import Bill, BillScraper
 from billy.scrape.votes import Vote
 
+from openstates.nh.legacyBills import NHLegacyBillScraper
+
 
 body_code = {'lower': 'H', 'upper': 'S'}
-bill_type_map = {'B': 'bill',
-                 'R': 'resolution',
-                 'CR': 'concurrent resolution',
-                 'JR': 'joint resolution',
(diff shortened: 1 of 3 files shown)


jmcarp:
Updated to parse committee memberships from committee pages. I noticed that the previous version of this parser only recognizes chair and member roles, but not vice chair--should we preserve that logic here?

mileswwatkins:
@jmcarp, vice-chair and other committee positions are totally fair-game; scrape 'em!

jmcarp:
This seems to be working as expected. Anybody have time to review?

jamesturk:
checked it out locally and it looked good, thanks!

jmcarp:
@jamesturk @mileswwatkins: if you all are in contact with NH, we could also ask them to add an association table between committees and legislators like in the legacy data to avoid some of the work added here.

showerst:
I think i may have found a table with this data -- http://www.gencourt.state.nh.us/downloads/members.txt according to http://www.gencourt.state.nh.us/downloads/ it was updated Feb 7. Are there any members with more than 2 committee assignments?

jmcarp:
@showerst: I looked at that table earlier, but didn't see committee roles (chair, vice chair, clerk).

showerst:
@jmcarp ah, got it. So close!

PR #1338, opened 2017-02-06
OR: Handle valid unicode characters as well as unicode encoded names 1330

jonConerly:
#1330 The method that was attempting to remove non-printable characters did not handle the cases where some non-printable characters were not removed via strip(), due to this [issue](https://bugs.python.org/issue13391).  The character being reported in this issue was the unicode zero-width space, this PR removes all control-character from the name and leaves normal unicode characters in place.

### DIFF
--- a/openstates/or/legislators.py
+++ b/openstates/or/legislators.py
@@ -1,6 +1,8 @@
+# -*- coding: utf-8 -*-
+import re
+import unicodedata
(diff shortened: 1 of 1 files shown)


jamesturk:
this looks great, thanks.  definitely something we should consider as a general purpose filter

PR #1341, opened 2017-02-07
MO committee scraper  link and session bug fixes #1319

### LINKED ISSUE #1319: MO Committee Members are incorrect
https://openstates.org/api/v1//committees/MOC000028/?state=mo&apikey=xxxxx
vs
http://house.mo.gov/CommitteeList.aspx (Budget Committee)

Some don't even show in the Legislators list such as:
Sue Allen
Eric Burlison
Lincoln Hough
and Caleb Jones

jonConerly:
#1319  Fixed bugs with session year looping and link identification.  

Two issues were keeping the scraper for running.  The first was that the link to "Assigned bills/Bills"  is cased differently several times on the [page](http://www.senate.mo.gov/standing-committees/), fixed the logic to handle the oddly mixed case links.  The other issue was that the current logic was checking for future terms in session_details, left most of that logic in place and just added the sanity check.

Cleared up some linting errors as well to make it flake8 compliant.

### DIFF
--- a/openstates/mo/committees.py
+++ b/openstates/mo/committees.py
@@ -1,8 +1,6 @@
-import datetime as datetime
+import datetime
 import re
 import lxml.html
-import xlrd
-import os
 from billy.scrape.committees import CommitteeScraper, Committee
 from openstates.utils import LXMLMixin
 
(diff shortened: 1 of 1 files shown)


jonConerly:
Turns out somebody already put in a PR for the same change before my push.  Closing.

PR #1374, opened 2017-02-14
Restore NH photos.

jmcarp:
This makes the scraper take longer to run, since we're pulling every house member page to figure out their seat number. Anybody have a better idea? Also, I'm wondering if we can ask the folks who provide the NH data to restore the employee ID to the table so that we can avoid this step eventually.


jamesturk:
@mileswwatkins am I remembering correctly that you spoke w/ NH earlier?  if we have a decent contact there it is worth trying to get that employee ID back.  if not, I'm fine w/ this approach as we have to do this extra step in a lot of states

mileswwatkins:
@jamesturk, I have not talked to NH at all. (I did talk with _VT_'s contracted devs when I worked at Sunlight.)

jamesturk:
Sent an email to a NH contact today, will give it a couple of days & merge if we don't hear back

PR #1376, opened 2017-02-14
Expose billy data and cache.

### LINKED ISSUE #1371: Link Docker's `data` (and maybe `cache`) directory to local file system
> @mshenfield [12:47 PM] I think we could make it easier by changing some of the configuration in `docker-compose` or the docker files so that it would write `data` to the openstates repository directory.

This is also something that @jamesturk and I have talked about internally. Let's get the `docker-compose` file updated so that users can see the `data` directory locally, and more easily hand-inspect the scraped JSON.

Maybe also link the `cache` directory, too, so that it's easier to throw out old scrapes? (But also fine if this is considered a bad idea, or if another solution to flushing old scrapes is suggested.)

jmcarp:
Resolves #1371.

For backwards compatibility, we could only set the data and cache paths if provided in the environment and pass them from docker-compose, but IMO it makes more sense to locate data, cache, and manual data in the same place.

### DIFF
--- a/billy_settings.py
+++ b/billy_settings.py
@@ -1,24 +1,20 @@
 import os
 
-from os.path import abspath, dirname, join
-
 import sys
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 # This shims around folks that have openstates/ on virtualenv's .pth, but
 # not the root. This throws openstates.utils off, and isn't worth fiddling
 # that much with.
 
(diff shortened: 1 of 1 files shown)


mileswwatkins:
Thanks! @jmcarp, can you also quickly update the docs accordingly to reflect these new directories, in https://github.com/openstates/documentation (which powers http://docs.openstates.org)?

PR #1381, opened 2017-02-16
Run mysql for ca scraper with docker-compose.

jmcarp:
Note: optionally running the CA downloads in `__init__` feels like a hack. Anybody have a better idea? Maybe we expose another command in docker-compose to load the CA database?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,6 @@ services:
     environment:
     - BILLY_MONGO_HOST=database
     - MYSQL_HOST=mysql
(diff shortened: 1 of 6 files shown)


showerst:
+1 for the idea of exposing a second command so that you'd have more fine grained control of when you want to (re)load the database. 

They only update the sources files daily, IIRC.

jmcarp:
Updated.

jamesturk:
This makes sense to me, definitely beats the old vagrant way of doing things.

PR #1382, opened 2017-02-16
Add ipdb to docker image.

jmcarp:
So that I can use the `--ipdb` flag.


jamesturk:
I'm torn on if we should separate out requirements-dev.txt, the Dockerfile is used in production (unlike the docker-compose file) so in practice there won't really be a difference and there's no real harm in installing it in prod despite not needing it.

Anything you think it really buys us?

jmcarp:
> Anything you think it really buys us?

Probably not, just force of habit. I just moved it back to requirements.txt.

PR #1383, opened 2017-02-16
WI: Ignore january special session with no bills, fix 2017 end date

showerst:
WI Added their january special to the website which broke the bills scraper. 

There don't seem to be any bills [here](http://docs.legis.wisconsin.gov/2017/proposals/JR7/) -- given the dates i'm wondering if they didn't make it to the website or if they just got rolled into the Regular.

If they do post the bills back later, let me know and I'll add it as a real session.

### DIFF
--- a/openstates/wi/__init__.py
+++ b/openstates/wi/__init__.py
@@ -134,6 +134,7 @@ metadata = {
                       'events', 
                       'influenceexplorer'],
     '_ignored_scraped_sessions': [
+        # The WI Website has multiple special sessions in the system with no data.
         'January 2017 Special Session',
         'February 2015 Extraordinary Session',
         '2007 Regular Session', 'April 2008 Special Session',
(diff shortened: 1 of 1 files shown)


openstates/wi/__init__.py
```diff
@@ -134,6 +134,7 @@
                       'events', 
                       'influenceexplorer'],
     '_ignored_scraped_sessions': [
+        'January 2017 Special Session',
```
jamesturk: mind adding a comment to explain why this was skipped, I figure this will raise a red flag at some point since it looks like a relatively current session being skipped (just seeing this I'm curious about Feb. '15 myself)

showerst: Done. If you'd rather we put it in as a session (which works, but it returns no bills) i'm fine with that. 

This is kind of a weird one, it ended a while ago so i'm not sure if they're just late getting the data or never going to add it. 

I'll try to remember to keep an eye on http://docs.legis.wisconsin.gov/2017/proposals/JR7/ to see if they backfill it.

PR #1394, opened 2017-02-19
[WIP] Start pupa va scraper.

jmcarp:
Based on https://github.com/openstates/openstates/tree/pupa-fl; thanks @jamesturk! There's a lot left to do here, but I thought I'd submit a WIP for visibility, and in case anybody's interested in reviewing. A few questions:

* How (if at all) does pupa learn about sessions? Should a pupa scraper scrape *all* sessions?
* Is `Person.add_membership` the correct way to describe committee memberships? It looks like this method doesn't accept term or chamber--how should those be described?


jmcarp:
This still needs work but is pretty close to the functionality of the billy scraper, so it would be helpful to get a review when there's time. We'll also want to merge https://github.com/opencivicdata/pupa/pull/261 or similar to keep rate limiting.

jmcarp:
A few questions left here:
* Is Person.add_membership the correct way to describe committee memberships? It looks like this method doesn't accept term or chamber--how should those be described?
* Is there a pupa equivalent to the summary field set on bills in billy?

Also, anybody have time to review?

@jamesturk

jamesturk:
Sorry for the delay, I'll review by this weekend.

openstates/va/__init__.py
```diff
@@ -1,185 +1,79 @@
...
     ]
-}
 
+    ignored_scraped_sessions = [
+        '001', '011', '012', '021', '031', '041', '042', '043', '051', '061',
+        '062', '071', '081', '082', '083', '091', '092', '941', '942', '943',
+        '951', '961', '971', '981', '982', '991'
+    ]
+
+    def get_organizations(self):
```
jamesturk: we'll have to make sure this gets updated, on the pupa-nc branch I put in a new generic get_organizations template that may be a useful starting point

openstates/va/people.py
```diff
@@ -0,0 +1,107 @@
...
+        for com in item.xpath('//ul[@class="linkSect"][1]/li/a/text()'):
+            self.obj.add_membership(com)
+
+
+class SenateDetail(MemberDetail):
+    role = 'Senator'
+    chamber = 'upper'
+
+
+class DelegateDetail(MemberDetail):
```
jamesturk: I like this

setup.cfg
```diff
@@ -1,2 +1,3 @@
 [flake8]
 max-line-length = 99
+ignore = E302
```
jamesturk: just curious- is this because of the defining of regex's before functions?  I hadn't seen that pattern before

openstates/va/bills.py
```diff
@@ -1,90 +1,109 @@
...
+            bill = Bill(bill_id, self.kwargs['session'], desc,
+                        chamber=chamber, classification=bill_type)
+
+            bill_url = link.get('href')
+            sponsor_url = BASE_URL + URL_PATTERNS['sponsors'].format(
+                self.kwargs['session'],
+                bill_id.replace(' ', ''),
+            )
+
+            list(self.scrape_page_items(BillSponsorPage, url=sponsor_url, obj=bill))
```
jamesturk: I see why this is going on but these could probably use a comment just so it is clear what is going on here  (and/or we should come up with a less hacky way for spatula to handle this case)

jamesturk:
And regarding your questions:
* I think bill.add_abstract is what you're looking for
* add_membership is the right way, terms are gone (they were always sort of fiction in billy, so we're aiming to add begin/end dates instead of fabricating 2-year terms) I think adding the committee by name should work but there is the case of two committees with the same name in both chambers, I'll need to dig in and see how we handle this in pupa as I don't remember offhand but it looks like this won't be an issue in VA

jmcarp:
Thanks for taking a look @jamesturk! I backfilled the billy metadata and adding some missing properties. This should be ready for more review when there's time.

jamesturk:
Looks good to me, I'm giving this a run locally and using the (very rough) billy2pupa/compare.py script to double check any output differences.

btw, Have you had an issue when running in fast mode where VA gives back partial pages?

jmcarp:
I haven't used fast mode here, but it wouldn't be surprising if it caused problems. I *think* VA has a rate limit, which is why the module needs to update `SCRAPELIB_RPM` and define an `accept_response` method to reject rate-limited responses.

jamesturk:
found some shortcomings in the conversion script and I'm working through those now but this looks like it is almost ready to go, I've created a pupa-va branch that I'm testing on (& made a couple of minor changes)

jamesturk:
We'll need to figure out the best way to surface committees from this scraper, billy let you add_role to arbitrary names and then created the committees, but in pupa you have to create Organization objects.

@jmcarp any thoughts on how the people scraper could best be changed to yield back Committee objects as well?  glad to talk through potential pupa.scrape changes for convenience

jmcarp:
Updated to explicitly yield committees from people scraper--wdyt @jamesturk ?

jmcarp:
BTW, it would be good to get https://github.com/opencivicdata/pupa/pull/261 ready around the time we merge this to make sure we aren't hitting VA rate limits.

Issue #1217, opened 2016-12-30
Senator Emails are mailforms - CT State Legisaltors


desireestephens:
I'm new to this but want to learn. It looks like Connecticut data is not surfacing actual email addresses for some of their representatives. Here it is: https://data.ct.gov/Government/Legislator-Database/rgw6-bpst/about

19 of senators are rendering a mailform as their email address. Whats the solution for this? How can I help? I've confirmed their emails exist with the syntax adopted by the rest of the senate emails. first.last@cga.ct.gov

mileswwatkins:
Awesome, Desiree, thanks for the info. Are you able to set up [Billy](https://github.com/openstates/billy) and run the Connecticut scraper locally (either with or without Docker)? If so, then we can work on how you'd tweak the scraper to handle these cases.

Check out the [documentation](https://billy.readthedocs.io/en/latest/index.html), and let me know if you have questions.

mattkenney:
Hi, new here, hoping to help.

I ran the CT legislator scraper and it raised an error on a invalid email address. The ct.gov CSV file lists a bit over a dozen state senators with contact form URLs in the email column. The scraper currently sets a legislator's email to None if the email value ends with "mailform.php" and continues, but one senator's contact form URL ends with "-contact". (I tried one of the "mailform.php" URLs and it redirected to a contact form ending with "-contact"; looks like the CT senate Democrats switched from PHP to .NET.)

I suggest changing the skip test so the CT legislator scraper can complete, i.e.

```python
-    if not email or email.endswith('mailform.php'):
+    if not email or email.startswith('http://') or email.startswith('https://'):
```

I can submit a pull request with this change if wanted.

Longer term, not sure if "contact form URL" should be added to the data model for legislators who do not have (or do not publish) an official email address.

jamesturk:
A PR with that change would be great, thanks Matt!

If you want you can add contact_form as an extra field (just set it on the legislator object), the scrapers will ingest it as an extra field  and it'll be available via the API at least

mattkenney:
OK, PR: https://github.com/openstates/openstates/pull/1395

Did add the contact_form as an extra field.

PR #1395, opened 2017-02-20
Senator Emails are mailforms - CT State Legisaltors #1217

mattkenney:
Re https://github.com/openstates/openstates/issues/1217

CT Legislator scraper was raising an error on an invalid email address. The scraper was trying to skip contact form URLs in the email field, but looks like a new one was added to the data that looks different from the ones that were there before.

This patch handles all the contact form URLs found in the email field, putting None for email address if the value present is actually a contact form URL, and putting those contact form URLs into an extra field, "contact_form".


jamesturk:
ran locally, looks great, thanks!

PR #1399, opened 2017-02-20
FL 2017 pupa scraper and necessary changes to run it

jamesturk:
this will be the first pupa scraper merged in, it can be run w/ 

``docker-compose run scrape fl people``

old scrapers continue to work via

``docker-compose run openstates nh``

New Stuff:

pupa2billy - convert pupa scrape JSON to old format using billy.scrape internals'

scrape.sh - a helper script for running pupa scrapers and getting a billy import

- runs scrape-only version of pupa scraper
- runs pupa2billy converter that converts pupa JSON to billy JSON
- runs a billy-update --import --report that uses the converted billy JSON'

Dockerfile-pupa - a pupa-ready Dockerfile that has a billy & pupa virtualenv


jamesturk:
after merging some simple Py3 fixes, it may be possible to run this all in a single (Python 3) virtualenv and avoid the double-venv dance that this does.  it'll be a sort of weird situation as I don't think billy.web will ever be Py3 compatible, but billy.{scrape,import,report} are all we care about here

jamesturk:
in a bit of a deviation from the norm, I'm going to run this branch in production for a short period, if all goes well I'll merge this in but definitely still open to feedback from anyone that is checking out pupa (cc @fgregg @divergentdave @jmcarp @mileswwatkins @showerst )

PR #1406, opened 2017-02-21
NH Legislators: add phone

mattkenney:
issue https://github.com/openstates/openstates/issues/1402

NH Legislators: add phone to legislator records


jmcarp:
How about adding email too?

mattkenney:
I think the legislator objects are already being populated with email, but there are email properties with null values in the office objects within the legislator objects. I can copy or move the email to the office object - probably want a separate issue/pull request for that to keep things organized.

PR #1436, opened 2017-02-28
Run openstates.org with compose.

jmcarp:
Depends on https://github.com/openstates/openstates.org/pull/45.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,13 @@ services:
     - INDIANA_API_KEY
     volumes:
     - .:/srv/openstates-web
+    depends_on:
+    - database
+  openstates.org:
+    image: openstates/openstates.org
+    command: python manage.py runserver 0.0.0.0:8000
+    environment:
+    - MONGO_HOST=database
     ports:
     - "8000:8000"
     depends_on:


mileswwatkins:
@jmcarp, two things, in trying to get parity with the setup in https://github.com/openstates/openstates.org/blob/master/HACKING.md:

- Does the `SECRET_KEY` environment variable need to be set when running OS.org?
- Do we need to have an easy way to pass a settings path to Django? Probably not, because I wouldn't expect this compose file to be used when actually doing OS.org development, but a thought.

I'm also happy to ignore these two points for now and just merge, since Billy and the associated front-end will only be around a few more months.

jmcarp:
* Looks like `SECRET_KEY` has a default in debug mode, so users won't have to set it unless they want to. We could explicitly pass that environment variable through, or instruct users to set relevant environment variables using `COMPOSE_` host variables or a `.env` file: https://docs.docker.com/compose/environment-variables/. WDYT?
* I'm wondering if those docs are out of date, because there isn't an `openstates.settings.dev` module in the repo. Anyway, if users do need to pass a custom settings module, maybe we could map a local path to a subfolder of `/openstates`, but agreed that we might not need to worry about this just now--my main interest here is just being able to look at the data I've scraped locally in the web interface.

BTW, I set up the compose config assuming that the docker image for openstates.org will live at dockerhub.com/r/openstates/openstates.org. If that's all right with you, can we get that set up on dockerhub before merging this?

jamesturk:
Just added a Dockerhub hook for this (building now) and I think those
instructions are out of date- I think this looks fine for now for local
testing purposes & we can tweak as needed if we did want people developing
& having more control of the environment (which is less likely b/c we're
aiming to kill this version off sooner than later)

PR #1462, opened 2017-03-05
Handle missing session type.

jmcarp:
Because some states don't currently include classifications in session details.


jamesturk:
ah good catch, we might want to make it warn about these as I think we need them in pupa

jmcarp:
Updated, wdyt?

Issue #1451, opened 2017-03-03
MT: committee scraper needs update for 2017


jamesturk:
State: MT

might be an easy one, needs committee URL updated & to be tested w/ new page

jamesthechamp:
hey, 
for 2017 url for committees i found were 
house
http://leg.mt.gov/content/Committees/Session/2017-house-committees.pdf
senate
http://leg.mt.gov/content/Committees/Session/2017-senate-committees.pdf
i had send pr https://github.com/openstates/openstates/pull/1465 with this urls.
please have a look at it

pmathur5k10:
Is this issue closed or shall I look into this ??

PR #1465, opened 2017-03-06
#1451 mt updated for 2017

jamesthechamp:
hey,
this is for issue #1451 
following are urls for montana house and senate committees which i added in scraper in this url please have a look at it
http://leg.mt.gov/content/Committees/Session/2017-house-committees.pdf
http://leg.mt.gov/content/Committees/Session/2017-senate-committees.pdf


jamesturk:
thanks!

Issue #1450, opened 2017-03-03
MO: vote scraper is failing for 2017


jamesturk:
State: MO

Missouri vote scraper fails on some complex PDF scraping

jamesthechamp:
hey,
i had send a pr #1471  for this issue 
please have a look at it.

PR #1471, opened 2017-03-08
#1450 MO bug solved

jamesthechamp:
this is for issue #1450 
please have a look at it


jamesturk:
thanks, ran this locally & looks good 👍

Issue #1447, opened 2017-03-03
ID: committee page has moved


jamesturk:
State: ID

https://legislature.idaho.gov/senate/committees.cfm no longer exists

jamesthechamp:
hey, this is shivansh bajaj from india, i am a full stack developer want to contribute and participate for gsoc 2017 under openstates. I successfully build openstates application. I would love to work on this issue
i saw that 
https://legislature.idaho.gov/senate/committees.cfm https://legislature.idaho.gov/house/committees.cfm pages are removed but we can get this data from 
https://legislature.idaho.gov/committees/senatecommittees/
https://legislature.idaho.gov/committees/housecommittees/
to save committees data but here scraper need be changed accordingly to new pages html
can i start working on this ?

jamesturk:
hi, thanks for your interest- this would be a great issue to take a look
at. any help would be greatly appreciated!

jamesthechamp:
hey, @jamesturk i had send a pr https://github.com/openstates/openstates/pull/1458 relate to this issue

PR #1472, opened 2017-03-08
#1447 ID: file changed issue solved

jamesthechamp:
this is for issue #1447


jamesturk:
ran & looked pretty reasonable locally, thanks for the cleanup!

Issue #1474, opened 2017-03-08
RI: "Dictrict Office" typo


mshenfield:
State: Rhode Island

The legislator District Office is mispelled "Dictrict Office".

[Example](https://openstates.org/ri/legislators/RIL000139/marvin-l-abney):

<img width="265" alt="screen shot 2017-03-07 at 11 48 11 pm" src="https://cloud.githubusercontent.com/assets/4812055/23691886/fc3f8800-0390-11e7-8266-68e11e8d983f.png">

pmathur5k10:
@mshenfield Sir I would like o work on this error. Can you explain this to me in brief as I am new to this org. Thanks !

PR #1476, opened 2017-03-08
issue #1474 : typo fixed

derekacosta:
issue #1474 : fixed


jamesturk:
hey thanks for this but unfortunately #1474 made it in just before this and also tackled the image issue so I went w/ that one, really appreciate it and sorry this didn't make it in

PR #1488, opened 2017-03-10
CO: Allow versions without publish dates

showerst:
CO Bill scraper was dying because they published a version without a pubdate.

### DIFF
--- a/openstates/co/bills.py
+++ b/openstates/co/bills.py
@@ -135,7 +135,10 @@ class COBillScraper(BillScraper, LXMLMixin):
 
         #skip the header row
         for version in versions:
-            version_date = version.xpath('td[1]/text()')[0].strip()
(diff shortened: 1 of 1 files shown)


openstates/co/bills.py
```diff
@@ -135,7 +135,10 @@ def scrape_versions(self, bill, page):
 
         #skip the header row
         for version in versions:
-            version_date = version.xpath('td[1]/text()')[0].strip()
+            if version.xpath('td[1]/text()'):
+                version_date = version.xpath('td[1]/text()')[0].strip()
+            else:
+                version_date = 'None'
```
jamesturk: should this be None?

jamesturk: ah I see how this works, thought it was saving this as a date.  fine by me

PR #1494, opened 2017-03-12
Pupa mi

jmcarp:
* [x] people
* [x] committees
* [x] bills
* [x] events


jmcarp:
@jamesturk: I didn't see an obvious pupa method corresponding to `Event.add_related_bill` in billy. I see that `EventAgendaItem` defines an `add_bill` method--do we need to add an agenda item to attach a related bill to an event?

jamesturk:
hmm, yeah I think that'd be the only way right now, are you going to have to make an agenda item up to do this?

jmcarp:
Updated to add related bills via agenda items.

jamesturk:
starting (overdue) review of this one

PR #1495, opened 2017-03-12
Add tool to populate USB drives with Open States quickstart material

mileswwatkins:
I haven't gotten to run it on Linux or Windows (which might affect the paths or escaping), but the rest has been checked and confirmed working, as far as writing to the thumb drives.

### DIFF
--- a/scripts/build-thumbdrive.sh
+++ b/scripts/build-thumbdrive.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+# Purpose:
+# Create a USB drive containing the latest documentation and
+# Docker images needed to run Open States and be able to submit GitHub PRs
+
+# Requirements:
+# - Git
+# - Docker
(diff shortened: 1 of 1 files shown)


mileswwatkins:
Our 4 GB thumb drives are too small to contain all five of the services, so we'll jettison the `openstates/openstates.org` image. No biggie.

mileswwatkins:
@jamesturk, this is now ready to go as far as I can tell, test it out on one of your thumb drives and let me know if I should tweak anything.

mileswwatkins:
Pinging @showerst, to build a USB drive using this (on any environment), and then test whether that USB drive will be useful on a Windows machine!

showerst:
@mileswwatkins got it, i'll give it a go tomorrow.

jamesturk:
👍 I might be slow to get time to test this as next two nights I'm busy (unless snow disrupts), I'll definitely test by weds night though

Issue #1445, opened 2017-03-03
DE: committee scraper needs updating for 2017


jamesturk:
State: DE

Delaware committee scraper needs to be updated

jamesthechamp:
Hey, 
i had a look at this one.
i added links for 2017-2018 session http://legis.delaware.gov/Committees/Senate/149 but it had some error
if no one is working on this .. can i go on..??

jamesturk:
looked at this to see what it would take- seems like you can POST to the underlying JSON, which is what other DE scrapers do: http://legis.delaware.gov/json/Committees/GetCommitteesByTypeId?ga=149

look at legislators.py for examples

hiteshgarg14:
As of now, the approach of the code is getting the full HTML code of page then extract the desired data from the table (using tr and td tags). Verified by `print lxml.html.tostring(page)`.  there is  javascript code in the source code of page that's why there is an issue #1497 

Okay I am on it. rewriting the script which will fetch data using POST to get JSON response.

PR #1496, opened 2017-03-12
de: committee scraper updated for 2017

hiteshgarg14:
#1445 
Updated committees scrapper by mapping of term names to session numbers for 2017.


jamesturk:
doesn't fix DE, but is a step in the right direction

PR #1518, opened 2017-03-16
Convert VT to Pupa

mileswwatkins:
- [x] people
- [x] bills
- [x] committees
- [x] events


jamesturk:
when running this I get a ton of "14:28:42 WARNING pupa: Bill H.R.7 appears to have no activity" but each one I spot checked definitely had actions

mileswwatkins:
Okay, I fixed that bug.

- The pupa2billy compare differences for bills and legislators are simply ordering of lists, as far as I can see.
- Events-wise, I think the weird Billy'd Pupa output (ie, all events missing participants and descriptors) may be due to pupa2billy conversion of event files? But let me know on that.
- The vote comparator broke for me here; if I get a chance I'll go check that out soon.

PR #1500, opened 2017-03-14
[WIP] convert NM to pupa

cliftonmcintosh:
This is a start for converting NM to pupa. It is just the legislators.

I chose to start here because NM is a scraper I have worked on in the past and was familiar with it.

Issue #1442, opened 2017-03-01
pupa conversion


jamesturk:
Our big push for this year is to modernize our infrastructure.  The best way to help with that is to grab a state and port its scrapers from Billy to Pupa. All contributions are extremely welcome, and this is a great way to get to know the OS scrapers!

Here's documentation on the process: http://docs.openstates.org/en/latest/contributing/pupa-conversion.html
- [x] New Hampshire - #1551 - Tim WIP
- [x] Illinois - #1386* - Forest WIP
- [x] Texas - #1526*
- [x] New York - #1664
- [x] Massachusetts - [WIP Branch](https://github.com/openstates/openstates/tree/pupa-ma) - James working on 
- [x] Puerto Rico - #1730 - James reviewing
- [x] Louisiana - #1606 - James reviewing
- [x] West Virginia - #1556* - James reviewing
- [x] California - #1648 - working on eval/fixes
- [x] Iowa - #1604 - James
- [x] Colorado - #1616 - James
- [x] Oklahoma - #1726 - resolving SSL issues & evaluating
- [x] Georgia - #1722 
- [x] Nevada - #1710
- [x] Alaska - #1708
- [x] Rhode Island - #1557*
- [x] Tennessee - #1543
- [x] North Dakota - #1725
- [x] Utah - #1724
- [x] Maryland - [WIP Branch](https://github.com/openstates/openstates/tree/pupa-md) , #1712
- [x] Washington - #1647 
- [x] Maine - #1558*
- [x] Mississippi - #1655
- [x] Alabama
- [x] Arizona
- [x] Arkansas
- [x] Connecticut
- [x] Delaware
- [x] District of Columbia
- [x] Florida
- [x] Hawaii
- [x] Idaho
- [x] Indiana
- [x] Kansas
- [x] Kentucky
- [x] Michigan
- [x] Minnesota
- [x] Missouri
- [x] Montana
- [x] Nebraska
- [x] New Jersey
- [x] New Mexico
- [x] North Carolina
- [x] Ohio
- [x] Oregon
- [x] Pennsylvania
- [x] South Carolina
- [x] South Dakota
- [x] Vermont
- [x] Virginia
- [x] Wisconsin
- [x] Wyoming

_Asterisks indicate incomplete PRs that haven't been updated in a while, and might be closed going forward._

mroswell:
I'll work on Maryland. (Happy to partner with any other Marylanders as well.)

showerst:
[...]

PR #1520, opened 2017-03-16
convert NM to pupa

cliftonmcintosh:
Related to #1442


cliftonmcintosh:
@jamesturk 

I have closed #1500 and opened this instead for converting NM to pupa.

showerst:
Just a head's up that they've recently added two new status codes you'll have to handle to scrape bills. Here are the billy entries, you'll want to change the action coding.

 '7675': ('bill recalled from the House for further consideration by the Senate.', 'other'),

'7808': ('Law Without Signature', 'governor:signed'),

cliftonmcintosh:
Thanks, @showerst 

I will make sure those new status codes are included.

cliftonmcintosh:
This is still in progress. The conversion of bill and vote scraping is a lot more complicated than it was for legislators and committees. Since votes are no longer a part of bills but have references to bills, I intend to remove vote scraping from within the bill scraper and put it in its own scraper. I don't see a way to yield votes from within the bill scraper.

cliftonmcintosh:
@jamesturk 

I think this is ready for review. I've run the scrapers and analyzed the mongo documents that are generated. They seem to be correct.

cliftonmcintosh:
The scrapers that required the biggest changes were the bills and votes scrapers. Specifically, votes were previously being tallied as part of bill scraping. I moved vote collection into its own scraper.

Minor changes made in addition to the conversion:
- switching camelCase variables and method names to snake_case.
- using single-quoted strings consistently in files (because that was the most common form of string already in them)
- switching some old-style string building to the .format function for strings.

jamesturk:
reviewing today

cliftonmcintosh:
Thanks for the update!

Issue #1521, opened 2017-03-16
ME: Legislator scraper is not scraping lower


mshenfield:
Running `docker-compose run openstates me --legislators --lower` I get

```
13:22:55 INFO scrapelib: GET - http://www.maine.gov/legis/house/dist_mem.htm
13:22:56 CRITICAL billy: Error: legislators scraper didn't save any objectslegislator locally, 
```

The [Maine bobsled report](http://bobsled.openstates.org/run-me-2017-03-15.html) confirms that only the 35 senators are being scraped in production, not the 151 house representatives.
<img width="380" alt="screen shot 2017-03-16 at 8 24 05 am" src="https://cloud.githubusercontent.com/assets/4812055/23998434/18df3af8-0a23-11e7-975c-a112f9bdc0fe.png">

ME _is_ populated with legislators from this term (e.g. [freshman representative Mark Lawrence](https://openstates.org/me/legislators/MEL000480/mark-w-lawrence/)) from a previous successful run.

jamesturk:
oof, nice catch

PR #1525, opened 2017-03-17
ME: Scrape representatives.

jmcarp:
Fix ME xpaths such that we scrape representatives instead of not scraping them.


jmcarp:
cc #1521

PR #1507, opened 2017-03-15
[WIP] convert DE to pupa

hiteshgarg14:
Related to #1442

-   [x]  people.py
-   [x]  bills.py
-   [ ]  events.py
-   [ ]  committees.py

PR #1535, opened 2017-03-17
Converted DE to pupa

hiteshgarg14:
Related to #1442
Re: #1507 

-  [x] people.py
-  [x] bills.py
-  [x] events.py
-  [x] committees.py

### DIFF
--- a/billy_metadata/de.py
+++ b/billy_metadata/de.py
@@ -1,0 +1,128 @@
+import lxml.html
+
+metadata = {
+    'name': 'Delaware',
+    'abbreviation': 'de',
+    'legislature_name': 'Delaware General Assembly',
+    'legislature_url': 'http://legis.delaware.gov/',
+    'capitol_timezone': 'America/New_York',
+    'chambers': {
+        'upper': {'name': 'Senate', 'title': 'Senator'},
+        'lower': {'name': 'House', 'title': 'Representative'},
+    },
+    'terms': [
+         {
+            'name': '1999-2000',
+            'start_year': 1999,
+            'end_year': 2000,
+            'sessions': ['140'],
+        },
+         {
+            'name': '2001-2002',
+            'start_year': 2001,
+            'end_year': 2002,
+            'sessions': ['141'],
+        },
+        {
+            'name': '2003-2004',
+            'start_year': 2003,
+            'end_year': 2004,
+            'sessions': ['142'],
+        },
+        {
+            'name': '2005-2006',
+            'start_year': 2005,
+            'end_year': 2006,
+            'sessions': ['143'],
+        },
+        {
+            'name': '2007-2008',
+            'start_year': 2007,
+            'end_year': 2008,
+            'sessions': ['144'],
+        },
+        {
+            'name': '2009-2010',
+            'start_year': 2009,
+            'end_year': 2010,
+            'sessions': ['145'],
+        },
+        {
+            'name': '2011-2012',
+            'start_year': 2011,
+            'end_year': 2012,
+            'sessions': ['146'],
+        },
+        {
+            'name': '2013-2014',
+            'start_year': 2013,
+            'end_year': 2014,
+            'sessions': ['147'],
+        },
+        {
+            'name': '2015-2016',
+            'start_year': 2015,
+            'end_year': 2016,
+            'sessions': ['148'],
+            },
+        {
+            'name': '2017-2018',
+            'start_year': 2017,
+            'end_year': 2018,
+            'sessions': ['149'],
+            },
+    ],
+    'session_details': {
+         '140': {
+            'display_name': '140th General Assembly (1999-2000)',
+            '_scraped_name': '1998 - 2000 (GA 140)',
+        },
+         '141': {
+            'display_name': '141st General Assembly (2001-2002)',
+            '_scraped_name': '2000 - 2002 (GA 141)',
+        },
+         '142': {
+            'display_name': '142nd General Assembly (2003-2004)',
+            '_scraped_name': '2002 - 2004 (GA 142)',
+        },
+         '143': {
+            'display_name': '143rd General Assembly (2005-2006)',
+            '_scraped_name': '2004 - 2006 (GA 143)',
+        },
+        '144': {
+            'display_name': '144th General Assembly (2007-2008)',
+            '_scraped_name': '2006 - 2008 (GA 144)',
+        },
+        '145': {
+            'display_name': '145th General Assembly (2009-2010)',
+            '_scraped_name': '2008 - 2010 (GA 145)',
+        },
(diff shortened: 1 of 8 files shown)


hiteshgarg14:
#1536

Events scraper updated.

openstates/de/bills.py
```diff
@@ -189,19 +218,21 @@ def scrape_vote(self, bill, vote_id):
...
         chamber = self.chamber_codes_rev[sponsor['ChamberId']]
-        bill.add_sponsor(type=sponsor_type,
-                         name=sponsor_name,
-                         district=sponsor_district,
-                         chamber=chamber)
+        bill.add_sponsorship(name=sponsor_name,
+                             classification=sponsor_type,
+                             entity_type='person',
+                             chamber=chamber,
+                             primary=True if sponsor_type == 'primary' else False,
```
jmcarp: Can be simplified to `primary=sponsor_type == 'primary'`.

openstates/de/bills.py
```diff
@@ -110,26 +132,28 @@ def scrape_legislators(self, session):
             'group': '',
             'filter': '',
         }
-        response = self.post(url=search_form_url, data=form, allow_redirects=True)
-        page = json.loads(response.content)
-        #print json.dumps(page, indent=4)
+
+        # response = self.post(url=search_form_url, data=form, allow_redirects=True)
+        page = requests.post(url=search_form_url, data=form, allow_redirects=True).json()
```
jmcarp: Is there an advantage to using `requests.post` instead of `self.post`?

openstates/de/bills.py
```diff
@@ -110,26 +132,28 @@ def scrape_legislators(self, session):
...
         if int(page['Total']) > 0:
             for row in page['Data']:
-                self.legislators[ str(row['PersonId']) ] = row
-                self.legislators_by_short[ str(row['ShortName']) ] = row
+                self.legislators[str(row['PersonId'])] = row
```
jmcarp: Since we're using python3, do we need to convert to a string here?

openstates/de/bills.py
```diff
@@ -139,40 +163,45 @@ def scrape_votes(self, bill, legislation_id):
         }
         response = self.post(url=votes_url, data=form, allow_redirects=True)
         if response.content:
-            page = json.loads(response.content)
+            # page = json.loads(response.content)
+            page = requests.post(url=votes_url, data=form, allow_redirects=True).json()
```
jmcarp: Does this send the same request as on line 164?

openstates/de/bills.py
```diff
@@ -53,52 +76,51 @@ def scrape_bill(self, row, chamber, session):
...
 
         fiscals = html.xpath('//div[contains(@class,"fiscalNote")]/a/@href')
         for fiscal in fiscals:
             self.scrape_fiscal_note(bill, fiscal)
 
         self.scrape_actions(bill, row['LegislationId'])
-        self.scrape_votes(bill, row['LegislationId'])
-
-        self.save_bill(bill)
+        self.scrape_votes(bill, row['LegislationId'], session)
```
jmcarp: Since `scrape_votes` is a generator, should we `yield from` here?

openstates/de/committees.py
```diff
@@ -1,52 +1,54 @@
...
 from openstates.utils import LXMLMixin
 
 
-class DECommitteeScraper(CommitteeScraper, LXMLMixin):
+class DECommitteeScraper(Scraper, LXMLMixin):
     jurisdiction = "de"
 
-    def scrape(self, chamber, term):
+    def scrape(self, chamber=None):
+        if chamber:
```
jmcarp: This would be more concise using the `chambers = [chamber] if chamber else ['upper', 'lower']` from earlier in this PR.

billy_metadata/de.py
```diff
@@ -0,0 +1,128 @@
+import lxml.html
```
jmcarp: Can we drop this import?

billy_metadata/de.py
```diff
@@ -0,0 +1,128 @@
...
+        },
+    },
+    'feature_flags': ['events', 'influenceexplorer'],
+    '_ignored_scraped_sessions': [
+
+    ],
+}
+
+
+def session_list():
```
jmcarp: Should this be moved to `Delaware.get_session_list` in `openstates/de/__init__.py`?

openstates/de/bills.py
hiteshgarg14: Well, I was caching the legislators, we'll need them for sponsors and votes. `row['PersonId'])` is of type `int` and while scraping the page we don't get `int` type data. that's why I choose to cache all legislators data into `str` format. we can do this either way.

openstates/de/bills.py
hiteshgarg14: `scrape_vote` is a generator.

hiteshgarg14:
Thanks, @jmcarp for reviewing.

openstates/de/bills.py
jmcarp: It looks like both `scrape_vote` and `scrape_votes` are generators. When I ran this branch, no votes were scraped.

hiteshgarg14:
There is one problem here:-
what can be the alternative of `executive` chamber?
```
elif 'Governor' in row['ActionDescription']:
                # TODO: <obj>.bill_chamber' is not in the enumeration: [u'upper', u'lower']
                action_chamber = 'executive'
```

jmcarp:
I'm not sure about the chamber issue you pointed out. @jamesturk: can you take a look when you get a chance?

jamesturk:
reviewing this now & fixing things as I go this time.

things to look at in other scrapers as you go:
* flake8 failures (unused variables & line lengths mostly)
* this scraper doesn't save any bills (fixed in my local branch)

```
        legislation_types = {
            'bill': 'HB',
            'concurrent resolution': 'HCR',
            'joint resolution': 'HJR',
            'resolution': 'HR',
            'bill': 'SB',
            'concurrent resolution': 'SCR',
            'joint resolution': 'SJR',
            'resolution': 'SR',
        }
```

openstates/de/bills.py
```diff
@@ -211,46 +243,42 @@ def scrape_actions(self, bill, legislation_id):
...
-            'Senate Joint Resolution': 'SJR',
-            'Senate Resolution': 'SR',
+            'bill': 'HB',
+            'concurrent resolution': 'HCR',
+            'joint resolution': 'HJR',
+            'resolution': 'HR',
+            'bill': 'SB',
+            'concurrent resolution': 'SCR',
+            'joint resolution': 'SJR',
+            'resolution': 'SR',
```
jamesturk: dictionaries can't have duplicate keys, changing this to tuples

PR #1538, opened 2017-03-18
Convert PA to pupa

jmcarp:
* [x] People
* [x] Bills
* [x] Committees
* [x] Events


jmcarp:
Ready for review. Heads up: it looks like event scraping for PA hasn't worked for a while, so I rewrote it as part of this patch. I'm also using `python-dateutil`, because native python datetimes are bad :trollface:

jamesturk:
haven't had a chance to debug why but right now it looks like votes aren't getting saved

jmcarp:
I forgot a `yield from` 😓. We should be collecting votes now.

jamesturk:
👍 checked and looked good.  For future reference: diff looked like more sources on bills, and a few action differences but all innocuous

PR #1543, opened 2017-03-18
Convert TN to pupa

mshenfield:
All conversions followed the generic migration plan outlined [in the pupa conversion docs](http://docs.openstates.org/en/latest/contributing/pupa-conversion.html), except where otherwise noted.

### Bills
* replaces `HTMLParser` with it's Python3 equivalent `html.parser` (used to unescape the email addresses, maybe unnecssarily but leaving it for now)
* replaces calls to `self.lxmlize` with
```python
html = self.get(url)
page = lxml.html.fromstring(html)
```

### Committees
Only generic conversion steps

### Bills and Votes
* ignores duplicate fiscal notes for house and senate - see the Amendments section for [this bill]( http://wapp.capitol.tn.gov/apps/BillInfo/default.aspx?BillNumber=HB0018&GA=110) for for an example.
* updates "other" votes to be "not voting", a more accurate reflection of the votes being cast

### Events
* Use `add_agenda_item` and then `EventAgendaItem` methods to recreate `add_bill` on old event scraper
* Pupa requires fully qualified timezone for when, but setting timezone to correct Central resulted in bad dates in Mongo. Passing to `Event` with UTC tzinfo, but passing correct timezone to `timezone` argument results in correct dates.
* Changed timezone from "US/Eastern" to "US/Central".
* Scraper was skipping events if the event's chamber _matched_ the `chamber` variable. I think this was a mistake and should be the opposite, (skips if doesn't match), so switched.

TODO:
- [x] legislators.py -> people.py
- [x] bills.py
- [x] events.py
- [x] committees.py

### DIFF
--- a/billy_metadata/tn.py
+++ b/billy_metadata/tn.py
@@ -1,0 +1,85 @@
+import datetime
+
+# start date of each session is the first tuesday in January after new years
+
+metadata = dict(
+    name='Tennessee',
+    abbreviation='tn',
+    capitol_timezone='America/Chicago',
+    legislature_name='Tennessee General Assembly',
+    legislature_url='http://www.legislature.state.tn.us/',
+    chambers={
+        'upper': {'name': 'Senate', 'title': 'Senator'},
+        'lower': {'name': 'House', 'title': 'Representative'},
+    },
+    terms=[
+        {'name': '106', 'sessions': ['106'],
+            'start_year': 2009, 'end_year': 2010},
+        {'name': '107', 'sessions': ['107'],
+            'start_year': 2011, 'end_year': 2012},
+        {'name': '108', 'sessions': ['108'],
+            'start_year': 2013, 'end_year': 2014},
+        {'name': '109', 'sessions': ['109', '109s1', '109s2'],
+            'start_year': 2015, 'end_year': 2016},
+        {'name': '110', 'sessions': ['110'],
+            'start_year': 2017, 'end_year': 2018},
+    ],
+    session_details={
+        '110': {
+            'type': 'primary',
+            'display_name': '110th Regular Session (2017-2018)',
+            '_scraped_name': '110th General Assembly'
+        },
+        '109s2': {
+            'type': 'special',
+            'start_date': datetime.date(2016, 9, 12),
+            'end_date': datetime.date(2016, 9, 14),
+            'display_name': '109th Second Extraordinary Session (September 2016)',
+            '_scraped_name': '2nd Extraordinary Session (September 2016)'},
+        '109s1': {
+            'type': 'special',
+            'start_date': datetime.date(2016, 2, 1),
+            'end_date': datetime.date(2016, 2, 29),
+            'display_name': '109th First Extraordinary Session (February 2016)',
+            '_scraped_name': '1st Extraordinary Session (February 2015)'},
+        '109': {
+            'type': 'primary',
+            'display_name': '109th Regular Session (2015-2016)',
+            '_scraped_name': '109th General Assembly'},
+        '108': {
+            'type': 'primary',
+            'display_name': '108th Regular Session (2013-2014)',
+            '_scraped_name': '108th General Assembly'},
+        '107': {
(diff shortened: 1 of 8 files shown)


billy_metadata/tn.py
```diff
@@ -0,0 +1,85 @@
...
+        '107th General Assembly',
+        '105th General Assembly', '104th General Assembly',
+        '103rd General Assembly', '102nd General Assembly',
+        '101st General Assembly', '100th General Assembly',
+        '99th General Assembly'
+    ]
+)
+
+
+def session_list():
```
jmcarp: Could you make this a method of the `Tennessee` class and name it `get_session_list`?

Also, it's probably about time to decide where `url_xpath` should live: https://github.com/openstates/openstates/blob/master/openstates/fl/__init__.py#L7-L8. I wound up moving this to `openstates.utils` in a few separate PRs--WDYT @jamesturk?

jmcarp:
Could you add Tennessee to `flake8.sh` and check for lint?

mileswwatkins:
Finally got to review the output data! Good work on clearing out those duplication bugs, and some others! Sorry that this took so long for me to hit, @mshenfield!

- people: cleared!
- committees: cleared!
- events
  - Missing committee participants (ie, the existing Billy scraper identifies which committees are participating, based on the meeting name, but the Pupa code does not)
    - Do note, the existing Billy scraper code identifies the participating committee's chamber _incorrectly_, so watch for that if you add committee participants back in
- bills
  - Similar to events, the Pupa scraper is missing some committee participants for actions that pertain to a committee (whereas the Billy scraper did identify those committees as related participants)
  - Some of the `Signed by Senate Speaker` and `Signed by H. Speaker` events are getting incorrect datetime stamps in Pupa, making them appear to take place _days after_ a bill has been signed into law. For example, check out SB 0116.
  - Some bills (_maybe_ bills that appear in both chambers, ie have alternate IDs?) are duplicated. For example, `_data/tn/bill_17764db4-23b9-11e7-9d51-0242ac130003.json` and `_data/tn/bill_354733e8-23b0-11e7-9d51-0242ac130003.json` have the exact same file content (those are for HB 0333). However, if this is because of the alternate IDs, I'm surprised that the two files have the same bill ID _and_ the same `other_identifiers.identifier`, so this may be more complex of an issue?
- votes
  - The bill duplication above (I believe) results in two copies of many (all?) votes. Again, check out HB 0333 for one example: this grep will return four votes (two identical pairs) `grep -nr _data/tn -e http://wapp.capitol.tn.gov/apps/BillInfo/default.aspx?BillNumber=HB0333&GA=110`.

Unrelated issue, while you're doing this PR and already adding a whitespace strip to action text, can you also add a whitespace strip to legislator District Office addresses? There're some ` `s and `\xa0`s that tail some of those.

Let me know if you'd like me to clarify any of the above, or provide more examples!

mshenfield:
Thanks for the excellent feedback! I'm preparing for to move next weekend, will have some time to take a look at it next week.

mshenfield:
Added fixes for:
* Double scraping bills/votes
* Whitespace on legislator address

It looks like pupa2billy is [skipping participants right now](http://github.com/openstates/openstates/blob/3214655a8ea7f5b13596ae6918397d69e92c7bef/pupa2billy/events.py#L32). This matches your comment on [on the VT conversion PR](https://github.com/openstates/openstates/pull/1518#issuecomment-296481749).

I'm not able to reproduce the issue with incorrect timestamps.  My output  [SB 0116](http://wapp.capitol.tn.gov/apps/BillInfo/default.aspx?BillNumber=SB0116&GA=110) locally for the "Signed by" actions was:

```python
{
    "date": ISODate("2017-04-17T00:00:00Z"),
    "action": "Signed by H. Speaker",
    "type": ["other"],
    "related_entities": [],
    "actor": "lower"
}, {
    "date": ISODate("2017-04-17T00:00:00Z"),
    "action": "Signed by Senate Speaker",
    "type": ["other"],
    "related_entities": [],
    "actor": "upper"
}
```
which matches the dates of these actions online.

PR #1552, opened 2017-03-19
UT: Convert to Pupa

jschiarizzi:
Updated scrapers to pupa 

- [x] Legislators (now people.py)
- [ ] Bills
- [x] Committees
- [ ] events


jschiarizzi:
Committees.py is done. Moving right along...

PR #1565, opened 2017-03-19
indiana pupa

rshorey:
Notes:
* requires Indiana API key (if you have one, you've got to include the word "token" as part of the key)
* I tried to include datetimes which are apparently available for bill actions and votes. But pupa choked, so I did not include the times. The time lines are still in the code but commented out should that change.
* removed the actions classifier. We were not using it and it was not entirely clear to me how to move it to pupa.
* Indiana takes several hours to scrape completely, even with `--fast`, because of aggressive waittimes and generally terrible site reliability
* I did it for the [nandoz](https://www.nandosperiperi.com/).


jamesturk:
testing this one

jamesturk:
this was merged w/ some rebasing, thanks rachel!

PR #1567, opened 2017-03-20
[WIP] Switch docker environment to alpine linux.

jmcarp:
This drops the image size from 876mb to 436mb. Did I do good @hmoco?


mileswwatkins:
Since we're not doing local building of images anymore, my thought is we look at this from a maintenance angle as much as a filesize angle: will Alpine be as useful and stable going into the future as Ubuntu or Debian?

jamesturk:
@jmcarp and/or @hmoco - any thoughts on Miles' question?

jmcarp:
I haven't read anything that convinces me that alpine will be much harder to maintain than ubuntu or debian (lmk if you've seen horror stories), but we'd have to occasionally build things from source (like mdbtools). Then again, I don't know how much benefit we'd get from a smaller image--the base layer won't change and might already be cached for contributors who already use docker. Do you all think image size has been a problem? If not, maybe we can close this and potentially revisit another time.

mileswwatkins:
I can think of a few cases where a smaller image size would be a benefit:

- Fitting more onto the 4GB thumb drives that Open States already owns (right now we can't fit all images, and if we wanted to pre-load a database dump as well that's out of the question given current sizes)
- Quicker first-time builds for anyone doing local development of OS Docker images
- Potentially faster builds for Docker Hub and any CI tools we'd use in the future
- Potentially faster kick-offs of production images on AWS
- Quicker Docker Hub downloads and less disk space taken up for Open States developers

Costs:

- Marginally more difficult to set up packages (ie, more custom building, potentially less rich documentation)
- Potentially marginally less supported going forward (what happens if a leaner Linux image comes along and eats Alpine's lunch?)
- We have a member of the Debian core team on OS developers, so we could ask him questions about that distro directly

I say that the former outweighs the latter by a long shot. Just had wanted to hear what y'all thought before I 👍'd.

mileswwatkins:
@jmcarp, I'm good to merge this in, but want to make sure first: have you tested it with some of the edge-case-y scrapers, like MA, CA, etc.? Just want to double-check that the extra packages work where they're needed!

jamesturk:
I'd like to try this, the build times have been a burden lately while trying to sort out some issues w/ the production images and I am hoping this will help w/ that.

jmcarp:
@jamesturk: I updated this to move the alpine work to a separate image, which is working for me so far.

jamesturk:
thanks, going to start building this on dockerhub and will switch a couple scrapers to it

PR #1574, opened 2017-03-20
AR Pupa conversion new PR.

In-vincible:
#1442, right now I have one issue that I don't see vote reports, though I am yielding them.
- [x] people
- [x] Committee
- [x] Events
- [x] bills
- [x] vote


jmcarp:
Could you add `openstates/ar` to `flake8.sh` and fix the handful of flake8 warnings?

billy_metadata/ar.py
```diff
@@ -0,0 +1,193 @@
...
+
+def session_list():
+    links = url_xpath('http://www.arkleg.state.ar.us/assembly/2013/2013R/Pages'
+        '/Previous%20Legislatures.aspx', '//a')
+    sessions = [a.text_content() for a in links if 'Session' in a.attrib.get(
+        'title', '')]
+    return sessions
+
+
+def extract_text(doc, data):
```
jmcarp: Is this function used, and if not, can we drop it?

billy_metadata/ar.py
```diff
@@ -0,0 +1,193 @@
...
+        'Regular Session, 1987 ',
+        'First Extraordinary Session, 1987',
+        'Second Extraordinary Session, 1987',
+        'Third Extraordinary Session, 1987',
+        'Fourth Extraordinary Session, 1987',
+    ],
+}
+
+
+def session_list():
```
jmcarp: Could you move this function to a method of the `Arkansas` class and rename it to `get_session_list`?

openstates/ar/__init__.py
```diff
@@ -1,197 +1,189 @@
...
-    },
-    'terms': [
+
+class Arkansas(Jurisdiction):
+    division_id = "ocd-division/country:us/state:ar"
+    classification = "government"
+    name = "Arkansas"
+    url = "http://www.arkleg.state.ar.us"
+    scrapers = {
+        'people': ARLegislatorScraper,
```
jmcarp: Should the event scraper be added here?

openstates/ar/events.py
```diff
@@ -35,51 +36,57 @@
 }
 
 
-class AREventScraper(EventScraper):
+class AREventScraper(Scraper):
     jurisdiction = 'ar'
```
jmcarp: Pupa doesn't use the `jurisdiction` class variable, so we can drop this now.

openstates/ar/bills.py
```diff
@@ -57,29 +61,31 @@ def scrape_bill(self, chamber, session):
...
             # TODO: Keep on eye on this post 2017 to see if they apply R going forward.
             session_code = '2017R' if session == '2017' else session
 
             version_url = ("ftp://www.arkleg.state.ar.us/Bills/"
                            "%s/Public/%s.pdf" % (
                                session_code, bill_id.replace(' ', '')))
-            bill.add_version(bill_id, version_url, mimetype='application/pdf')
+            bill.add_version_link(bill_id, version_url, media_type='application/pdf')
 
             self.scrape_bill_page(bill)
```
jmcarp: It looks like `scrape_bill_page` is now a generator, so we need to `yield from` this call to save votes.

billy_metadata/ar.py
In-vincible: yeah this is not used droping it.

jmcarp:
There are some print statements and commented-out code that I'm guessing might have been used for debugging. Could you look through the PR and either delete commented-out blocks if they're not needed, or uncomment and fix if we should keep them?

openstates/ar/__init__.py
In-vincible: blunder spotted :100: really embarrassed :p

In-vincible:
one issue than I believe everything is solved, how to add notes to events, like in billy we used to simply set event['notes']?

In-vincible:
#1599 Fixed.

PR #1583, opened 2017-03-21
CA: Add code to compute email from name

samtregar:
CA assemblymembers and senators have working emails but they're not shown on public pages anymore.

### DIFF
--- a/openstates/ca/legislators.py
+++ b/openstates/ca/legislators.py
@@ -1,5 +1,6 @@
 import re
 import collections
+import unicodedata
 from operator import methodcaller
 
 import lxml.html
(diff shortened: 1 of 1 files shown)


jamesturk:
thanks for this, I think we'll need to make a judgment call here.

this is a bit unusual as we typically only surface data that we've collected, any idea why they removed these?

are they no longer a preferred method of contact?

also, I'm a bit concerned what happens here if two people wind up w/ the same email address

samtregar:
I imagine they got tired of the spam you get by having an email address on a searchable webpage.  The email addresses still work, which you can verify by emailing something else (senator.foo@ca.sen.gov for example) - you'll see that it bounces.

I don't know what they'd do if they had two reps with the same last name.  I imagine they don't know either since they don't have that problem now!

samtregar:
I could add a check for collisions so we'd get a scrape error if two people had the same last name.  Let me know if you want that.

samtregar:
Can I get a ruling on this?  If it's not going in then I need to work on a plan B to curate a list of email addresses not in OpenStates.  If it's going in then I can work on patches for other states (ex. TX).  Thanks!

jamesturk:
sorry for the delay, I'll give this a 👍 for now, I'd rather not make this precedent- I think we should be sure to carefully consider when/where we do things like this, but given the potential reach here I think the pros outweigh the cons for now.  if we get complaints from CA legislators or find out that the email addresses are essential black holes now we should be sure to clear these out

Issue #1563, opened 2017-03-19
NH: some votes still missing in current term


estaub:
I can't tell how many votes are missing - I just have this one case which I happened to want to work on:

From NH: [NH HB 478 roll calls](http://www.gencourt.state.nh.us/bill_status/Roll_calls/Billstatus_billrollcalls.aspx?lsr=240&sy=2017&sortoption=billnumber&txtsessionyear=2017&txtbillnumber=hb478)

From OpenStates:
```
{
    "id": "NHB00006048", "type": ["bill"], "state": "nh",
    "title": "prohibiting discrimination based on gender identity.", "votes": [], "actions": [
    {
        "date": "2017-01-12 09:35:27", "type": ["bill:introduced", "committee:referred"],
        "actor": "lower",
        "action": "Introduced 01/05/2017 and referred to Health, Human Services and Elderly Affairs HJ 3 P. 14",
        "related_entities": []
    }, {
        "date": "2017-02-08 03:58:42", "type": ["other"], "actor": "lower",
        "action": "==CANCELLED== Public Hearing: 02/14/2017 01:15 PM LOB 305-307",
        "related_entities": []
    }, {
        "date": "2017-02-15 02:38:21", "type": ["other"], "actor": "lower",
        "action": "==RESCHEDULED== Public Hearing: 02/21/2017 01:15 PM LOB 305-307",
        "related_entities": []
    }, {
        "date": "2017-02-22 09:54:07", "type": ["other"], "actor": "lower",
        "action": "==CANCELLED== Executive Session: 02/28/2017 09:00 AM LOB 205",
        "related_entities": []
    }, {
        "date": "2017-02-24 08:11:54", "type": ["other"], "actor": "lower",
        "action": "Executive Session: 02/23/2017 LOB 205", "related_entities": []
    }, {
        "date": "2017-02-28 02:51:52", "type": ["committee:passed:favorable"], "actor": "lower",
        "action": "Committee Report: Ought to Pass for 03/08/2017 (Vote 15-2; RC)",
        "related_entities": []
    }, {
        "date": "2017-03-09 10:48:55", "type": ["other"], "actor": "lower",
        "action": "Lay on Table (Rep. Packard): MA RC 187-179 03/09/2017", "related_entities": []
    }, {
        "date": "2017-03-09 05:32:15", "type": ["other"], "actor": "lower",
        "action": "Remove from Table (Rep. Butler): MF RC 168-180 03/09/2017",
        "related_entities": []
    }
[...]

PR #1594, opened 2017-03-24
NH: Fix vote names scraper

mshenfield:
Fixes two issues with the NH voter name scrape.

* Similar to an earlier fix for the vote counts, this adds `splitlines()` to the file so that it is processed line by line instead of character by character.
* In addition, this handles an extra two columns being returned when splitting each line.

Confirmed that votes were not being saved with the bug in the voter name scrape, and they are with the fix. This fixes the underlying cause of #1563.


mshenfield:
Returns both votes, with a number of individual voter names recorded:

```js
db.votes.find({"bill_id": "NHB00000388"}).count()
2
```

PR #1603, opened 2017-03-30
Convert WY to pupa

jmcarp:
* [x] People
* [x] Committees
* [x] Bills
* [ ] ~~Events~~

### DIFF
--- a/billy_metadata/wy.py
+++ b/billy_metadata/wy.py
@@ -1,0 +1,93 @@
+metadata = {
+    'name': 'Wyoming',
+    'abbreviation': 'wy',
+    'legislature_name': 'Wyoming State Legislature',
+    'legislature_url': 'http://legisweb.state.wy.us/',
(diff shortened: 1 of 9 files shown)


jmcarp:
@jamesturk @mileswwatkins: two questions before this is ready to review.
* Is there a pupa equivalent to `Bill.description` from billy?
* It looks like Wyoming has changed from HTML event pages to PDF, and the PDFs aren't consistently formatted. We can get event dates and abbreviated titles ("Mgt Council") from the calendar page at http://legisweb.state.wy.us/LSOWEB/Events.aspx, but anything else is going to require some potentially brittle PDF munging. Since the Wyoming event scraper isn't working now anyway, I'm wondering if we should disable it instead of porting.

jamesturk:
👍  to disabling the event scraper


 bill.add_abstract is the best analog for description IIRC

jmcarp:
Thanks, updated.

PR #1607, opened 2017-03-31
CO: Fix chamber-less joint bill actions

showerst:
CO Added some new actions with no chamber marked, which broke the bill scraper. All the current ones without chamber are joint, so adding that to the logic.


In-vincible:
it should be merged quickly after checking, so that I can take forward the bill conversion of co without conflicts.

PR #1638, opened 2017-04-12
Deal with missing NJ emails

samtregar:
It seems they've been removed from the Access DBs on the FTP site, but they still work and appear on many reps personal sites (although not in a form we could easily scrape).  The formula is simple enough - just AsmLAST@njleg.org and SenLAST@njleg.org.

### DIFF
--- a/openstates/nj/legislators.py
+++ b/openstates/nj/legislators.py
@@ -7,6 +7,7 @@ from billy.scrape.legislators import LegislatorScraper, Legislator
 from .utils import clean_committee_name, MDBMixin
 
 import scrapelib
+import unicodedata
 
 class NJLegislatorScraper(LegislatorScraper, MDBMixin):
     jurisdiction = 'nj'
@@ -56,6 +57,14 @@ class NJLegislatorScraper(LegislatorScraper, MDBMixin):
             email = None
             if rec["Email"]:
                 email = rec["Email"]
+
+            # Email has been removed from the Access DB, but it's
+            # still AsmLAST@njleg.org and SenLAST@njleg.org - many
+            # reps have these emails on their personal pages even if
+            # they're gone from the DB file
+            if not email:
+                email = self._construct_email(chamber, last_name)
+
             try:
                 photo_url = photos[rec['Roster Key']]
             except KeyError:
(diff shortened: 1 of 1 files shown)


jamesturk:
I'm concerned about basing this on last name alone.  I know we made an exception before but it seems to me like if they've removed these we may want to let them go & possibly explore a manual solution in the future

samtregar:
If they don't want to be emailed at these addresses, why are they still on their personal sites?  I suppose I could write a scraper that walks them, but of course they're all different so it would be more like a web crawler than the usual...

samtregar:
Example of what I'm talking about: http://district10.senatenj.com/contact.php

samtregar:
Actually now that I'm looking at it, these districtXX.senatenj.com sites do have a nice standard format and have both reps and sens emails.  If you won't take this patch I think I could do a scraper for them.  Let me know.

jamesturk:
We should probably try to verify why they were removed, in other cases this has been b/c the emails have become a black hole, so serving them isn't useful anyway.

Either way, guessing the email addresses will inevitably lead us to giving people bad data.

jamesturk:
I'll email a contact at the NJ Senate to try to find out more

samtregar:
Do you have a contact?  I don't have any way to find out why they were removed.  Will you take a patch that scrapes their personal sites?

samtregar:
Sounds good, thanks!

samtregar:
Any progress on this?

jamesturk:
just sent a second email and cc'd a Senator that had been in touch before

jamesturk:
No response yet from any of the NJ emails... do we know anyone in NJ that is more likely to get a reply?  Hard to tell if the email is a black hole or they just don't want to talk to us

samtregar:
I don't have any contacts personally.  I can try emailing a few at random to see if I can get a response though.

samtregar:
Confirmed, at least one of these emails are live!  I just got a (hilariously late) reply from sensweeney@njleg.org.

jamesturk:
ha alright, I guess we'll give them a try.  if you have a chance to fix this PR to work w/ the new NJ scraper that'd be great, otherwise I'll try to find some time in the next week or so

PR #1639, opened 2017-04-12
Add code to find emails on the PA contact-your-rep page

samtregar:
Looks like this only gets us 37 email addresses, all senate, but it's better than none.

### DIFF
--- a/openstates/pa/people.py
+++ b/openstates/pa/people.py
@@ -64,9 +64,6 @@ class PALegislatorScraper(Scraper):
             raise AssertionError("Legislator photo parsing needs to be rewritten")
 
     def scrape_email_address(self, page, leg_id):
-        if leg_id == "1011":
-            import pdb; pdb.set_trace()
-
         # find the right block on the contact page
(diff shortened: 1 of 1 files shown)


openstates/pa/people.py
```diff
@@ -56,11 +63,21 @@ def scrape_photo_url(self, url, page, person):
         else:
             raise AssertionError("Legislator photo parsing needs to be rewritten")
 
-    def scrape_email_address(self, url, page):
-        if re.search(r'var \S+\s+= "(\S+)";', page):
-            vals = re.findall(r'var \S+\s+= "(\S+)";', page)
-            return '%s@%s%s' % tuple(vals)
+    def scrape_email_address(self, page, leg_id):
+        if leg_id == "1011":
```
jamesturk: let's make sure to remove this

jamesturk:
looks pretty good, mind fixing the flake8 errors and removing the ipdb, then 👍

openstates/pa/people.py
samtregar: Ha, whoops!  Will do.

samtregar:
Dropped the debug code (gah) and fixed the flake8 warnings.

Issue #1478, opened 2017-03-08
Add Git hook to wipe PYC files on-checkout


mileswwatkins:
To avoid confusing issues for all devs, we should wipe PYC files whenever a new branch is checked out, so that there isn't persistent behavior despite having different PY files.

Via https://git-scm.com/docs/githooks#_post_checkout

jamesturk:
https://stackoverflow.com/questions/12222186/are-git-hooks-pushed-to-the-remote-when-i-git-push  I think this isn't possible?

mileswwatkins:
Ah, damn!

We _could_ add it as a step in our Dockerfiles, since they `ADD` the whole of the project; since we're switching to a Dockerized system, and that's what our docs suggest, this seems like a solid solution.

A step right after the `ADD` that's, like, `RUN find ${PROJECT_PATH} -name '*.pyc' -delete`?

jamesturk:
SGTM 👍

jamesturk:
actually right after merging this I realized, this only gets when the docker image is created... I'm not 100% sure this will fix the problem

jamesturk:
reopened on purpose?

mileswwatkins:
Yup, I agree that the problem isn't fixed.

jamesturk:
Maybe we set PYTHONDONTWRITEBYTECODE if this is still causing problems?

jamesturk:
let's run w/ this for now

PR #1640, opened 2017-04-12
proposed solution to #1478

jamesturk:


jmcarp:
I was also thinking about handling .pyc files with a .dockerignore: https://github.com/jmcarp/openstates/blob/9442c938f819b50612b8fb98defcacc13a05ef95/.dockerignore. I don't have a preference though--this also LGTM.

PR #1643, opened 2017-04-14
MA rollcall scrape

nicklryan:


openstates/ma/bills.py
```diff
@@ -194,13 +199,162 @@ def scrape_action_page(self, bill, bill_url, page_number):
...
+                    # handles vote count lines
+                    cut_name = raw_name.split('-')
+                    clean_name = ''
+                    if cut_name[len(cut_name) - 1].strip(' .').isdigit():
+                        del cut_name[-1]
+                        clean_name = ''.join(cut_name)
+                    else:
+                        clean_name = raw_name.strip()
+
+                    # handle name exceptions - probably somewhere better to put these
```
jamesturk: you can remove these, we typically do the matching after the fact

openstates/ma/legislators.py
```diff
@@ -71,8 +71,12 @@ def scrape_member(self, chamber, term, member_url):
...
         full_name = root.xpath('//h1/span')[0].tail.strip()
 
-        email = root.xpath('//a[contains(@href, "mailto")]/@href')[0]
-        email = email.replace('mailto:', '')
+        try:
+            email = root.xpath('//a[contains(@href, "mailto")]/@href')[0]
+            email = email.replace('mailto:', '')
+        except:
+            email = ''
+            print("seat may be vacant")
```
jamesturk: please use a self.info log call for this

openstates/ma/bills.py
```diff
@@ -194,13 +199,162 @@ def scrape_action_page(self, bill, bill_url, page_number):
...
+                        clean_name = "Ives, Kathleen O'Connor"
+                    elif clean_name == "Humason, Donald F., Jr.":
+                        clean_name = "Donald F. Humason, Jr."
+
+                    # update vote object with names
+                    if mode == 'y':
+                        vote.yes(clean_name)
+                    elif mode == 'n':
+                        vote.no(clean_name)
+                    #TODO: Other votes
```
jamesturk: let's make sure we don't exclude those

openstates/ma/bills.py
```diff
@@ -194,13 +199,162 @@ def scrape_action_page(self, bill, bill_url, page_number):
...
+                vote_tally.append('x')
+            #Present
+            elif line == 'P':
+                vote_tally.append('p')
+
+            #True for all records 2009 - 2017 - brittle code, this will change in the future
+            elif line == 'Mr. Speaker':
+                voters.append('DeLeo')
+            # Handle legislators with same last name
+            elif line == 'Moran F.':
```
jamesturk: we can resolve these after the fact, no need to do this here

nicklryan: I'll remove where I hardcoded names. Does this include the Mr. Speaker items?

nicklryan:
Removed edge case stuff, fixed other votes, and got code working for 2015 and 2016 which don't have votes. Should be ready for next review

estaub:
If there was any hope that this works as a 2017 vote scraper, it doesn't look like it.  See e.g. https://malegislature.gov/Bills/190/H58/RollCall versus https://openstates.org/ma/bills/190th/H58/ .

nicklryan:
Are you sure the scrape has run on the website yet? That bill was the one I used for testing and it is working when I run the scrape on my own. I don't think james has accepted the pull request yet so I don't think the data is up yet.

estaub:
@nicklryan Sorry you saw that, I deleted it when, upon reading more carefully, I realized it's not merged yet AFAICT.

nicklryan:
No prob, just happy someone will be using it

jamesturk:
yep, sorry for the delay on the review.  from what I could tell this works pretty well & in the interest of getting the data in front of more eyes going ahead w/ a merge 👍 thanks for tackling this one!

PR #1644, opened 2017-04-15
[WIP] sd (South Dakota) to pupa

In-vincible:
#1442 
- [x] metadata
- [x] people 
- [x] committee
- [x] bills
- [x] votes


In-vincible:
I have commented out the "organization" parameter of VoteEvent as of now.

openstates/sd/people.py
```diff
@@ -0,0 +1,125 @@
...
+import lxml.html
+
+
+class SDLegislatorScraper(Scraper):
+    latest_only = True
+
+    def scrape(self, chambers=None):
+        url = 'http://www.sdlegislature.gov/Legislators/default.aspx' \
+              '?CurrentSession=True'
+        if chambers is None:
```
jmcarp: The `scrape` method seems to accept `chamber` instead of `chambers` most of the time. I'm guessing we should keep this consistent.

openstates/sd/people.py
```diff
@@ -0,0 +1,125 @@
+import re
+
+from pupa.scrape import Person, Scraper, Organization
+
+import lxml.html
+
+
+class SDLegislatorScraper(Scraper):
+    latest_only = True
```
jmcarp: I don't think `latest_only` exists in pupa.

openstates/sd/people.py
```diff
@@ -0,0 +1,125 @@
...
+                name = link.text.strip()
+
+                yield from self.scrape_legislator(name, chamber,
+                                                  '{}&Cleaned=True'.format(
+                                                    link.attrib['href']))
+
+    def scrape_legislator(self, name, chamber, url):
+        page = self.get(url).text
+        page = lxml.html.fromstring(page)
+        page.make_links_absolute(url)
```
jmcarp: What do you think about using `LXMLMixin` so we can call `self.lxmlize` and simplify some of this repeated code?

openstates/sd/bills.py
```diff
@@ -159,51 +176,55 @@ def scrape_vote(self, bill, date, url):
         else:
             raise ScrapeError("Bad chamber: %s" % location)
 
-        committee = ' '.join(location.split(' ')[1:]).strip()
-        if not committee or committee.startswith('of Representatives'):
-            committee = None
+        # committee = ' '.join(location.split(' ')[1:]).strip()
```
jmcarp: Can we still pass the committee to the vote via the `organization` field? cc @jamesturk

billy_metadata/sd.py
```diff
@@ -0,0 +1,98 @@
+import datetime
+
+
+settings = dict(
```
jmcarp: Do you know if we still need these settings? If so, they should be in `openstates/sd/__init__.py`.

In-vincible: I will come back to this one, after looking into this .

openstates/sd/people.py
In-vincible: totally!

openstates/sd/people.py
In-vincible: okay I will change this.

billy_metadata/sd.py
In-vincible: I am not sure about this one actually, I remember checking your mi PR for pupa conversion, for this purpose, you seemed to have avoided settings dictionary, though I am not sure if I can avoid it for SD.

In-vincible:
I have made all the changes above, apart from these 2:

# Passing the committee to the vote via the organization field throws error with classification field.
# Format to add settings dictionary to "sd/__init__.py", or may be neglect it like in mi.

PR #1655, opened 2017-04-19
[WIP] MS "Mississippi" pupa

In-vincible:
#1442 
- [x] metadata
- [x] people 
- [x] committee
- [x] bills
- [x] votes


In-vincible:
Not ready for review right now, few small glitches to consider.
1. replacing with char, gives type error, this to be fixed(though removing the replace works fine which is right now state).
2. To decide weather to avoid sponsor urls or save them, if save them where to save them(simple add_source to bill is also an option).

In-vincible:
lint to be fixed.

In-vincible:
@jamesturk ready for review.

hiteshgarg14:
Compared the scraped data.

Votes & Bills differ on dates.
I think that it's correct:-  [ `_tz = pytz.timezone('CST6CDT')`]

Rest is good to go :+1:

PR #1550, opened 2017-03-19
[WIP] DC to pupa

jalbertbowden:

PR #1678, opened 2017-04-27
Fix regression in metadata conversion.

jmcarp:
Looks like we accidentally deleted the lower chamber metadata in #1550. This also allows `_ignored_scraped_sessions` to be missing.


jamesturk:
good catch, thanks!

PR #1723, opened 2017-05-12
MT, NC: session_list() movement

hiteshgarg14:
Just noticed it. So moved `session_list()` to __init__.py file for MT and NC.

Note:- Removed `extract_text()` from Montana's __init__.py file. If we need this one then should I add `text_after_line_numbers() & pdfdata_to_text()` to `openstates/utils`? These functions are used in other state __init__.py files also.

### DIFF
--- a/openstates/mt/__init__.py
+++ b/openstates/mt/__init__.py
@@ -84,6 +84,6 @@ class Montana(Jurisdiction):
         yield lower
 
     def get_session_list(self):
-        from billy.scrape.utils import url_xpath
+        from openstates.utils.lxmlize import url_xpath
         return url_xpath('http://leg.mt.gov/css/bills/Default.asp',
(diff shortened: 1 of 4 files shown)


openstates/mt/__init__.py
```diff
@@ -82,3 +82,8 @@ def get_organizations(self):
         yield legislature
         yield upper
         yield lower
+
+    def get_session_list(self):
+        from openstates.utils.lxmlize import url_xpath
```
jmcarp: No big deal, but I'm not sure why the import is happening inside the method. Any reason not to move to the module level?

hiteshgarg14: Not a particular reason. just copied the `function` from `mt.py` file and updated the import path.

PR #1727, opened 2017-05-13
[WIP] AZ: Scrape votes.

jmcarp:
This is close, but could use attention from somebody who's more familiar with details of votes than me to verify output. For example, I'm not sure what we should do with "votes" like this one: https://apps.azleg.gov/api/BillStatusFloorAction?billStatusActionId=177494&billStatusId=68871&includeVotes=true. Should we include votes that don't concern passing a bill or amendment like this one?

@jamesturk @mileswwatkins

### DIFF
--- a/openstates/az/bills.py
+++ b/openstates/az/bills.py
@@ -1,12 +1,13 @@
 import json
+import urllib
 import datetime
 
-from pupa.scrape import Scraper, Bill
+from lxml import html
+from pupa.scrape import Scraper, Bill, VoteEvent
+
 from . import utils
 from . import action_utils
 from . import session_metadata
-
-from lxml import html
 
 
 BASE_URL = 'http://www.azleg.gov/'
@@ -45,6 +46,7 @@ class AZBillScraper(Scraper):
         bill = self.scrape_versions(bill, internal_id)
         bill = self.scrape_sponsors(bill, internal_id)
         bill = self.scrape_subjects(bill, internal_id)
+        yield from self.scrape_votes(bill, page)
 
         bill_url = 'https://apps.azleg.gov/BillStatus/BillOverview/{}?SessionId={}'.format(
                     internal_id, session_id)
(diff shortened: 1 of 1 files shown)


jamesturk:
Right now we won't include those.  In theory the OCD Vote model can handle them just fine, billy certainly can't.

Anything particular you want me to look over in the output?  I can give it a run while I'm waiting on one of the never-ending MA runs

jmcarp:
For visibility, discussing a question I have about this at https://openstates.slack.com/archives/C2DG0RBS6/p1496526137477033.

jamesturk:
I think it would be OK to ignore these when scraping votes.

jmcarp:
@jamesturk this looks a little more complicated than I was hoping. I checked the javascript that actually displays whether an action counts as passed or failed on the AZ site, and the unminified version looks like this:

```javascript
    r.getSimplePassFail = function() {
        return (r.floorCommDetails()[0].Ayes || 0) != 0 && (r.floorCommDetails()[0].Nays || 0) != 0 ? (r.floorCommDetails()[0].Ayes || 0) > (r.floorCommDetails()[0].Nays || 0) ? "Passed" : "Failed" : ""
    }
    ;
    r.getPassFail = function(n, t, i, u, f, e, o, s) {
        if (r.billInfo() == null)
            return "";
        var h = "Failed";
        return i == "E" || u == 1 || f || 0 ? r.bodyTab() == "H" ? (n || 0) > 40 && (h = "Passed") : (n || 0) > 20 && (h = "Passed") : (e || 0) != 0 && o == "" ? r.bodyTab() == "H" ? (n || 0) > 45 && (h = "Passed") : (n || 0) > 23 && (h = "Passed") : s ? h = "Passed" : (n || 0) > (t || 0) && (r.bodyTab() == "H" ? (n || 0) > 30 && (h = "Passed") : (n || 0) > 15 && (h = "Passed")),
        h
    }
    ;
```

I'm wondering if we should try asking AZ how to decide if an action passed or failed--or at least for a more human-readable version of this code.

jamesturk:
oof yeah, I don't have any good contacts in AZ, but if you have a chance to reach out that'd be great

openstates/az/bills.py
```diff
@@ -213,6 +215,51 @@ def actor_from_action(self, bill, action, self_chamber):
...
+                vote.set_count('no', action['Nays'] or 0)
+                vote.set_count('other', (
+                    (action['Present'] or 0) + (action['Absent'] or 0) +
+                    (action['Excused'] or 0) + (action['NotVoting'] or 0)))
+
+                for v in action['Votes']:
+                    vote_type = {
+                        'Y': 'yes',
+                        'N': 'no',
+                    }.get(v['Vote'], 'other')
```
fgregg: We generally prefer more brittle scraping patterns. This code could continue to "work" even if yes votes are changed from "Y" to "y" or "yes"

openstates/az/bills.py
```diff
@@ -213,6 +215,51 @@ def actor_from_action(self, bill, action, self_chamber):
...
+                        else 'fail'
+                    ),
+                    start_date=action_date.strftime('%Y-%m-%d'),
+                    bill=bill,
+                )
+                vote.add_source(resp.url)
+                vote.set_count('yes', action['Ayes'] or 0)
+                vote.set_count('no', action['Nays'] or 0)
+                vote.set_count('other', (
+                    (action['Present'] or 0) + (action['Absent'] or 0) +
```
fgregg: Absent is not an "other" vote.

openstates/az/bills.py
```diff
@@ -213,6 +215,51 @@ def actor_from_action(self, bill, action, self_chamber):
...
+                vote = VoteEvent(
+                    chamber={
+                        'S': 'upper',
+                        'H': 'lower',
+                    }[header['LegislativeBody']],
+                    motion_text=action['Action'],
+                    classification='passage',
+                    result=(
+                        'pass'
+                        if action['UnanimouslyAdopted'] or action['Ayes'] > action['Nays']
```
fgregg: Is this always true. Does every bill need a simple majority to pass?

openstates/az/bills.py
```diff
@@ -213,6 +215,51 @@ def actor_from_action(self, bill, action, self_chamber):
...
 
+    def scrape_votes(self, bill, page):
+        base_url = 'https://apps.azleg.gov/api/BillStatusFloorAction'
+        for header in page['FloorHeaders']:
+            params = {
+                'billStatusId': page['BillId'],
+                'billStatusActionId': header['BillStatusActionId'],
+                'includeVotes': 'true',
+            }
+            resp = self.get('{}?{}'.format(base_url, urllib.parse.urlencode(params)))
```
fgregg: Under the hood, `self.get` is from the [requests library](http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls). So you can just do this as

```python
resp = self.get(base_url, params=params)
```

openstates/az/bills.py
```diff
@@ -213,6 +215,51 @@ def actor_from_action(self, bill, action, self_chamber):
...
+            actions = json.loads(resp.content.decode('utf-8'))
```
fgregg: this can just be

```python
actions = resp.json
```

openstates/az/bills.py
```diff
@@ -1,13 +1,14 @@
 import json
+import urllib
```
fgregg: don't need this (see comment's below)

PR #1743, opened 2017-05-21
MI: Handle committee names with misspelled "committeee".

jmcarp:
Handle committee titles like "Commerce Committeee", which otherwise
resolves to "Commercee".

Should fix issues like:

screen shot 2017-05-20 at 8 42 36 pm

If we have a contact at Michigan, we could also ask them to fix the typo in the header of http://www.senate.michigan.gov/committee/commerce.html.


jamesturk:
ha 👍

PR #1749, opened 2017-05-24
[WIP] Convert TX to pupa

scichelli:
Ready for review:

- Person
- Committees
- Bills
- Votes
- Events

### DIFF
--- a/openstates/tx/bills.py
+++ b/openstates/tx/bills.py
@@ -211,54 +211,54 @@ class TXBillScraper(Scraper, LXMLMixin):
             introduced = False
 
             if desc == 'Amended':
-                atype = 'amendment:passed'
+                atype = 'amendment-passage'
             elif desc == 'Amendment(s) offered':
-                atype = 'amendment:introduced'
+                atype = 'amendment-introduction'
             elif desc == 'Amendment amended':
-                atype = 'amendment:amended'
+                atype = 'amendment-amendment'
             elif desc == 'Amendment withdrawn':
-                atype = 'amendment:withdrawn'
+                atype = 'amendment-withdrawal'
             elif desc == 'Passed' or desc == 'Adopted':
-                atype = 'bill:passed'
+                atype = 'passage'
             elif re.match(r'^Received (by|from) the', desc):
                 if 'Secretary of the Senate' not in desc:
-                    atype = 'bill:introduced'
+                    atype = 'introduction'
                 else:
-                    atype = 'bill:filed'
+                    atype = 'filing'
             elif desc.startswith('Sent to the Governor'):
                 # But what if it gets lost in the mail?
-                atype = 'governor:received'
+                atype = 'executive-receipt'
             elif desc.startswith('Signed by the Governor'):
-                atype = 'governor:signed'
+                atype = 'executive-signature'
             elif desc == 'Vetoed by the Governor':
-                atype = 'governor:vetoed'
+                atype = 'executive-veto'
             elif desc == 'Read first time':
-                atype = ['bill:introduced', 'bill:reading:1']
+                atype = ['introduction', 'reading-1']
                 introduced = True
             elif desc == 'Read & adopted':
-                atype = ['bill:passed']
+                atype = ['passage']
                 if not introduced:
                     introduced = True
-                    atype.append('bill:introduced')
+                    atype.append('introduction')
             elif desc == "Passed as amended":
-                atype = 'bill:passed'
+                atype = 'passage'
             elif (desc.startswith('Referred to') or
                     desc.startswith("Recommended to be sent to ")):
-                atype = 'committee:referred'
+                atype = 'referral-committee'
             elif desc == "Reported favorably w/o amendment(s)":
-                atype = 'committee:passed'
+                atype = 'committee-passage'
             elif desc == "Filed":
-                atype = 'bill:filed'
+                atype = 'filing'
             elif desc == 'Read 3rd time':
-                atype = 'bill:reading:3'
+                atype = 'reading-3'
             elif desc == 'Read 2nd time':
-                atype = 'bill:reading:2'
+                atype = 'reading-2'
             elif desc.startswith('Reported favorably'):
-                atype = 'committee:passed:favorable'
+                atype = 'committee-passage-favorable'
             else:
-                atype = 'other'
+                atype = None
 
-            if 'committee:referred' in atype:
+            if 'referral-committee' in atype:
                 repls = [
                     'Referred to',
                     "Recommended to be sent to "
--- a/openstates/tx/bills.py
+++ b/openstates/tx/bills.py
@@ -2,16 +2,15 @@ import datetime
 import ftplib
 import re
 import time
-import urlparse
+from urllib import parse as urlparse
 import xml.etree.cElementTree as etree
 
-from billy.scrape import ScrapeError
-from billy.scrape.bills import BillScraper, Bill
+from pupa.scrape import Scraper, Bill
+from pupa.scrape.base import ScrapeError
 from openstates.utils import LXMLMixin
 
 
-class TXBillScraper(BillScraper, LXMLMixin):
-    jurisdiction = 'tx'
+class TXBillScraper(Scraper, LXMLMixin):
     _FTP_ROOT = 'ftp.legis.state.tx.us'
     CHAMBERS = {'H': 'lower', 'S': 'upper'}
     NAME_SLUGS = {
@@ -39,7 +38,7 @@ class TXBillScraper(BillScraper, LXMLMixin):
             raise
         ftp.login()
         ftp.cwd('/' + dir_)
-        self.log('Searching an FTP folder for files ({})'.format(dir_))
+        self.logger.info('Searching an FTP folder for files ({})'.format(dir_))
 
         lines = []
         ftp.retrlines('LIST', lines.append)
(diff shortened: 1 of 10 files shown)


scichelli:
Step 3 of [Converting Metadata](http://docs.openstates.org/en/latest/contributing/pupa-conversion.html#converting-metadata) instructs to move `session_list` to `get_session_list` in `tx/__init__.py`,  but since Greg hadn't done this, yet the scraper still ran successfully, I [searched the repo for get_session_list](https://github.com/openstates/openstates/search?p=1&q=get_session_list&type=&utf8=%E2%9C%93). I see many states' definition of the function, but not usages, I don't think. Should I be adding a `get_session_list` for TX, or perhaps submitting a PR to update that instruction to just delete from `billy_metadata/nc.py`?

jamesturk:
get_session_list is used when the scraper is invoked to make sure that new sessions haven't been added on the state site, please do include it for TX :)

jamesturk:
also- thanks for picking this one up!  as you can see we're down to the last few, do you have an idea of when you think you might be ready w/ this one?

scichelli:
@jamesturk After I do step 2 (Committees) on my own, I'll have a better idea for how long this could take. I know these outstanding conversions are blocking the project's switchover to pupa. What kind of timeline are you hoping for--if I finished before the end of June, would that be adequate, or do I need to corral a team to work in parallel?

jamesturk:
No worries- so we're down to the final 8 with probably 2-3 more going live
today.  I think we might be done in the next week or so, which might mean
we'd need to split this up.  If you can let us know when you're done w/
your first piece we can take another look at where we are then and
depending on if other states are still outstanding we can evaluate then :)

openstates/tx/people.py
```diff
@@ -172,17 +192,18 @@ def _scrape_representative(self, url, term, parties):
 
         # Vacant house "members" are named after their district numbers:
         if re.match(r'^\d+$', scraped_name):
-            return None
+            yield None
```
scichelli: Greg had a concern here, "I'm a little concerned this may blow something up, but there aren't any empty district seats at the moment."

openstates/tx/bills.py
```diff
@@ -259,30 +268,34 @@ def scrape_bill(self, session, history_url):
...
                 extra['committees'] = ctty
 
-            bill.add_action(actor, action.findtext('description'),
-                            act_date, type=atype, **extra)
+            bill.add_action(
+                action.findtext('description'),
+                act_date,
+                chamber=actor,
+                classification=atype
+            )
```
scichelli: This used to pass a dictionary of extra info, `**extra`. If there isn't a way to just send along that extra info, then I need to find analogous [functions for passing that extra info](http://opencivicdata.readthedocs.io/en/latest/scrape/bills.html#overview) and remove any properties that no longer have a home.

openstates/tx/bills.py
```diff
@@ -202,54 +211,54 @@ def scrape_bill(self, session, history_url):
             introduced = False
```
scichelli: This variable is set but not used.

openstates/tx/bills.py
```diff
@@ -259,30 +268,34 @@ def scrape_bill(self, session, history_url):
...
+                act_date,
+                chamber=actor,
+                classification=atype
+            )
 
         for author in root.findtext('authors').split(' | '):
             if author != "":
-                bill.add_sponsor('primary', author, official_type='author')
+                bill.add_sponsorship(author, classification='primary',
+                                     entity_type='person', primary=True)
```
scichelli: For all four of these calls to `add_sponsorship`, how should I determine whether the entity_type should be `person` or `organization` (for a committee)?

openstates/tx/people.py
```diff
@@ -75,38 +75,53 @@ def _get_chamber_parties(self, chamber, term):
```
scichelli: Local variable `district` might be referenced before assignment

openstates/tx/bills.py
jamesturk: gotcha, feel free to drop this

openstates/tx/bills.py
jamesturk: have you seen any committees in the data?  do they identify themselves in any way (i.e. do they all say committee?) 

if not it might be OK to assume they're all person

openstates/tx/bills.py
jamesturk: do you know what info is in the extra dict?

openstates/tx/people.py
jamesturk: I can take a look at this during the final review, thanks for flagging

jamesturk:
looks like this is coming along well- nice work so far &  let us know if we can help in any way, aiming to try to knock out IL, NH and TX review by early next week if possible

jamesturk:
just did a quick review of people & committees - data looks pretty good, almost all matches (there's one issue right now in the pupa2billy portion where multiple offices w/ the same name get overwritten so I might look at a solution for those)

will start reviewing bills & votes soon

jamesturk:
haven't had a chance to try it yet, but I think that dropping '.content' from the self.get() call before line 125 might fix it

jamesturk:
I made a couple of minor changes if you want to pick from them: https://github.com/openstates/openstates/tree/pupa-tx

still running bills scraper

scichelli:
@jamesturk Thanks! :D Votes is not ready yet, so don't waste your time on that just yet. I'll pick up your changes (Saturday); thank you.

openstates/tx/bills.py
scichelli: I'm mistaken. A bill can have an array of classifications, and this variable makes sure it only tacks the word "introduction" on when it should. But with it being set to `False` at the top of the for loop, can it ever be `True` when adding a second action? I would think it gets reset for each `action`. What am I missing?

openstates/tx/bills.py
scichelli: So far what I've seen are all people.

openstates/tx/people.py
scichelli: PyCharm called it out, but in actual practice, it's not a problem. This is in a loop walking the cells of a table, and for each row it will always encounter the 3rd cell before the 7th.

openstates/tx/bills.py
scichelli: The three properties were `action_number`, `comment`, and `committee`. I couldn't find a likely home for `comment` in the [Bills or Action classes](https://github.com/opencivicdata/pupa/blob/master/pupa/scrape/bill.py), so I removed that code; we can restore it if there's a good spot to put it. `committee` can now be added as a related_entity to the action, so I did that. Since `action_number` was the only remaining item in the dictionary, I turned it into just a variable and kept its usage as-is. However, it did used to get passed along to the old call to `add_action`, so maybe the action has a place for it that I should be setting?

openstates/tx/tests/test_vote_parsing.py
```diff
@@ -1,27 +1,31 @@
 import os
 
-from nose.tools import *  # noqa
+import unittest
```
scichelli: This switch from nose to unittest was to assuage the flake8 linter.

Also, I'm not sure how to run these. I think the [instructions on the readme](https://github.com/openstates/openstates/#testing) work for billy scrapers but not pupa, or I've broken something here.

openstates/tx/votes.py
```diff
@@ -315,58 +315,41 @@ def scrape(self, chamber, session):
         if len(session) == 2:
             session = "%sR" % session
 
-        #As of 1/30/15, the 84th session does not have journals on the ftp
```
scichelli: Checked 6/3/17 and there still are not 84th session journals, nor 85th (current), so it doesn't seem to be a supported resource anymore. I removed the commented-out code.

openstates/tx/votes.py
```diff
@@ -258,55 +242,71 @@ def clean_bill_id(bill_id):
...
+            result='pass' if mv.passed else 'fail',
+            classification='passage' if mv.passed else 'other',
+            legislative_session=session[0:2],
+            bill=mv.bill_id,
+            bill_chamber=mv.chamber
+        )
+
+        v.set_count('yes', mv.yeas or 0)
+        v.set_count('no', mv.nays or 0)
+        v.set_count('not voting', mv.present or 0)
```
scichelli: `absent` isn't included in the header that gives counts for Yeas and Nays that these other values are pulled from. Could try counting it from the list of names in `mv.votes['absent']`, if you think that's a good idea.

openstates/tx/votes.py
```diff
@@ -258,55 +242,71 @@ def clean_bill_id(bill_id):
...
     for el in root.xpath('//div{}'.format(''.join(vote_selectors))):
         mv = MaybeVote(el)
         if not mv.is_valid:
             continue
 
-        v = Vote(None, None, 'passage' if mv.passed else 'other', mv.passed,
-                 mv.yeas or 0, mv.nays or 0, mv.present or 0)
-        v['bill_id'] = mv.bill_id
-        v['bill_chamber'] = mv.chamber
-        v['is_amendment'] = mv.is_amendment
```
scichelli: I didn't find a new place to record `is_amendment` or `method` ('record' or 'viva voce') anymore.

scichelli:
@jamesturk Thanks for all your help so far. Votes and Events are now ready for review. :)

jamesturk:
things look great, thanks for your hard work on this!

the remaining differences were things we never officially supported from what I could tell 👍

Issue #1709, opened 2017-05-09
NJ: duplicate votes


pqnelson:
Gail Phoebus voted "no" and "other" for a vote [Reported out of Assembly committee with amendments](https://openstates.org/nj/votes/NJV00023908/), which seems incorrect.

I'm uncertain what's causing this error, otherwise I would endeavour to help.

**Addendum.** After some more digging, it appears to be a committee vote. The raw data comes from the [CA2016-2017](ftp://www.njleg.state.nj.us/votes/CA2016-2017.zip) votes, specifically lines 3837 _et seq._ which read:

```
"A",2361,"Disab.-proh. snowplow block access",2/8/2016 13:00:00,"AHU","M",0,"r/Aca","Vainieri Huttle, Valerie","D","Yes"
"A",2361,"Disab.-proh. snowplow block access",2/8/2016 13:00:00,"AHU","M",0,"r/Aca","Tucker, Cleopatra G.","D","Yes"
"A",2361,"Disab.-proh. snowplow block access",2/8/2016 13:00:00,"AHU","M",0,"r/Aca","Jones, Patricia Egan","D","Yes"
"A",2361,"Disab.-proh. snowplow block access",2/8/2016 13:00:00,"AHU","M",0,"r/Aca","Phoebus, Gail","R","Abstain"
"A",2361,"Disab.-proh. snowplow block access",2/8/2016 13:00:00,"AHU","M",0,"r/Aca","Howarth, Joe","R","Abstain"
"A",2361,"Disab.-proh. snowplow block access",2/8/2016 13:00:00,"AHU","M",0,"r/Aca","McKnight, Angela V.","D","Yes"

"A",2361,"Disab.-proh. snowplow block access",4/4/2016 13:00:00,"AAP","M",0,"r/Aca","Conaway, Herb","D",""
"A",2361,"Disab.-proh. snowplow block access",4/4/2016 13:00:00,"AAP","M",0,"r/Aca","Burzichelli, John J.","D","Yes"
"A",2361,"Disab.-proh. snowplow block access",4/4/2016 13:00:00,"AAP","M",0,"r/Aca","Lampitt, Pamela R.","D","Yes"
"A",2361,"Disab.-proh. snowplow block access",4/4/2016 13:00:00,"AAP","M",0,"r/Aca","Schaer, Gary S.","D","Yes"
[...]

PR #1766, opened 2017-05-29
NJ: Separate votes by committee.

jmcarp:
Distinguish between committee and floor votes so that vote counts aren't
combined across different vote events.

Per @fgregg, we might also want to capture the committee as a `bill_action`, but I'm not sure how that would work--cc @jamesturk

### DIFF
--- a/openstates/nj/bills.py
+++ b/openstates/nj/bills.py
@@ -382,6 +382,7 @@ class NJBillScraper(Scraper, MDBMixin):
                         date = rec["Session_Date"]
                         action = rec["Action"]
                         leg_vote = rec["Legislator_Vote"]
+                        vote_parts = (bill_id, chamber, action)
                     else:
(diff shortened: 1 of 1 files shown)


jamesturk:
when you say capture committee as a bill action what do you mean?  as in add_related_entity?

jmcarp:
The issue described in #1709 is that NJ incorrectly combines committee votes with floor votes with the same bill ID. This patch separates committee and floor votes, so that we get two events instead of one, but doesn't indicate which event is the committee vote and which is the floor vote. I'm trying to understand how to encode those details.

jamesturk:
I'd like to go ahead and merge this so that we can fix #1709 - the larger question of how to properly associate votes w/ committees going forward is a good one.  pupa supports this as a VoteEvent object can be associated w/ any organization, but we'd need to have special logic in pupa2billy to flatten this down to the expected chamber values (upper/lower) for billy until we're done w/ OS API v1.

PR #1795, opened 2017-06-19
WV: UnboundLocalError

hiteshgarg14:
#1794

This PDF file is the main cause of the problem. There may be others in future like this. so not sure how to handle these type of warning/errors.
```
02:04:15 INFO scrapelib: GET - http://www.legis.state.wv.us/legisdocs/2017/1X/votes/house/00662.pdf
Syntax Warning: May not be a PDF file (continuing anyway)
Syntax Error: Couldn't find trailer dictionary
Syntax Error: Couldn't find trailer dictionary
Syntax Error: Couldn't read xref table
Traceback (most recent call last):
.........................................
```

### DIFF
--- a/openstates/wv/bills.py
+++ b/openstates/wv/bills.py
@@ -288,13 +288,16 @@ class WVBillScraper(Scraper):
                     if not name:
                         continue
                     votes[vote_type].append(name)
-
-        vote = Vote(chamber='lower',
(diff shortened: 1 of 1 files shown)


jamesturk:
I think we should alter the code to set date to None (to avoid the UnboundLocalError) and then check if date wasn't set in the parsing and print an error instead of saving the vote if the PDF has the error that is causing this to fail

PR #1801, opened 2017-06-23
CO: header index_error solved

hiteshgarg14:
#1800


hiteshgarg14:
After taking inspiration from https://medium.com/@jpmckinney/fighting-null-in-web-scrapers-126c4b3d487a
handling IndexError with pupa Warning.

Issue #1782, opened 2017-06-12
TX not parsing legislators' names for all votes


scichelli:
State: Texas

I'm working on finding the root cause; just logging a ticket here to make it easy to communicate status.

When scraping votes, the scraper does not always separate the list of names into individual records, and instead creates one record with all the legislators' names in a list.

For example, download the TX CSV from [Open States Downloads](https://openstates.org/downloads/) and find the rows for vote `TXV00004597`. There are only three, one for each vote type, instead of one for each legislator, and for example the name for the 'yes' vote is a list, 'Bettencourt, Birdwell, Buckingham, Campbell, Creighton...' On the web, you can see that https://openstates.org/tx/votes/TXV00004597/ is not able to display the list of votes with links to the legislators' pages; contrast with https://openstates.org/tx/votes/TXV00000001/ which correctly shows a table of votes.

estaub:
Is this fixed, or become even worse?  The referenced vote (https://openstates.org/tx/votes/TXV00004597/) is no longer present, which may be good or bad.

scichelli:
Hi @estaub. I don't think _worse_, although I too noticed that some IDs that were previously present no longer were, which told me I didn't yet understand the overall system well enough. Also, I've been struggling to come up to speed with Docker and Docker Compose, to be able to isolate the Texas scraper's problems locally. That's why this issue is languishing.

The Texas scraper could certainly be improved, but perhaps it would be best to close this particular github issue, and open separate ones as we identify them.

Flaws I'm aware of:
* In many House journals, R. Anderson's name has the wrong punctuation, making it split into separate names. You'll see `Anderson, C.; Anderson; R.;`, where that semi-colon between `Anderson` and `R` ought to be a comma. Our scraper is tripped up by this typo.
* When the Senate has duplicate last names, they'll write them like `Taylor of Galveston`, and the scraper doesn't handle this.
* It doesn't match a vote event back to the legislator when the legislator's name contains unicode characters, e.g., `Rodr\u00edguez`.
* I've been told that Eddie Lucio, Jr., and Eddie Lucio, III, are not getting treated as separate people.
* A subject matter expert on our team told me the Bill Subjects are not containing the text she expects. I haven't dug into this.

I'm okay with closing this issue for the sake of housekeeping until more specific bugfix issues can be created.

estaub:
@scichelli How about creating the point issues (or a bucket list issue for them) _and_ closing this?  (I'm not a project member, don't look to me for direction!)
[...]

PR #1816, opened 2017-06-29
[WIP] TX Votes: parsing legislators' names more successfully

scichelli:
First attempt at addressing issue #1782.

The existing scraper would split the list of names on semi-colons. It looks like, for the latest session at least, the Senate used commas. (The House still used semi-colons.) You can see the effect of this on, for example, [SB 16](https://openstates.org/tx/bills/85/SB16/), where the House vote page lists a separate entry for each person, but the Senate vote page doesn't split out the names.

Good:
image

Bad:
image

The attempt in this PR is to split by semi-colons, and if that result doesn't seem to have created a plausible number of separate names, try splitting by commas. This feels admittedly hacky. 

If it's not too much trouble, I would love advice on better ways to attempt the fix as well as a solid strategy for verifying that I didn't make it worse. I'm just spot-checking, which isn't robust enough.

### DIFF
--- a/openstates/tx/votes.py
+++ b/openstates/tx/votes.py
@@ -54,35 +54,43 @@ def clean_journal(root):
 
 def names(el):
     text = (el.text or '') + (el.tail or '')
+    split_name_list = text.split(';')
+    if len(split_name_list) < 7:
+        # probably failed to properly split on semi-colons; try commas:
+        split_name_list = text.split(',')
 
     names = []
-    for name in text.split(';'):
+    for name in split_name_list:
         name = name.strip().replace('\r\n', '').replace('  ', ' ')
 
         if not name:
             continue
 
-        if name == 'Gonzalez Toureilles':
-            name = 'Toureilles'
-        elif name == 'Mallory Caraway':
-            name = 'Caraway'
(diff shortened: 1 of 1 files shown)


openstates/tx/votes.py
```diff
@@ -54,35 +54,43 @@ def clean_journal(root):
...
-        if name == 'Gonzalez Toureilles':
-            name = 'Toureilles'
-        elif name == 'Mallory Caraway':
-            name = 'Caraway'
-        elif name == 'Martinez Fischer':
-            name = 'Fischer'
-        elif name == 'Rios Ybarra':
-            name = 'Ybarra'
-
+        name = clean_name_special_cases(name)
```
scichelli: This special handling of names was [introduced in 2010](https://github.com/openstates/openstates/commit/05595720f7a637458f20849f01ffde7472e70780). I wonder if it's still relevant... In any case, I moved it to its own function to keep the main function more streamlined.

jamesturk:
this looks pretty good, I'm going to go ahead and merge it.  

in general a good way to test these things is to look in the database after you do a run and see how many things have (for example) only one voter

PR #1822, opened 2017-06-29
Dc: argument of type 'NoneType' error solved and SC: photo_url error solved

hiteshgarg14:
#1819 
#1820

### DIFF
--- a/openstates/sc/people.py
+++ b/openstates/sc/people.py
@@ -50,12 +50,18 @@ class SCPersonScraper(Scraper):
 
             # District # - County - Map
             district = district.split()[1]
-
-            photo_url = leg_doc.xpath('//img[contains(@src,"/members/")]/@src')[0]
-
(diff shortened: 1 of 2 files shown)


openstates/sc/people.py
```diff
@@ -50,12 +50,18 @@ def scrape_chamber(self, chamber):
...
-            photo_url = leg_doc.xpath('//img[contains(@src,"/members/")]/@src')[0]
-
-            person = Person(name=full_name, district=district,
-                            party=party, primary_org=chamber,
-                            image=photo_url)
+            try:
+                photo_url = leg_doc.xpath('//img[contains(@src,"/members/")]/@src')[0]
+            except IndexError:
+                self.warning("No Photo URL for {}".format(full_name))
+                photo_url = None
```
jamesturk: if you set photo_url to '' here, you won't have to special case the constructor below

PR #1825, opened 2017-07-01
OK: Replace proxy server usage with curl shell command

rsimoes:
This is a workaround for getting pages off okhouse.gov's TLS 1.0 server, which python's core ssl module seems to have trouble talking to.

My system `curl` is compiled with GnuTLS, which may be the key to it working. I'm not sure whether a `curl` compiled with OpenSSL will also be successful.


jamesturk:
awesome, thanks for this- will give it a test run then merge very soon

PR #1864, opened 2017-07-21
[bugfix] GA, ME votes assigned to legislators

### LINKED ISSUE #1812: GA votes need name matching
The GA state senate is missing votes (e.g., [GAV00006507](https://openstates.org/ga/votes/GAV00006507/) doesn't have a single vote) **and** the state Assembly is missing votes (e.g., [GAV00006524](https://openstates.org/ga/votes/GAV00006524/) doesn't have a single vote). This appears quite widespread for the current term in GA, that votes seem to be missing on nearly every motion.

**Addendum.** The problem appears to be the clerks record votes as "LAST NAME, district number" which messes up the vote parsing. But if we simply remove the ", district number" part of the string, the votes are identified with the legislator correctly, which seems to fix the problem.

unfold-inc:
This fixes #1812

### DIFF
--- a/openstates/ga/bills.py
+++ b/openstates/ga/bills.py
@@ -1,3 +1,4 @@
+import re
 from collections import defaultdict
 
 from pupa.scrape import Scraper, Bill, VoteEvent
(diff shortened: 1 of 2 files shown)


jamesturk:
so the issue in both of these cases is that this loses identifying information on the voter that we use in ambiguous cases (hence being commented out in one case)

that said, this may be expedient if there are no current duplicates- I haven't had a chance to verify check if that is/isn't the case yet

Issue #1160, opened 2016-11-19
VT: inconsistent district naming, Grand Isle


estaub:
Senator Mazza has legislator.district field of `Grand Isle`, but the name and partial ID of the upper district is actually `Chittenden-Grand Isle`, making it impossible to relate the legislator to the district (the district's `legislators` array is empty).

I'll be continuing to post these data-integrity issues, but fear they may make it hard to see larger, more urgent concerns.  If there's some way I should tag them or or otherwise handle them, please sing out.

PR #1891, opened 2017-08-17
VT: Remove special casing of Grand Isle district

rsimoes:
Grand Isle is an actual state senate district (https://ballotpedia.org/Vermont_State_Senate_Grand_Isle_District). I assume there was a case where the state house district of Grand Isle-Chittenden was incorrectly listed somewhere, but I'm not seeing that now.

### DIFF
--- a/openstates/vt/people.py
+++ b/openstates/vt/people.py
@@ -40,8 +40,6 @@ class VTPersonScraper(Scraper, LXMLMixin):
                 state_email = None
 
             district = info['District'].replace(" District", "")
-            if district == 'Grand Isle':
-                district = 'Chittenden-Grand Isle'
 
             leg = Person(
                 primary_org=self.CHAMBERS[info['Title']],


estaub:
Reference #1160 for original motivation, which I believe is still valid internal to OpenStates.  In other words, I think that if this should be changed, the name on the district itself should change to `Grand Isle` (sans Chittenden) as well in __init__.py.  But I'm not a member or anything.

The Census is still using `Grand Isle-Chittenden` - see e.g. https://www.census.gov/geo/maps-data/maps/sldl/st50_vt.html .  I also downloaded the shapefile at https://www.census.gov/geo/maps-data/data/cbf/cbf_sld.html, and it uses the `Grand Isle-Chittenden` form as well.  

For future reference - for Open States, what's the ultimate source of truth for this kind of stuff?

estaub:
Sorry, more.  

My Census references above appear to be irrelevant, in that OpenStates uses `Chittenden-Grand Isle`, not `Grand Isle-Chittendon`.  

The Census uses `Grand Isle-Chittendon` for both the House and Senate districts.

I don't know where the OpenStates form comes from.  The Vermont legislature itself appears to uses `Grand Isle-Chittendon` for the House, but `Grande Isle District` for the Senate - see [here](http://legislature.vermont.gov/PeoplePage_Controller/PeopleSearchForm?SearchBy=Town&Name=0&Town=41&SecurityID=02d5ccdde168959f223bcccced612d7f3201cf51&Biennium=2018&action_doPeopleSearch=Search).  This is somewhat consistent with how they treat other districts - see e.g. [Rutland](http://legislature.vermont.gov/PeoplePage_Controller/PeopleSearchForm?SearchBy=Town&Name=0&Town=81&SecurityID=02d5ccdde168959f223bcccced612d7f3201cf51&Biennium=2018&action_doPeopleSearch=Search)

Issue #1853, opened 2017-07-18
NE: only "yes" votes recorded on roll calls; abstentions aren't even counted.


estaub:
It may make sense to split this into two issues.

See e.g. https://openstates.org/ne/votes/NEV00001528/, from http://nebraskalegislature.gov/bills/view_votes.php?KeyID=2085.

Individual legislator votes are not recorded for "No" or "Other" votes.

The yes/no/other vote counts do not include abstentions (`Present - Not Voting`) under 'Other'.

julianedwards:
Hello,

I am new to this community but would like to start contributing. I found the source of error for the exclusion of 'Present - Not Voting' counts in the 'Other' category: 
Whenever an 'other_match' is found, the 'other' count in 'vote' is overwritten. Since the 'Present - Not Voting' line always comes before 'Excused - Not Voting' line in the legislative journal, if both appear for a particular vote the 'Excused - Not Voting' count will overwrite the 'Present - Not Voting' count. Therefore, some vote counts do not include 'Present - Not Voting' abstentions under 'Other'.

<img width="456" alt="screen shot 2017-10-04 at 1 13 17 pm" src="https://user-images.githubusercontent.com/16562432/31189432-36184d3a-a906-11e7-8573-58ec1147f445.png">

I can get started on this fix right now.

julianedwards:
As for the individual legislator votes not being counted for 'No' and 'Other' votes, I am not seeing that issue. Are you referring to the list of names under 'Full Roll Call'? Or something else?
https://openstates.org/ne/votes/NEV00001594/

estaub:
@julianedwards These now appear correct, probably due to other fixes.  The original example URL I pointed at, https://openstates.org/ne/votes/NEV00001528/, has disappeared, which is a different issue (see above).

PR #1929, opened 2017-10-04
Fixed abstentions counting in NE roll call count.

julianedwards:
This is in reference to: #1853


jamesturk:
thanks! this looks great

PR #1956, opened 2017-11-07
Update Dockerfile

doubleswirve:
Hello,

Was running into an issue w/ https://hub.docker.com/r/openstates/openstates/ image, where running `docker run openstates/openstates nc --fastmode` would result in the following error:

```
nc: 1: nc: [/opt/openstates/openstates/pupa-scrape.sh]: not found
```

By wrapping the `ENTRYPOINT` executable in double quotes, I was able to build an image locally and run the `docker run ...` command from above successfully.

Additionally, I noticed the `mdbtools` package was listed twice, so I removed to second occurrence. Happy to undo if that was intentional.

The only other change was a whitespace removal triggered by saving the file in my Atom text editor.

Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,9 +31,9 @@ RUN apt-get update && apt-get install -y \
     mysql-server \
     libmysqlclient-dev \
     mongodb-org-tools \
-    mdbtools \ 
+    mdbtools \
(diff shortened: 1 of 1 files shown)


jamesturk:
thanks 👍

PR #1973, opened 2017-11-21
TN: Unescape ampersands in motions

estaub:
Fixes https://github.com/openstates/openstates/issues/1691#issuecomment-346080631.
Tennessee HTML-escapes motions.
This fix is safe even if they fix the issue.


jamesturk:
thanks 👍

Issue #1907, opened 2017-09-07
MO: All 2017 individual roll call votes are missing for both houses


estaub:
(nothing more to say)

estaub:
The House has no individual roll call votes at all for 2017, i.e. the `yes_votes`, `no_votes`, and `other_votes` arrays are always empty. The first comment is updated to reflect this.

estaub:
After reading the scraper code at https://github.com/openstates/openstates/blob/master/openstates/mo/votes.py 
(with Python-ignorant eyes)
I believe that the House vote scraper is simply not implemented, as it's very hard (at best).

```py
    def _scrape_lower_chamber(self, session):
        # house_url = 'http://www.house.mo.gov/journallist.aspx'
        yield from ()
#  Ugh, so, the PDFs are in nasty shape. Scraping them is a mess, with
#  crazy spacing to break up the names. Most votes aren't on bills, but rather
#  the agenda of the day.
```

estaub:
See above, "with crazy spacing to break up the names".  If this refers to the insertion of letter-spacing within words, I've gotten good results on this in the past with a filter that removes spacing smaller than a certain limit.  It was a long time ago, I doubt I have the code.

jamesturk:
after looking at this, we only have 69 recorded votes for MO and all are classified as legislature

estaub:
I just renamed this to reflect the current status.  I don't know whether there are unresolved votes; my stuff strips them out immediately as useless.

showerst:
for future reference here's a sample vote roll-call PDF: https://house.mo.gov/BillTracking/PDFViewer/web/viewer.html?file=/billtracking/bills181/jrnpdf/jrn025.pdf#page=12

I haven't researched every edge case but I'd call that "borderline unscrapeable".

estaub:
I've turned the Senate vote scraper back on (merge pending).  See openstates/openstates-scrapers#2315.

PR #1978, opened 2017-11-30
fix for new MO House website

agelimson:


estaub:
Is it at all likely that this would fix #1907?  
A quick scan of the diff makes me think probably not.

agelimson:
No, that's more complicated I think. The issue it fixes is:

Yesterday the MO House website was updated, so all the MO House scrapers are broken and with this change they'll be working again.

PR #1980, opened 2017-12-01
readme: command options, acquiring/using state API keys.

estaub:
No doubt someone more experienced can doc this better; please review carefully.


estaub:
@jamesturk Are there issues with these changes?  The current guidance is pretty broken.

PR #2080, opened 2018-01-12
UT: Fix people scraper and simplify to rely solely on JSON source

mileswwatkins:


jamesturk:
are the old comments in there about the representatives data not being complete still true?

jamesturk:
code looks good though, 👍  to merg

mileswwatkins:
The house rep data has been beefed up since those comments; the JSON's now complete for both H and S members. Thanks for the eyes!

PR #2083, opened 2018-01-12
UT: Switch to much cleaner JSON for committee scraping

mileswwatkins:
@jamesturk, which is preferable:

- cleaning up the committee names so that they're within Open States norms (ie, removing trailing " Committee" or " Subcommittee" text)
- maintaining consistency with committee names already in the DB (which, for UT, did not remove trailing " Committee" or " Subcommittee" text)

In this PR I assume the former.


jamesturk:
that sounds right to me, merging this

PR #2094, opened 2018-01-16
SD: Get running for 2018

mileswwatkins:
Site is largely the same, but some significant tweaks that were breaking the scrapers in major ways.

After this PR is merged, the state must be re-activated in Bobsled's Task Definitions.


mileswwatkins:
Confirmed working locally; have spot-checked all fields for several Bills, Memberships, and VoteEvents.

PR #2103, opened 2018-01-21
NM: Handle failure of PDF vote scraping for individual House members

mileswwatkins:
Right now, [one](https://www.nmlegis.gov/Sessions/18%20Regular/votes/HB0001HVOTE.PDF) of the six vote PDFs for the state is failing. I've created [an issue](https://github.com/openstates/openstates/issues/2102) to address this down the line. Look like a good enough solution for now?


mileswwatkins:
Oh, and the validation of `vote_record['date']` is unnecessary given that `build_vote()` already runs `.strftime()` on that variable later on.

PR #2107, opened 2018-01-22
IA: Add independent party

mileswwatkins:
@jamesturk, just in case there're any extra steps to adding another state party `Organization` in production. This is running and importing properly for me locally, though.


jamesturk:
parties in metadata are deprecated, actually have a branch prepared to remove them altogether

PR #2108, opened 2018-01-22
MT: De-duplicate actions by adding an extra Pupa ID

mileswwatkins:
There are dup actions in some bills. Perhaps these are accidents, but they look like they _may_ be legitimately different actions, just with the same info. No other clear way to de-duplicate, and Pupa will choke on-`--import` otherwise.

Eg of problematic action lists:

- http://laws.leg.mt.gov/legprd/LAW0203W$BSRV.ActionQuery?P_SESS=20171&P_BLTP_BILL_TYP_CD=&P_BILL_NO=&P_BILL_DFT_NO=LC0267&P_CHPT_NO=&Z_ACTION=Find&P_SBJT_SBJ_CD=&P_ENTY_ID_SEQ=
- http://laws.leg.mt.gov/legprd/LAW0203W$BSRV.ActionQuery?P_SESS=20171&P_BLTP_BILL_TYP_CD=&P_BILL_NO=&P_BILL_DFT_NO=LC0400&P_CHPT_NO=&Z_ACTION=Find&P_SBJT_SBJ_CD=&P_ENTY_ID_SEQ=

Error: http://bobsled.openstates.org/run-MT-2018-01-21.html

### DIFF
--- a/openstates/mt/bills.py
+++ b/openstates/mt/bills.py
@@ -261,8 +261,7 @@ class MTBillScraper(Scraper, LXMLMixin):
         return bill, list(votes)
 
     def add_actions(self, bill, status_page):
-
-        for action in reversed(status_page.xpath('//div/form[3]/table[1]/tr')[1:]):
+        for idx, action in enumerate(reversed(status_page.xpath('//div/form[3]/table[1]/tr')[1:])):
             try:
                 actor = actor_map[action.xpath("td[1]")[0].text_content().split(" ")[0]]
                 action_name = action.xpath("td[1]")[0].text_content().replace(actor, ""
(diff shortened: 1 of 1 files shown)


jamesturk:
extras aren't used for deduping so I'm not sure if this would have any effect

mileswwatkins:
Ah, thanks for the clarification! Ideas, then, on how to proceed? Should we really just de-dup sequential actions that have the same text and date? I'd be fine with that for now, and leave a comment in with details.

mileswwatkins:
Just spoke with Pat Murdo, a drafter of one of the bills that has duplicate actions. She says that the duplicate actions are real, albeit often due to small hiccups in the legislative processing (eg, the wrong person was sent a pre-introduction letter, so it had to be sent again, bookkept as a separate row in the Bill Actions table).

mileswwatkins:
Switched to just throwing out dups.

jamesturk:
in light of my investigation (opencivicdata/pupa#307), let's revert this change- no issue w/ duplicate actions

Issue #2105, opened 2018-01-21
CT: Sponsors includes extraneous text


mzagaja:
State: CT (be sure to include in ticket title)

Pulling down 2018 data from the API I'm getting JSON like:

```JSON
{
  "id": "CTB00017146",
  "type": [
    "joint resolution"
  ],
  "state": "ct",
  "title": "RESOLUTION CONCERNING THE EXPENSES OF THE JANUARY SPECIAL SESSION, 2018.",
  "votes": [],
  "actions": [],
  "all_ids": [
    "CTB00017146"
  ],
  "bill_id": "HJ 403",
  "chamber": "lower",
  "session": "2017",
  "sources": [
    {
      "url": "ftp://ftp.cga.ct.gov/pub/data/bill_info.csv"
    },
    {
      "url": "http://www.cga.ct.gov/asp/cgabillstatus/cgabillstatus.asp?selBillType=Bill&bill_num=HJ403&which_year=2017"
    }
  ],
  "sponsors": [
    {
      "name": "b'Candelora, Vincent J.'",
      "type": "primary",
      "leg_id": null,
      "committee_id": null
    },
    {
      "name": "b'Duff, Bob'",
      "type": "primary",
      "leg_id": null,
      "committee_id": null
    },
    {
      "name": "b'Witkos, Kevin D.'",
      "type": "primary",
      "leg_id": null,
      "committee_id": null
    },
    {
      "name": "b'Klarides, Themis'",
      "type": "primary",
      "leg_id": null,
      "committee_id": null
    },
    {
      "name": "b'Ritter, Matthew'",
      "type": "primary",
      "leg_id": null,
      "committee_id": null
    }
  ],
  "subjects": [],
  "versions": [
    {
      "url": "ftp://ftp.cga.ct.gov/2017/tob/h/2017HJ-00403-R00-HB.htm",
      "name": "R00",
      "+date": null,
      "doc_id": "CTD00040674",
      "mimetype": "text/html"
    }
  ],
  "documents": [],
  "companions": [],
  "created_at": "2018-01-04 05:37:00",
  "updated_at": "2018-01-05 05:37:37",
  "action_dates": {
    "last": null,
    "first": null,
    "signed": null,
    "passed_lower": null,
    "passed_upper": null
  },
  "alternate_titles": [],
  "alternate_bill_ids": []
}
```

The bill sponsor names now have a `b'` before them and a bonus `'` after them.

PR #2127, opened 2018-02-05
CT: Decode sponsor names in Bills

pll33:
Addresses #2105 

Convert byte strings back to strings by decoding utf-8

Alternatively: skip encoding `name` at [L337](https://github.com/openstates/openstates/blob/5edbb75a5f99237f4589f0c5fe807878b7a6859b/openstates/ct/bills.py#L337)


mileswwatkins:
Good catch @pll33!

PR #2134, opened 2018-02-08
VA: Handle empty actions, add logging

mileswwatkins:
Let me know if this is the right way to handle logging in the `va` module, or if I'm off the path. Thanks for the pointers!

### DIFF
--- a/openstates/va/__init__.py
+++ b/openstates/va/__init__.py
@@ -1,9 +1,13 @@
+import logging
 from pupa.scrape import Jurisdiction, Organization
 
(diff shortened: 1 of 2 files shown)


jamesturk:
logging.getLogger(__name__) returns the logger to use for that module

https://docs.python.org/3/library/logging.html has an explanation worth reading, but logging.warning would use a different logger than logging.getLogger('va').warning

mileswwatkins:
Lemme know if this is correct now, I believe this is the approach we talked about earlier.

PR #2153, opened 2018-02-19
2018 session added

sudhanshuchopra:


mileswwatkins:
@sudhanshuchopra, you also have to add the session in the `./billy_metadata/wy.py` file, both in the `session_details` array and in the appropriate `terms.sessions` property.

billy_metadata/wy.py
```diff
@@ -33,6 +33,13 @@
             'end_year': 2018,
             'sessions': ['2017'],
         },
+        {
+            'name': '2018-2019',
```
mileswwatkins: The `2018` session is part of the `2017-2018` term, not a standalone `term` in and of itself.

Issue #2154, opened 2018-02-19
ME: Incorrectly handling deceased legislator and her replacement


mileswwatkins:
(Moving over @yocheved's issue from the docs repo: https://github.com/openstates/documentation/issues/48)

Gina Mason died in 2017, and has been replaced by Richard Mason. However, Open States is not capturing Richard as the active legislator for Maine's House of Representatives seat 56.

Here's where the state site shows Gina and Richard's information, including death date and special election information. This is the webpage that we scrape, so it looks like we need to make an adjustment to the People scraper so that it can handle deceased and replacements. http://legislature.maine.gov/house/dist_mem.htm

sriks123:
@mileswwatkins do we have to store both Gina M as deceased and Richard M as active

mileswwatkins:
No, Gina Mason is not active; only Richard Mason should be scraped.

sriks123:
Yes I have implemented accordingly and made a PR can you review it?

PR #2155, opened 2018-02-20
ME: Added support for deceased members #2154

sriks123:
Reolves issue #2154

### DIFF
--- a/openstates/me/people.py
+++ b/openstates/me/people.py
@@ -62,12 +62,19 @@ class MEPersonScraper(Scraper):
             info_search = re.search(INFO_RE, leg_info, re.VERBOSE)
 
             if not info_search:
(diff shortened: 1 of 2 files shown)


sriks123:
@jamesturk please review this PR.

mileswwatkins:
It works, merging it in. Not super-happy that we have to special-case this in such a kluge-y way, but I noticed that the House does offer a spreadsheet with structured data, and that we could just scrape that; created a new, separate issue: https://github.com/openstates/openstates/issues/2158

mileswwatkins:
By the way, @sriks123, next time you have a small PR that contains many commits, could you git-squash them down into a single one? Just to avoid having a messy Git history.

sriks123:
Thanks for merging and yeah I will keep that in mind.

Issue #1872, opened 2017-07-23
MS: poorly-formatted district addresses


mileswwatkins:
Mississippi legislator addresses (both chambers) appear correctly-formatted for the capitol. However, for district offices, many (or all) are in the format of:

```
${street address}

${city}
${ZIP code}
```

when it should naturally be

```
${street address}
${city}, MS ${ZIP code}
```

narendra36:
Can I work on this ?

mileswwatkins:
closed by https://github.com/openstates/openstates/pull/2160

PR #2160, opened 2018-02-21
 Fixed MS: poorly-formatted district addresses #1872 

narendra36:
Issue MS: poorly-formatted district addresses #1872 fixed and bad formatting of auther's link(was giving 404 error) corrected.

### DIFF
--- a/openstates/ms/bills.py
+++ b/openstates/ms/bills.py
@@ -97,7 +97,7 @@ class MSBillScraper(Scraper):
                 main_sponsor = main_sponsor[0]
                 main_sponsor_link = details_root.xpath('string(//P_LINK)').replace(" ", "_")
                 main_sponsor_url = ('http://billstatus.ls.state.ms.us/%s/'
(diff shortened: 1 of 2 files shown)


narendra36:
I have made necessary changes that was told by @jamesturk in my previous pull request which I have closed. Please review this one also. @jamesturk @mileswwatkins

mileswwatkins:
Thanks, @narendra36! Working well, but a few notes on how to improve the semantics:

- a more Pythonic way to express `if home_address != "":` is to say `if home_address:`; likewise, instead of stating `if home_address == "":` you can say `if not home_address:`
- try using Python's `elif` construction to make the conditional logic more compact; right now, you have `if … else if… else`
- if a legislator is missing either `home_address` or `home_city`, I'd say that we should leave `home_address_total` empty, since there isn't a full mailing address available; this should let you simplify your logic

narendra36:
Thanks @mileswwatkins for review :) 
Check it now I have improved syntax and logic as you suggested.

PR #2162, opened 2018-02-21
fix for issue 2159

### LINKED ISSUE #2159: OH: `get_session_list` cannot "see" current session
Ohio's `get_session_list` appears to have the current session _manually appended_. Not to mention, it hasn't been updated in a year. This should be fixed, to automatically "guess" the current session instead.

https://github.com/openstates/openstates/blob/master/openstates/oh/__init__.py#L91-L92

shreyakupadhyay:
Adding the current session dynamically and removing manually added current session. This fix is for Ohio scraper.

fixes #2159

### DIFF
--- a/openstates/oh/__init__.py
+++ b/openstates/oh/__init__.py
@@ -46,14 +46,7 @@ class Ohio(Jurisdiction):
             "end_date": "2017-12-31"
         }
     ]
-    ignored_scraped_sessions = [
-        "127",
-        "126",
-        "125",
-        "124",
-        "123",
-        "122"
-    ]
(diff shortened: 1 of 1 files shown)


mileswwatkins:
@shreyakupadhyay, think there could be a simpler way, see my note in the GitHub Issue: https://github.com/openstates/openstates/issues/2159#issuecomment-367467477

mileswwatkins:
Great, and one last thing; you can now empty the `ignored_scraped_sessions` list. That list contains session IDs that are scraped but not included in `legislative_sessions`.

PR #2164, opened 2018-02-21
Fix NY Assembly member missing emails.

tcarobruce:
Also, fix a bug which caused members without emails to be assigned the previously-processed email address.


mileswwatkins:
Thanks for the catch!

PR #2168, opened 2018-02-22
Fix missing email addresses for OK state house.

tcarobruce:
Emails are no longer on legislator pages for reps in the lower chamber, but can be found on the contact form.

### DIFF
--- a/openstates/ok/people.py
+++ b/openstates/ok/people.py
@@ -47,10 +47,10 @@ class OKPersonScraper(Scraper, LXMLMixin, LXMLMixinOK):
             email_node = page.get_element_by_id('txtMemberEmail')
             email = email_node.value
         except KeyError:
-            email = ''
+            email = None
 
-        if not validate_email_address(email):
-            email = ''
+        if email and not validate_email_address(email):
+            email = None
 
         return email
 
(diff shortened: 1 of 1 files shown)


openstates/ok/people.py
```diff
@@ -38,23 +38,21 @@ def _extract_phone(self, office_info):
...
+    def _get_rep_email(self, district):
+        # Get the email address from an input on the contact page
+        url = "https://www.okhouse.gov/Members/Contact.aspx?District=" + district
+        page = self.curl_lxmlize(url)
+
+        try:
+            email_node = page.get_element_by_id('txtMemberEmail')
+            email = email_node.value
+        except KeyError:
+            email = ''
```
mileswwatkins: Can we use `None` instead of `''` here? Doesn't matter much, but it'll be clearer to others later that we weren't, say, planning to append anything to the empty string variable later.

mileswwatkins:
Web contact forms are just the bee's knees.

Just that one request, then this is good to merge in.

tcarobruce:
I made that change -- let me know if you think anything else needs fixing.

mileswwatkins:
LGTM

PR #2171, opened 2018-02-24
Add code to load FL house emails from their PDF directory.

samtregar:
Note that if you take this PR you may also need to disable whatever is currently overriding the scraper to return some email addresses from the API (ex: FLL000150).

### DIFF
--- a/openstates/fl/__init__.py
+++ b/openstates/fl/__init__.py
@@ -1,10 +1,13 @@
 # encoding=utf-8
+import logging
 from pupa.scrape import Jurisdiction, Organization
 from .bills import FlBillScraper
 from .people import FlPersonScraper
 from .committees import FlCommitteeScraper
 # from .events import FlEventScraper
 from openstates.utils import url_xpath
+
+logging.getLogger(__name__).addHandler(logging.NullHandler())
 
 
 class Florida(Jurisdiction):
--- a/openstates/fl/people.py
+++ b/openstates/fl/people.py
@@ -9,6 +9,7 @@ from spatula import Spatula, Page
 
 log = logging.getLogger('fl')
 
+
 class SenDetail(Page):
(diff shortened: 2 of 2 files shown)


samtregar:
I can try!

samtregar:
@mileswwatkins fixed!

mileswwatkins:
Testing it now, but may not finish running in the short period I have. As a safeguard, can we have the scraper error out if it tries to assign any _one_ of the email addresses to _more than one_ legislator? Also, can we `.add_source` to the legislator if an email is applied from the PDF?

mileswwatkins:
Finished testing, and looks to be working well! Good to merge once these enhancements're made.

One more request: if a Rep's email is _not_ found in the PDF, can you add an informational log message noting this? May help debugging down the road, or handling other name special-casings (like `La ` -> `La`. And since there're few Reps whose emails aren't found, it won't litter the log. Check out how we set up logging in `openstates/va/__init__.py` and `openstates/va/bills.py`; since FL uses Spatula, you can't just do `self.log`.

samtregar:
@mileswwatkins Should be good to go now!

mileswwatkins:
Looks great! Seems like there's still one last linting error, after that's patched I'll merge. (On my end, I integrated Flake8 into my code editor to avoid finding about about them post-commit.)

samtregar:
@mileswwatkins The error isn't from this PR, oddly.  Maybe a version change in flake8?  Here it is:

openstates/vt/bills.py:181:100: E501 line too long (106 > 99 characters)

mileswwatkins:
Oh, I think it was from @showerst's hotfix earlier today. Shame on you ;)

showerst:
Whoops didn't notice the test fail when I committed, mea culpa!

PR #2172, opened 2018-02-24
ME: Switched to spreadsheets for scraping lower-chamber representatives

### LINKED ISSUE #2158: ME: Switch to using spreadsheet data, for lower-chamber People scraping
Currently, we scrape HTML for Maine's House of Representatives. Scraping this HTML list has gotten a bit dicey as special cases pop up, such as a during-term special election to replace a deceased member:

screen shot 2018-02-20 at 11 26 53

Fortunately, the House publishes an Excel spreadsheet of members, that is much cleaner and already handles the special cases. We should switch to scraping that spreadsheet instead of parsing the HTML. The spreadsheet is available here: http://legislature.maine.gov/house/128house.xlsx

cc https://github.com/openstates/openstates/pull/2155

kush789:
Fixes #2158

### DIFF
--- a/openstates/me/people.py
+++ b/openstates/me/people.py
@@ -24,6 +24,22 @@ def clean_phone(phone):
     return phone
 
 
+def get_ascii_chunks(name):
+    ascii_chunks = []
+    current_chunk = ''
+
+    for ch in name:
+        if ord(ch) < 128:
+            current_chunk += ch
+        elif len(current_chunk) > 0:
+            ascii_chunks.append(current_chunk)
+            current_chunk = ''
+    if len(current_chunk) > 0:
+        ascii_chunks.append(current_chunk)
+
+    return ascii_chunks
+
+
 class MEPersonScraper(Scraper):
     def scrape(self, chamber=None):
         if chamber in ['upper', None]:
@@ -78,13 +94,6 @@ class MEPersonScraper(Scraper):
                     # str typecasting will raise error for non-ascii characters.
                     d[field] = sh.cell(rownum, col_num).value.encode('utf-8').strip()
 
-            try:
-                # xlrd reads this value as a float. However we save it as a string.
-                district_number = str(int(float(d['district'])))
-            except ValueError:
-                # Will reach here in case district number is blank. We will skip this row.
-                continue
-
             district_name = d['legal_res']
 
             first_name = d['first_name']
@@ -102,20 +111,41 @@ class MEPersonScraper(Scraper):
             member_name = " ".join((first_name, middle_name, last_name))
             member_name = re.sub(r'\s+', ' ', member_name).strip()
 
+            try:
+                # xlrd reads this value as a float. However we save it as a string.
+                district_number = "{:.0f}".format(float(d['district']))
+            except ValueError:
+                # Will reach here in case district number is blank. We will skip this row.
+                self.info("Skipping non voting member {}".format(member_name))
+                continue
+
             party = d['party']
             party = _party_map[party]
 
             # Determine legislator's URL to get their photo.
             # Roster URL list does not include nicknames, have to be removed before we search.
             if '"' in middle_name:
-                # Name contains nickname
-                roster_name = " ".join((first_name, ''.join(middle_name.split('"')[2:]),
-                                        last_name))
-                roster_name = re.sub(r'\s+', ' ', roster_name).strip()
+                # middle name contains nickname
+                roster_middle_name = ''.join(middle_name.split('"')[2:])
+                roster_middle_name = re.sub(r'\s+', ' ', roster_middle_name).strip()
             else:
-                roster_name = member_name
+                roster_middle_name = middle_name
 
-            URL_XPATH = '//a[contains(text(), "{}")]/@href'.format(roster_name)
+            # Searching for different parts of the name with an 'and' operator
+            # Since first_name middle_name last_name order not necessarily
+            # maintained in the roster
+            URL_XPATH = URL_XPATH = '//a[contains(text(), "Representative")'
+
+            # Get chunks of ascii characers in the name. lxml incorrectly decodes
+            # sometimes, hence skipping the non ascii parts of the names
+            for chunk in get_ascii_chunks(first_name):
+                URL_XPATH += ' and contains(text(), "{}")'.format(chunk)
+            for chunk in get_ascii_chunks(roster_middle_name):
+                URL_XPATH += ' and contains(text(), "{}")'.format(chunk)
+            for chunk in get_ascii_chunks(last_name):
+                URL_XPATH += ' and contains(text(), "{}")'.format(chunk)
+            URL_XPATH += ']/@href'
+
             try:
                 (leg_url, ) = roster_doc.xpath(URL_XPATH)
             except ValueError:


mileswwatkins:
Thanks @kush789. Please fix the linting errors found by Travis CI, and then we'll review this PR.

kush789:
Ah I see, that was the reason the build was breaking. Fixed it now

openstates/me/people.py
```diff
@@ -31,75 +31,99 @@ def scrape(self, chamber=None):
...
+            'house_leadership': 20,
+            'committees': 21
+        }
 
-            # Get the photo url.
-            html = self.get(leg_url).text
-            doc = lxml.html.fromstring(html)
-            doc.make_links_absolute(leg_url)
-            (photo_url,) = doc.xpath('//img[contains(@src, ".jpg")]/@src')
+        url = (
```
mileswwatkins: This can be on its own line, outside of parentheses. Also, make sure to `Person.add_source()` for this URL, so that it's noted as a primary data source in the database.

mileswwatkins: Also, it looks like this URL would always return 2017–2018 legislators. Can you replace `128` in the URL with `self.latest_session`?

openstates/me/people.py
```diff
@@ -31,75 +31,99 @@ def scrape(self, chamber=None):
...
+            d = {}
+            for field, col_num in mapping.items():
+                try:
+                    d[field] = str(sh.cell(rownum, col_num).value).strip()
+                except UnicodeEncodeError:
+                    # str typecasting will raise error for non-ascii characters.
+                    d[field] = sh.cell(rownum, col_num).value.encode('utf-8').strip()
+                except IndexError:
+                    # This col_num doesn't exist in the sheet.
+                    pass
```
mileswwatkins: Sounds like we should never expect a column to not exist, right? If the spreadsheet's structure has changed, we should need to rewrite `mapping`. If that's the case, then don't include this `except IndexError:` code block.

openstates/me/people.py
```diff
@@ -31,75 +31,99 @@ def scrape(self, chamber=None):
...
+
+            district_number = d['district']
```
mileswwatkins: If a member doesn't have a district (eg, the final two rows in the spreadsheet), they should be skipped.

openstates/me/people.py
```diff
@@ -31,75 +31,99 @@ def scrape(self, chamber=None):
...
+            leg_url = 'http://legislature.maine.gov/house/hsebios/'
+            leg_url += last_name[:4].lower() + first_name[0].lower()
+            if len(middle_name) > 0:
+                leg_url += middle_name[0].lower()
+            else:
+                leg_url += '_'
+            leg_url += '.htm'
+
+            # Determine legislator's URL to get their photo
+            photo_url = 'http://legislature.maine.gov/house/photo128/'
```
mileswwatkins: Same thing as above, with replacing `128` with `self.latest_session`.

openstates/me/people.py
```diff
@@ -31,75 +31,99 @@ def scrape(self, chamber=None):
...
+            jr_sr = d['jr_sr']
+
+            member_name = " ".join((first_name, middle_name, last_name, jr_sr))
+            member_name = re.sub(r'\s+', ' ', member_name).strip()
+
+            party = d['party']
+            party = _party_map[party]
+
+            # Determine legislator's URL
+            leg_url = 'http://legislature.maine.gov/house/hsebios/'
```
mileswwatkins: Does this pattern _always_ work? If it's iffy, might be worth adding something like this afterwards, to help debug or build in the future:

```
try:
    self.head(leg_url)
except scrapelib.HTTPError:
    self.warning("Could not correctly guess legislator URL for {}".format(member_name))
```

openstates/me/people.py
```diff
@@ -31,75 +31,99 @@ def scrape(self, chamber=None):
...
+            leg_url += '.htm'
+
+            # Determine legislator's URL to get their photo
+            photo_url = 'http://legislature.maine.gov/house/photo128/'
+            photo_url += last_name[:4].lower() + first_name[0].lower()
+            if len(middle_name) > 0:
+                photo_url += middle_name[0].lower()
+            else:
+                photo_url += '_'
+            photo_url += '.jpg'
```
mileswwatkins: I'd suggest the same pattern as above, to use `self.head(photo_url)` to confirm that this photo actually exists.

mileswwatkins: Also, it seems like you could cut down on code, if the same `last_name[:4].lower() + first_name[0].lower() + (middle_name[0].lower() or '_')` exists for both leg _and_ photo URLs.

openstates/me/people.py
```diff
@@ -31,75 +31,99 @@ def scrape(self, chamber=None):
...
+            office_email = d['email']
+            business_phone = d['business_phone']
+            home_phone = d['home_phone']
+            cell_phone = d['cell_phone']
 
             person = Person(
                 name=member_name,
                 district=district_number,
-                primary_org='lower',
+                primary_org='upper',
```
mileswwatkins: Check yourself before you wreck yourself.

openstates/me/people.py
```diff
@@ -119,30 +143,30 @@ def scrape_reps(self):
...
                         'House GOP Office, Room 332 State House, 2 State House Station, '
                         'Augusta, Maine 04333-0002'
                     )
-
                     person.add_contact_detail(
                         type='address', value=leg_address, note='Party Office')
 
             if office_email:
-                office_email = office_email[0]
                 person.add_contact_detail(type='email', value=office_email, note='District Office')
```
mileswwatkins: Seems like these are `Capitol Office` email addresses. they all end in `legislature.maine.gov`.

openstates/me/people.py
```diff
@@ -119,30 +143,30 @@ def scrape_reps(self):
```
mileswwatkins: Looks like every legislator in this spreadsheet has an address. I'm fine if you chop this "default address" section out, and just don't assign an address to the legislator if it isn't present.

openstates/me/people.py
mileswwatkins: Not your purview, but can you also replace `128` with `self.latest_session` for the upper-house scraping URL, later in this Python module?

kush789: The url for the spreadsheet for the upper-house has been placed inside parentheses. Is there a reason behind that? Is it because the line is crossing a certain number of characters and needs to be split?

openstates/me/people.py
kush789: This should be the case for the upper-house scraper as well (later in the module), right?

openstates/me/people.py
kush789: Oops, this was quite careless of me X_X

openstates/me/people.py
kush789: Okay, I realised this is a **very** bad idea. Works most of the times, but sometimes the urls are being formed randomly, taking different length combinations from different parts of the name. I've changed my approach, falling back on the way it was being done previously for the upper house, ie, finding the url from the house roster. This seems to work for everyone, other than for Owen D. Casás. This has something to do with the special character, because of which lxml cannot find the element

kush789:
My bad, I thought I was ignoring a postgres error. Have made changes to the module, seems to be running for me now, please have a look

kush789:
I tried to figure out why it was breaking for the one case (Representative Owen D. Casás). The problem is that the last name in the roster (http://legislature.maine.gov/house/dist_mem.htm) is different than what is there in the spreadsheet.

Name in roster (After lxml parses it) -> CasÃ¡s (Owen D. Cas\xc3\x83\xc2\xa1s)
Name in spreadsheet -> Casás (Owen D. Cas\xc3\xa1s)

mileswwatkins:
@kush789, so the issue you're hitting with Casás is that the encodings on his names don't match? Is there a solution you'd suggest to handling that?

openstates/me/people.py
```diff
@@ -31,67 +32,112 @@ def scrape(self, chamber=None):
...
+            for field, col_num in mapping.items():
+                try:
+                    d[field] = str(sh.cell(rownum, col_num).value).strip()
+                except UnicodeEncodeError:
+                    # str typecasting will raise error for non-ascii characters.
+                    d[field] = sh.cell(rownum, col_num).value.encode('utf-8').strip()
+
+            try:
+                # xlrd reads this value as a float. However we save it as a string.
+                district_number = str(int(float(d['district'])))
```
mileswwatkins: Is there cleaner syntax for this?

openstates/me/people.py
```diff
@@ -31,67 +32,112 @@ def scrape(self, chamber=None):
...
+            except ValueError:
+                # Will reach here in case district number is blank. We will skip this row.
```
mileswwatkins: Can you add a log here, to let the user know that we're skipping the row? Probably a `self.info`, since it's "expected" that we'd skip non-voting members.

openstates/me/people.py
```diff
@@ -73,41 +77,60 @@ def scrape_reps(self):
...
                     # str typecasting will raise error for non-ascii characters.
                     d[field] = sh.cell(rownum, col_num).value.encode('utf-8').strip()
-                except IndexError:
-                    # This col_num doesn't exist in the sheet.
-                    pass
 
-            district_number = d['district']
+            try:
+                # xlrd reads this value as a float. However we save it as a string.
+                district_number = str(int(float(d['district'])))
```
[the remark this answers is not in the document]
kush789: The content of d['district'] is something like '121.0'.
Calling an int() typecast on it will not work, since the string has a decimal. It needs to be first converted to a float, then typecast it to an int, and then back to a str since we save it that way. Hence the complicated syntax.

We could just truncate the last two characters in the string ('.0'), but I'm not sure whether that'll be straightforward to understand if someone's looking at the code sometime in the future. What do you think?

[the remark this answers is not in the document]
mileswwatkins: Python has some powerful string formatting, check out https://stackoverflow.com/a/15263885

eg: `'{:.0f}'.format(23.214523)`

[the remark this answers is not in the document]
mileswwatkins: More holistic: https://pyformat.info/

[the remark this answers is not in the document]
kush789: Ahh, this is perfect. Didn't think of it, thanks!

kush789:
@mileswwatkins For a situation like Casás, we can skip out chunks of special characters in the name and use 'and' while matching. This should be able to generalise well, I believe.

mileswwatkins:
If the character encodings were the same, then would we still need to strip non-ASCII characters? Python 3's natively-Unicode, so it should be able to handle equalities like this gracefully. Or does the legislature's data actually differ in terms of the characters displayed, not just in terms of character code points?

kush789:
The system actually breaks because of the way lxml is parsing the html. The name appears to be the same in the spreadsheet and the online roster (when I copy and compare both of them). However, in the lxml parsed tree, the encoding is different (it adds 2 extra bytes for some reason). It seems to be more of a lxml bug, rather than a difference in the legislature's data. 

In this kind of a situation, I think this is the way to go. We'll never actually be able to tell beforehand when such a situation comes up. We could try this approach in case a normal full name string matching fails, what do you think?

mileswwatkins:
Okay, that makes sense to me; make sure to add in a verbose comment to explain why we're doing this, so that it makes sense to other devs in the future (including you and me).

kush789:
Another minor detail, the (first_name second_name last_name) order is not necessarily maintained in the roster online. So rather than combining and searching, I'm doing a search for each of the three parts with an 'and'. a

Added comments explaining what's happening, and referred to this PR in the comments as well.

openstates/me/people.py
```diff
@@ -102,20 +111,41 @@ def scrape_reps(self):
...
+                roster_middle_name = ''.join(middle_name.split('"')[2:])
+                roster_middle_name = re.sub(r'\s+', ' ', roster_middle_name).strip()
             else:
-                roster_name = member_name
+                roster_middle_name = middle_name
+
+            # Searching for different parts of the name with an 'and' operator
+            # Since first_name middle_name last_name order not necessarily
+            # maintained in the roster
+            URL_XPATH = URL_XPATH = '//a[contains(text(), "Representative")'
```
mileswwatkins: Double-assignment is probably a typo.

openstates/me/people.py
```diff
@@ -102,20 +111,41 @@ def scrape_reps(self):
...
+                URL_XPATH += ' and contains(text(), "{}")'.format(chunk)
+            for chunk in get_ascii_chunks(roster_middle_name):
+                URL_XPATH += ' and contains(text(), "{}")'.format(chunk)
+            for chunk in get_ascii_chunks(last_name):
+                URL_XPATH += ' and contains(text(), "{}")'.format(chunk)
+            URL_XPATH += ']/@href'
 
-            URL_XPATH = '//a[contains(text(), "{}")]/@href'.format(roster_name)
             try:
                 (leg_url, ) = roster_doc.xpath(URL_XPATH)
```
mileswwatkins: Programmatically constructing XPath query strings isn't the best practice, and could be hard to read and debug. How about something very similar, but that does the filtering ex-post?

```
roster_name_chunks = get_ascii_chunks(first_name) + get_ascii_chunks(roster_middle_name) + get_ascii_chunks(last_name)
leg_url_elements = roster_doc.xpath('//a[contains(text(), "Representative")]')
leg_url = None
for el in leg_url_elements:
    if all(chunk in el.text_content() for chunk in roster_name_
chunks):
        # Avoid accidentally repeat-matching URLs
        assert leg_url is None
        leg_url = el.attrib['href']
if leg_url is None:
    self.warning("Could not find legislator URL for {}".format(member_name))
```

mileswwatkins: I didn't 100% check that code, but the flow should be right, maybe look for syntax issues before using ;)

mileswwatkins:
@kush789, are you still working on this PR?

PR #2185, opened 2018-03-03
NY: Fix Assembly vote scraping

### LINKED ISSUE #1960: NY: No 2017 bills contain any Assembly votes
There are separate issues (#1384 , #1831) about missing roll calls.  This is about there being no 2017 Assembly votes at all in any bills.  I presume the other issues will remain after this is fixed, hence the new issue.

schedutron:
This is my first iteration to fix #1960. This sort of reverts [c29e88790b3f462c](https://github.com/openstates/openstates/commit/c29e88790b3f462c854539de9bc7144928ed2517#diff-c0b833d744907b858a610c3983ecd2c8L58) for `models.py`. I've tested each part of the new scraper on the interpreter, but I'm unsure about whether we need to scrape committee votes as well. As of now, this just scrapes individual floor votes.

### DIFF
--- a/openstates/ny/models.py
+++ b/openstates/ny/models.py
@@ -4,7 +4,7 @@ import re
 import inspect
 
 from itertools import islice
-  
+
 from billy.scrape.votes import Vote
 from openstates.utils import LXMLMixin
 from .utils import Urls
@@ -69,7 +69,7 @@ class AssemblyBillPage(LXMLMixin):
             pre = doc.xpath('//pre')[0].text_content().strip()
 
             no_votes = 'There are no votes for this bill in this legislative '
-                       'session.'
+            'session.'
 
             if pre == no_votes:
                 raise ValueError('No votes for this bill.')
--- a/openstates/ny/models.py
+++ b/openstates/ny/models.py
@@ -1,8 +1,12 @@
+import collections
+import datetime
 import re
 import inspect
 
+from itertools import islice
+  
+from billy.scrape.votes import Vote
 from openstates.utils import LXMLMixin
-
 from .utils import Urls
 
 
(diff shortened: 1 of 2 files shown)


openstates/ny/models.py
```diff
@@ -34,20 +38,83 @@ def __init__(self, scraper, session, bill, details):
             'senate': senate_url})
 
     def _scrub_name(self, name):
-        junk = [
+        junk = (re.compile(rgx) for rgx in [
```
mileswwatkins: I don't think the tweak of this regex flow makes the code more readable. Is it necessary to change for functionality's sake? I'd be happy to keep it how it currently is, otherwise.

mileswwatkins:
@schedutron, are you still working on this PR?

schedutron:
Sorry for getting late on this, but yes, I'm still working on this.

estaub:
@schedutron, I was thinking about working on the NY Assembly votes, but see that it looks like you're most of the way there!  Are you still on it?

schedutron:
@estaub, I'll give it one final try, and will let you know if I'm unable to do it. Just give me a week :)

schedutron:
@estaub The problem I'm facing is that I get an `ImportError` for `from pupa.scrape import VoteEvent`, which has persisted for long. If you'd like to work on this now, please go ahead.

estaub:
@schedutron 
This may not be a real error, depending on exactly what you're doing and what it's saying.
You may know a lot more or a lot less than I do, so bear with me!

How are you running the scraper?  Using docker-compose?  

Does the error look like it's trying to connect to a postgresql database?  If so, try running with `--scrape` on the end of the command line.  (There's an example of it on the readme.)  If that succeeds, your code may be just fine!

You may want to post more of the error; other folks (or I) may have better ideas.

PR #2186, opened 2018-03-03
New Vote Scraper for MD

### LINKED ISSUE #2093: MD: Rewrite vote scraping
Looks like Maryland has started serving its floor votes as PDFs, rather than HTML. For example:

- 2018 bill: http://mgaleg.maryland.gov/webmga/frmMain.aspx?pid=billpage&stab=02&id=hb0001&tab=subject3&ys=2018RS
- bill from earlier session: http://mgaleg.maryland.gov/webmga/frmMain.aspx?pid=billpage&stab=02&id=sb0001&tab=subject3&ys=2016rs

So, I've stripped out the vote scraping for now. It will need to be rewritten from scratch.

Note that while the source table _does_ provide vote result, vote counts, and date, this _cannot_ be used to build a `VoteEvent` in the meantime, since the date provided is the legislative day, not the calendar day.

In-vincible:
solves issues close #2093 and #2188, I am running few more tests and then I will fix lint errors.

### DIFF
--- a/openstates/md/bills.py
+++ b/openstates/md/bills.py
@@ -171,7 +171,6 @@ class MDBillScraper(Scraper):
     
     def parse_bill_votes_new(self, doc, bill):
         elems = doc.xpath('//a')
-        print("shito")
         for elem in elems:
             href = elem.get('href')
             if href and "votes" in href and href.endswith('pdf') and ("Senate" in href or "House" in href):
@@ -191,23 +190,18 @@ class MDBillScraper(Scraper):
             chamber = 'lower'
         
         date = dparse(lines[0].split('Date:')[1], fuzzy=True)
-        print(date)
         countIndex=-1
         for index,line in enumerate(lines):
             if 'Yeas' in line and 'Nays' in line:
                 countIndex = index
                 break
-        print("CountIndex = "+str(countIndex))
         yes_count = -1
         no_count = -1
         other_count = 0
         if countIndex > 0:
             counts = re.split(r'\s{2,}', lines[countIndex].strip())
-            print(counts)
             for count in counts:
-                print(count)
                 number,string = count.split(' ',1)
-                print(number)
                 number = int(number)
                 if string == 'Yeas':
                     yes_count = number
@@ -215,7 +209,6 @@ class MDBillScraper(Scraper):
                     no_count = number
                 else:
                     other_count += number
-        print(yes_count,no_count,other_count)
         passed = yes_count > no_count
         motion = "No motion given"
         vote = VoteEvent(
@@ -231,42 +224,56 @@ class MDBillScraper(Scraper):
         vote.set_count('no', no_count)
         vote.set_count('other', other_count)
         pageIndex = countIndex+2
+        test_yes = 0
+        test_no = 0
+        test_other = 0
         while pageIndex < len(lines):
-            if 'Voting Yea' in lines[pageIndex]:
+            if 'Voting Yea' in lines[pageIndex] or lines[pageIndex].strip() == '':
                 pageIndex += 1
                 continue
             if 'Voting Nay' in lines[pageIndex]:
                 pageIndex += 2
                 break
-            names = lines[pageIndex].strip().split()
-            print(names)
+            names = re.split(r'\s{2,}', lines[pageIndex].strip())
             for name in names:
                 vote.yes(name)
+            test_yes += len(names)
             pageIndex += 1
         while pageIndex < len(lines):
             if 'Not Voting' in lines[pageIndex]:
-                pageIndex += 2
+                pageIndex += 1
                 break
-            names = lines[pageIndex].strip().split()
-            print(names)
+            elif lines[pageIndex].strip() == '':
+                pageIndex += 1
+                continue
+            names = re.split(r'\s{2,}', lines[pageIndex].strip())
+            test_no += len(names)
             for name in names:
                 vote.no(name)
             pageIndex += 1
+        Names = []
         while pageIndex < len(lines):
-            if 'Excused from Voting' in lines[pageIndex] or 'Excused (Absent)' in lines[pageIndex]:
+            if 'Excused from Voting' in lines[pageIndex] or 'Excused (Absent)' in lines[pageIndex] or '*' in lines[pageIndex] or lines[pageIndex].strip() == '':
                 pageIndex += 1
                 continue
-            names = lines[pageIndex].strip().split()
-            print(names)
+            names = re.split(r'\s{2,}', lines[pageIndex].strip())
+            test_other += len(names)
+            Names.extend(names)
             for name in names:
-                vote.other(name)
+                vote.vote('other', name)
             pageIndex += 1
+        if test_yes != yes_count:
+            print("Yes Failed::::::",test_yes,yes_count)
+        if test_no != no_count:    
+            print("No Failed::::::", test_no, no_count)
+        if test_other != other_count:
+            print("Other Failed::::::", test_other, other_count)
+            print(Names)
         return vote
 
     def parse_vote_page(self, vote_url, bill):
         vote_html = self.get(vote_url).text
         doc = lxml.html.fromstring(vote_html)
-        print(vote_url)
         # chamber
         if 'senate' in vote_url:
             chamber = 'upper'
@@ -368,7 +375,6 @@ class MDBillScraper(Scraper):
         html = self.get(url).text
         doc = lxml.html.fromstring(html)
         doc.make_links_absolute(url)
-        print("#####yes###")
         try:
             title = doc.xpath('//h3[@class="h3billright"]')[0].text_content()
             # TODO: grab summary (none present at time of writing)
--- a/openstates/md/bills.py
+++ b/openstates/md/bills.py
@@ -4,6 +4,9 @@ import datetime
 import lxml.html
 from pupa.scrape import Scraper, Bill, VoteEvent
 
+from dateutil.parser import parse as dparse
+from pupa.utils.generic import convert_pdf
+
 CHAMBERS = {
     'upper': ('SB', 'SJ'),
     'lower': ('HB', 'HJ'),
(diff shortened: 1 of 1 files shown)


openstates/md/bills.py
```diff
@@ -165,10 +168,105 @@ def parse_bill_votes(self, doc, bill):
...
+        filename, response = self.urlretrieve(vote_url)
+        text = convert_pdf(filename, type='text').decode()
+        lines = text.splitlines()
+
+        if 'senate' in vote_url:
+            chamber = 'upper'
+        else:
+            chamber = 'lower'
+
+        date = dparse(lines[0].split('Date:')[1], fuzzy=True)
```
mileswwatkins: We try to avoid `dateutil`, since when provided with messy legislative dates, it can make incorrect assumptions about how to parse them. Are the dates all in the same format, so that we could `datetime.datetime.strptime` them?

In-vincible:
Will push one more commit fixing the old scraper.

openstates/md/bills.py
mileswwatkins: In general, it's also ideal to not add any dependencies to the project. If this were to be absolutely necessary, the PR would need an explanation, and to touch `requirements.txt` as well.

In-vincible:
Yes man, actually I am working on 3 different PRs, forgot to use different branch name for this one, others are fine.

And automatic issue closing is really cool add to my knowledge. :D

Actually I checked the docker file found dateutil there, was not sure though, but then running the ```pupa update --scrape``` didn't require any more installation of dateutil...but no problem i can totally extract date with a simple regex command, will push changes in next commit.

mileswwatkins:
Huh, had no idea that `dateutil` was installed by the Dockerfile. I think that's for the `billy` environment, rather than the OCD/Pupa one, but not 100% sure. And for stripping date, use `datetime.datetime.striptime` rather than regex, since that creates a native date object; there are pretty good docs and examples around on it.

In-vincible:
@mileswwatkins Yes dateutil is only in billy, mistake from my side. datetime.
I think this is ready for next round of review.
resolves #2093

mileswwatkins:
It looks like you're linting to a line length of 80 characters, and making some changes to existing code to enforce that. You can update your linter to use the `setup.cfg` file, which dictates a 99-character line limit, and undo the linting alterations you've made. https://github.com/openstates/openstates/blob/master/setup.cfg

Reviewing the output now.

In-vincible:
@mileswwatkins ohh my editor did it, recently installed this auto formatting extension which fixes linting automatically, really sorry I will fix this right away.

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
                 vote.add_source(href)
                 yield vote
 
+    def parse_bill_votes_new(self, doc, bill):
+        elems = doc.xpath('//a')
```
mileswwatkins: This `//a` query is very broad; can it be better-scoped?

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
...
+
+        # MD has a habit of listing votes twice
+        seen_votes = set()
+
+        for elem in elems:
+            href = elem.get('href')
+            if (href and "votes" in href and href.endswith('pdf') and
+                    ("Senate" in href or "House" in href) and href not in seen_votes):
```
mileswwatkins: Here you check for `Senate` and `House`; later you assign chamber based on `senate` and `house`. Which capitalization test is correct?

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
...
+
+        text = convert_pdf(filename, type='text').decode()
+        lines = text.splitlines()
+
+        if 'senate' in vote_url:
+            chamber = 'upper'
+        else:
+            chamber = 'lower'
+
+        date_string = lines[0].split('Date:')[1].strip()
```
mileswwatkins: Can you ensure that you're getting `Calendar Date` and not `Legislative Date`? Seems like the `lines[0]` should take care of that, but just in case. Legislative dates are very funky.

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
...
+        date = datetime.datetime.strptime(date_string, "%b %d, %Y %I:%M (%p)")
+
+        pageIndex = -1
```
mileswwatkins: You can start with it at `None` instead, I think that'd be more Pythonic.

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
...
+        for index, line in enumerate(lines):
+            if 'Yeas' in line and 'Nays' in line:
+                pageIndex = index
+                break
+
+        yes_count = -1
```
mileswwatkins: Why do these counters start at -1? If this is necessary, please add a comment to clarify.

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
...
+                number = int(number)
+
+                if string == 'Yeas':
+                    yes_count = number
+                elif string == 'Nays':
+                    no_count = number
+                else:
+                    other_count += number
+
+        passed = yes_count > no_count
```
mileswwatkins: This will be erroneous/nonsensical if `pageIndex` was never set, right?

openstates/md/bills.py
mileswwatkins: Also, follow Pythonic variable naming, underscores not camelCase.

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
...
+
+        motion = re.split(r'\s{2,}', lines[pageIndex-3].strip())[0]
+        if motion == '':
+            motion = "No motion given"
```
mileswwatkins: Add a warning here, at least. I might even be inclined to `assert motion`. Are there any cases where a motion isn't found, for the current session?

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
...
+                yes_names_count += len(names)
+            elif voteIndex == 1:
+                no_names_count += len(names)
+
+            totalvotes += len(names)
+            for name in names:
+                vote.vote(voteTypes[voteIndex], name)
+            pageIndex += 1
+
+        if yes_count != yes_names_count:
```
mileswwatkins: I'd argue that if any of these counts don't match, then we've failed our scrape, and should either 1) throw an error or 2) not record individuals' votes (ie, set the `Vote.yes`, `Vote.no`, etc lists to empty/null).

openstates/md/bills.py
```diff
@@ -165,10 +170,132 @@ def parse_bill_votes(self, doc, bill):
...
+            self.warning("Yes Votes Count and Number of Names don't match")
+        if no_count != no_names_count:
+            self.warning("No Votes Count and Number of Names don't match")
+        if totalvotes != yes_count+no_count+other_count:
+            self.warning(
+                "Total Votes Count and Total Number of Names don't match")
+
+        return vote
```
mileswwatkins: `yield`, not `return`, right? I think they'll actually both work here, but `yield` is more Pythonic since you're within a generator, I believe.

openstates/md/bills.py
mileswwatkins: It looks like the uppercase version is what should be checked for; I'm seeing falsely-assigned `lower` when I spot-check the data (eg, for http://mgaleg.maryland.gov/2018RS/votes/Senate/0378.pdf).

In-vincible:
@mileswwatkins yes, "excused","absent", etc can easily be categorized, on it.

openstates/md/bills.py
In-vincible: I can run manual tests, and confirm if it is picking legislative date anywhere, other than that do you think I have to make any change in code as well?

openstates/md/bills.py
In-vincible: Okay

openstates/md/bills.py
In-vincible: I prefered -1, because if some case occurs with a bad pdf design, it will throw an error as yes_count in model only allows value 0 or positive.

openstates/md/bills.py
In-vincible: Yes, but an error will be thrown already as they will be negative, and if they are not negative its very unlikely that it will be erroneous, for that there has to be "Yeas" and "Nays" in one single line, with non sensical number at perfectly those positions, in any other case it will throw error. either when int() is used or when yes_count or no_count is negative.

openstates/md/bills.py
In-vincible: +1

openstates/md/bills.py
In-vincible: Yes, here actually I was throwing error only locally, thought we may want to keep running it. Doing it right away.

openstates/md/bills.py
In-vincible: Actually we need to use ```return``` here, because when we use ```yield``` it doesn't allow us to add source in the parent function, which return allows.

openstates/md/bills.py
In-vincible: @mileswwatkins I didn't get this one.

openstates/md/bills.py
In-vincible: Okay, I did this one to the table containing bill docs.

openstates/md/bills.py
In-vincible: Got it Fixed!!

openstates/md/bills.py
In-vincible: More I can check for existence of word ```calendar``` in ```lines[0].split('Date:')[0]```, will it do?

openstates/md/bills.py
mileswwatkins: Okie dokie

openstates/md/bills.py
mileswwatkins: Yeah, can you `split('Calendar Date:')`, or something like that?

openstates/md/bills.py
mileswwatkins: You shouldn't rely on the model to catch bad vote counts that way; you should do it yourself (eg, `if yes_count == 0 and no_count == 0 …` or `if not pageIndex …` then throw an error).

openstates/md/bills.py
mileswwatkins: I think it's cleaner to explicitly raise the error, rather than relying on a data quality check so far down the import pipeline (that, who knows, could be removed at some point).

openstates/md/bills.py
mileswwatkins: Good call :)

openstates/md/bills.py
In-vincible: Actually I am relying here on the validator, which works in ```--scrape``` time, but after hearing you, I think checking it here only is the best idea.

openstates/md/bills.py
In-vincible: +1

openstates/md/bills.py
In-vincible: Totally

In-vincible:
@mileswwatkins I have made all the changes I think.

openstates/md/bills.py
```diff
@@ -189,44 +188,43 @@ def parse_vote_pdf(self, vote_url, bill):
...
             if 'Yeas' in line and 'Nays' in line:
-                pageIndex = index
+                page_index = index
                 break
 
-        yes_count = -1
-        no_count = -1
-        other_count = 0
+        vote_counts = 5*[None]
+        voteTypes = ['yes', 'no', 'not voting', 'excused', 'absent']
```
mileswwatkins: Similar comment on the camelCase vs snake_case variable naming.

openstates/md/bills.py
```diff
@@ -239,54 +237,41 @@ def parse_vote_pdf(self, vote_url, bill):
...
             if not currentLine or 'Voting Yea' in currentLine:
-                pageIndex += 1
+                page_index += 1
                 continue
 
             if any(showStopper in currentLine for showStopper in showStoppers):
-                pageIndex += 1
-                voteIndex = (voteIndex + 1) if (voteIndex < 2) else 2
+                page_index += 1
+                voteIndex = (voteIndex + 1)  # if (voteIndex < 4) else 4
```
mileswwatkins: Remove the in-line comment.

openstates/md/bills.py
mileswwatkins: Also, this `vote_counts` construction is a little non-standard, but it's valid syntax. And I see what you're going for.

In-vincible:
@mileswwatkins done last few changes.

mileswwatkins:
👍

Issue #2189, opened 2018-03-04
mi: Contact link not Working for one or more senator 


In-vincible:
Main link is : http://www.senate.michigan.gov/senatorinfo_list.html
In this contact link of johnson is broken. we are only loosing email because of this. But we may want to look into this, if we want other source or do something about it, though looks low priority to me. The PR referencing this issue is not solving this issue.

mileswwatkins:
@In-vincible, is Johnson supposed to be there, and is not present (eg, they're a new member to the legislature, and their page hasn't been created yet)? Or is Johnson an outgoing member, and they _shouldn't_ be present?

In-vincible:
He resigned 3 days ago, but he is still present in the main list, but his contact link is broken.

mileswwatkins:
Okay, then this'll go away soon, and he's no longer active/significant anyway. Going to close this issue, but good thinking regarding opening it up!

PR #2190, opened 2018-03-04
Email extraction omitted for broken contact links

In-vincible:
This solves #2187.
This PR, skips the email data for broken links, and lets the scraper continue.
I have created #2189, for the broken link, if we want to get the info in any case.

### DIFF
--- a/openstates/mi/people.py
+++ b/openstates/mi/people.py
@@ -1,4 +1,5 @@
 import re
+import requests
 
 import lxml.html
 import scrapelib
@@ -15,6 +16,7 @@ class MIPersonScraper(Scraper):
         elif chamber == 'lower':
             yield from self.scrape_lower(chamber)
         else:
+            requests.exceptions.TooManyRedirects
             yield from self.scrape_upper(chamber)
             yield from self.scrape_lower(chamber)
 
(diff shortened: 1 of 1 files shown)


openstates/mi/people.py
```diff
@@ -15,6 +16,7 @@ def scrape(self, chamber=None, session=None):
         elif chamber == 'lower':
             yield from self.scrape_lower(chamber)
         else:
+            requests.exceptions.TooManyRedirects
```
mileswwatkins: What does this statement do?

openstates/mi/people.py
```diff
@@ -136,25 +138,27 @@ def scrape_upper(self, chamber):
...
+                else:
+                    # not using the most common template, but maybe they
+                    # dropped their email on the page somewhere
+                    links = contact_doc.xpath('//a') or []
+                    text_email = [a for a in links
+                                  if 'mailto:' in (a.get('href') or '')]
+                    if text_email:
+                        email = text_email[0].text
+            except requests.exceptions.TooManyRedirects:
+                print("Conatct Link Not Working for %s" % name)
```
mileswwatkins: Instead of `print("message")`, can you use `self.warning("message")`? This uses more formalized Python `logging`.

mileswwatkins: Also, small typo in the error message itself.

In-vincible:
@mileswwatkins Really Sorry I think this commit, was rather carelessly done, was traveling, I will put forth a PR with the changes.

mileswwatkins:
Oh man, I made a mistake too, must've hit "Merge" before I meant to, haha. Thanks for the follow-up PR!

PR #2194, opened 2018-03-05
WI senate vote adding function re-written

### LINKED ISSUE #1909: WI: All 2017 individual roll call votes are missing for Senate, some Assembly too
The Senate currently has no individual roll call votes recorded in 2017.

In-vincible:
resolves #1909
Will check for assembly votes too.

### DIFF
--- a/openstates/wi/bills.py
+++ b/openstates/wi/bills.py
@@ -331,21 +331,27 @@ class WIBillScraper(Scraper):
         doc = lxml.html.fromstring(html)
         trs = doc.xpath('//table[@class="senate"]/tbody/tr')
 
-        voteTypes = ['yes', 'no', 'other']
-        counts = {}  # Vote counts for yes, no, other
+        vote_types = ['yes', 'no', 'other']
+        vote_counts = {}  # Vote counts for yes, no, other
+        name_counts = {}
 
         for index in range(0, 5, 2):
 
-            voteType = voteTypes[int(index/2)]
+            vote_type = vote_types[int(index/2)]
             names = trs[index].xpath('.//table//td/text()')
-            voteCount = int(trs[index].xpath('./td/text()')[0].split('-')[1])
-            counts[voteType] = voteCount
+            vote_count = int(trs[index].xpath('./td/text()')[0].split('-')[1])
+            vote_counts[vote_type] = vote_count
+            name_counts[vote_type] = len(names)
 
             for name in names:
-                vote.vote(voteType, name.strip())
+                vote.vote(vote_type, name.strip())
 
-        for voteType in voteTypes:
-            vote.set_count(voteType, counts[voteType])
+        for vote_type in vote_types:
+            vote.set_count(vote_type, vote_counts[vote_type])
+
+        if name_counts != vote_counts:
+            self.error("Yes votes and number of Names doesn't match")
+            raise ValueError("Vote Count and number of Names don't match")
 
     def add_house_votes(self, vote, url):
         try:
@@ -378,6 +384,8 @@ class WIBillScraper(Scraper):
                     vote.vote('other', name)
 
         if yes_names_count != int(ayes_nays[0][0]):
-            self.warning("Yes votes and number of Names doesn't match")
+            self.error("Yes votes and number of Names doesn't match")
+            raise ValueError("Vote Count and number of Names don't match")
         if no_names_count != int(ayes_nays[0][1]):
-            self.warning("No votes and number of Names doesn't match")
+            self.error("No votes and number of Names doesn't match")
+            raise ValueError("Vote Count and number of Names don't match")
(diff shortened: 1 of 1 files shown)


In-vincible:
close #1909

mileswwatkins:
It's helpful if your PR title starts with the state abbreviation, for reference; eg `WI: …`

mileswwatkins:
Also, is this ready for review?

In-vincible:
Yes

openstates/wi/bills.py
```diff
@@ -375,12 +368,24 @@ def add_house_votes(self, vote, url):
...
             name = td.text_content()
             for vote_td in td.xpath('./preceding-sibling::td'):
                 if vote_td.text_content() == 'Y':
                     vote.vote('yes', name)
+                    yes_names_count += 1
                 elif vote_td.text_content() == 'N':
                     vote.vote('no', name)
+                    no_names_count += 1
                 elif vote_td.text_content() == 'NV':
                     vote.vote('other', name)
```
mileswwatkins: It looks like this category is actually `not voting`, not `other`, right?

openstates/wi/bills.py
```diff
@@ -329,36 +329,29 @@ def add_senate_votes(self, vote, url):
...
+            name_counts[vote_type] = len(names)
+
+            for name in names:
+                vote.vote(vote_type, name.strip())
+
+        for vote_type in vote_types:
+            vote.set_count(vote_type, vote_counts[vote_type])
+
+        if name_counts != vote_counts:
+            self.error("Yes votes and number of Names doesn't match")
```
mileswwatkins: Wrong error message?

openstates/wi/bills.py
```diff
@@ -329,36 +329,29 @@ def add_senate_votes(self, vote, url):
...
-                        vote.vote('other', name.strip())
-                        other_count += 1
-            elif text.strip():
-                raise ValueError('unexpected block in vote')
-
-        if yes_count or no_count or other_count:
-            vote.set_count('yes', yes_count)
-            vote.set_count('no', no_count)
-            vote.set_count('other', other_count)
+        for index in range(0, 5, 2):
```
mileswwatkins: This seems like weird construction, both non-standard Python and hard to read (for future maintainers). Could you do something like this instead?

- XPath for `<td>` that has text of `(AYES|NAYS|NOT VOTING) - \d+`
- get the names from the `<table>` in the [following-sibling](https://www.w3schools.com/xml/xpath_axes.asp) `<td>`

openstates/wi/bills.py
```diff
@@ -329,36 +329,29 @@ def add_senate_votes(self, vote, url):
             return
 
         doc = lxml.html.fromstring(html)
+        trs = doc.xpath('//table[@class="senate"]/tbody/tr')
 
-        yes_count = no_count = other_count = 0
+        vote_types = ['yes', 'no', 'other']
```
mileswwatkins: The third type is `not voting`, right?

openstates/wi/bills.py
In-vincible: +1

openstates/wi/bills.py
In-vincible: 100% agreed, doing it.

openstates/wi/bills.py
In-vincible: I am not really sure, absent, excused were not given, are they are taking all others in ```not voting``` only, that's why I thought maybe other would be right to use here. But yeah ```not voting``` is mentioned here. I should with ```not voting``` then?

mileswwatkins: Yes, if the state labels them as "Not Voting" then we should use that Open Civic Data category.

In-vincible:
@mileswwatkins ready for next round of review.

openstates/wi/bills.py
```diff
@@ -350,7 +353,7 @@ def add_senate_votes(self, vote, url):
             vote.set_count(vote_type, vote_counts[vote_type])
 
         if name_counts != vote_counts:
-            self.error("Yes votes and number of Names doesn't match")
+            self.error("Names not extracted correctly")
             raise ValueError("Vote Count and number of Names don't match")
```
mileswwatkins: Are `self.error` and `raise …` redundant?

mileswwatkins: Looks like there are a couple cases further down in the code, too.

In-vincible: I will delete self.error then!

In-vincible: I personally think, we should throw error, as we will get notified for that, and hence we can act on it.

In-vincible:
@mileswwatkins I have removed ```self.error``` line, thought should stick with raise error, so that we get notified in case of descrepancies.

PR #2205, opened 2018-03-16
NH: Fix lower-house role title

mileswwatkins:
@jamesturk, no rush on this. Just curious as to whether there're any special steps we need to take when making changes/corrections to post/role metadata like this.


jamesturk:
should be fine 👍

PR #2230, opened 2018-03-30
PA: ignore sessions from the '60's

estaub:
No issue yet; Bobsled will file one tonight if not approved.

Looks like they renamed some sessions; I left the old names ignored also, in case they flip back.


showerst:
I pushed a fix for this without noticing you'd beat me to it Ed. Next time i'll check the PR's first!

estaub:
@showerst No problem!

PR #2280, opened 2018-05-06
MA: Updated committee rule

EricPaulson:
It seems appropriate to add 'referral-committee' to the any rule that has to do with referring to committees. I've also noticed a large chunk of actions that go unclassified. Should new categories be added for larger coverage? Some example are:

Placed on file
Senate concurred
Accompanied a new draft (and several others starting with 'Accompanied')
Hearing scheduled
Reprinted as amended
Committee of conference appointed
Emergency preamble adopted

That's the majority of them. What would you think about either assigning categories to these, and creating new categories where needed? I'd be happy to help of you want to expand on this.

### DIFF
--- a/openstates/ma/actions.py
+++ b/openstates/ma/actions.py
@@ -50,7 +50,7 @@ _categorizer_rules = (
          ['amendment-passage']),
     Rule(['Amendment #\\S+ \\((?P<legislator>.+?)\\) Rejected'],
          ['amendment-failure']),
-    Rule(['referred to (?P<committees>.+)'], []),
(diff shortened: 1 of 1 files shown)


showerst:
Yes we'll gladly take action categorization updates, especially for those that map to existing actions.

If it's easier for you to do many in one PR rather than pr-per-type, that's fine.

Adding a new category type is a bit of a process, I have an open issue on https://github.com/opencivicdata/python-opencivicdata/issues/116 -- feel free to propose new ones there and we'll talk it out w/ the opencivicdata folks.

EricPaulson:
Wonderful! I'll get involved as much as I can to get everything up to date.

PR #2288, opened 2018-05-08
IA: Fix motion regexes when trailing quote is missing (but question-mark is present)

estaub:
... and vice versa


estaub:
Fixes #2279.

PR #2296, opened 2018-05-10
NH: Include supposedly "former" members in legislature.

estaub:
There's no issue posted for this, but see comment https://github.com/openstates/openstates/issues/2220#issuecomment-386479706 .  

This problem is not visible in the openstates db because of another problem that I can't fix; I can see it because I've started scraping into my own DB, just so I can have access to NH data, and also perhaps because it's a new DB (so the "former" legislators aren't already in it as legacy data).

This will need to be undone when/if they start validly marking legislators as "Former".


estaub:
They still haven't fixed this; see http://www.gencourt.state.nh.us/downloads/Members.txt

PR #2313, opened 2018-05-18
OR - add special session

estaub:
The scrape will continue to fail until results from the May 21st session are posted.  It's so close I didn't bother with a temporary ignore.


showerst:
I'll keep an eye on this and merge it when the data goes live.

PR #2333, opened 2018-05-23
Co vote scraper revamped.

### LINKED ISSUE #2318: CO: No votes scraped in 2018.
It looks scrapable to me, but I didn't delve deeply.

In-vincible:
resolves #2318 

Motion not available for the http://leg.colorado.gov/content/sb18-033vote563ce6.

### DIFF
--- a/openstates/co/bills.py
+++ b/openstates/co/bills.py
@@ -291,40 +291,32 @@ class COBillScraper(Scraper, LXMLMixin):
                                      amendment_letter)
 
     def scrape_votes(self, bill, page):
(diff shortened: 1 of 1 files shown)


estaub:
@In-vincible On a "third reading" vote (see action on [bill page](https://leg.colorado.gov/bills/sb18-033)), I suspect (not sure!!!) you can infer no motion as meaning "PASS".

In-vincible:
@estaub for this particular case it seems "PASS" is a nice fit, but what if motion goes empty in any another case in future, We need to know when the motion is not scraped.

In-vincible:
@jamesturk I think this is ready for review.

PR #2342, opened 2018-05-25
MO - add special session to term

estaub:
This is a guess at what's wrong with MO; I'm not set up to test billy import.  

If it makes no sense that this should have caused the problem, I'd be very suspicious that the github master wasn't redeployed into the bobcat run last night.

This may address issue #2341


estaub:
Redeployment probably happened last night, based on  the fact that MN ran green.  See #2336.

showerst:
@estaub this needed on other change (the session key in the session_details object has to match the string in the list of terms) , so i just did it on my own branch.  Thanks for the PR though, you were on the right track. It was my bug, not having a test process for this makes it easy to bug. Thankfully the billy imports be going away pretty soon.

PR #2352, opened 2018-06-01
NY assembly votes

### LINKED ISSUE #1960: NY: No 2017 bills contain any Assembly votes

In-vincible:
resolves #1960,
Since #2185 worked on this issue earlier, I worked on top of it.
I was facing error with validation of texts.
I have commented the validation part in ny/utils.py, I have added separate commit for that I am not sure how to handle that part, please review this part and suggest what should be done.

### DIFF
--- a/openstates/ny/utils.py
+++ b/openstates/ny/utils.py
@@ -45,7 +45,7 @@ class UrlData(object):
     @CachedAttr
     def text(self):
         text = self.scraper.get(self.url).text
-        self.urls_object.validate(self.name, self.url, text)
+        # self.urls_object.validate(self.name, self.url, text)
         return text
 
     @CachedAttr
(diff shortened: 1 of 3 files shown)


openstates/ny/utils.py
```diff
@@ -45,7 +45,7 @@ def __repr__(self):
     @CachedAttr
     def text(self):
         text = self.scraper.get(self.url).text
-        self.urls_object.validate(self.name, self.url, text)
```
jamesturk: what error was this raising?

In-vincible: File "/media/bikram/0bb967d3-86e9-4eee-934c-c0fc3f76980d/home/bikram/openstates/openstates/ny/utils.py", line 148, in validate
    for validator in self._validators[name]:
AttributeError: 'Urls' object has no attribute '_validators'

jamesturk:
OK, feel free to leave that commented out if it works without it, we need
to revisit that ny/utils stuff at some point but I'd rather get this
working first

In-vincible:
@jamesturk its working after commenting it.

PR #2356, opened 2018-06-03
Update alpine dockerfile.

jmcarp:
Looks like the alpine docker image got out of sync with the default image. This patch bumps the base alpine version and adds a few missing dependencies. After the update, I'm able to use the alpine image to run scrapes again.

Also, it looks like the alpine image we're publishing on dockerhub is actually the ubuntu image.

Also, wdyt about moving towards using alpine by default? The ubuntu image is big.

### DIFF
--- a/Dockerfile-alpine
+++ b/Dockerfile-alpine
@@ -1,4 +1,4 @@
-FROM        alpine:3.4
+FROM        alpine:3.7
 MAINTAINER  James Turk <james@openstates.org>
 
 ENV PYTHONIOENCODING 'utf-8'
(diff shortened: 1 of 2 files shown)


jamesturk:
hey- thanks for taking a look at this, I think some things weren't working quite right yet but I hadn't made time to figure out what.  I'll go ahead & merge this for now & try to figure out why the dockerhub build isn't right- I'm all for making this the actual Dockerfile once we've verified all scrapes work w/ it (which we can do by testing a few at a time in prod)

PR #2368, opened 2018-06-12
WY: Fix bill imports.

jmcarp:
* Add executive organization
* Comment out unsupported bill version extras

Verified that scraping and importing WY bills works locally with this patch.

### DIFF
--- a/openstates/wy/__init__.py
+++ b/openstates/wy/__init__.py
@@ -110,6 +110,7 @@ class Wyoming(Jurisdiction):
                 label=str(n), role=lower_title,
                 division_id='{}/sldl:{}'.format(self.division_id, n))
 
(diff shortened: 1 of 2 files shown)


jmcarp:
We could add an extras field to the bill versions model to get the commented-out code to work. Should we?

jamesturk:
In this case I'm trying to think if there's a better way to represent that data in a more discoverable way but it does seem like maybe extras on versions would make sense (with the caveat that discoverability is still not great, so I'm not sure who will wind up using this data)  I'll go ahead and merge this as-is, but if you want to create a ticket (and/or PR if you're up for it) to add extras to versions on python-opencivicdata I think it makes sense to allow

PR #2369, opened 2018-06-13
VT: Extract site id from sessions.

jmcarp:
The jurisdiction model doesn't support extras on sessions, so we can't
store site ids there.


jmcarp:
cc @showerst

PR #2393, opened 2018-06-24
Add "nonpartisan" party.

jmcarp:
Example at
https://www.leg.state.nv.us/App/Legislator/A/Senate/Current/8.

Also, there's probably a better place to put parties than a shell script in this repo. Should this live in `python-opencivicdata` or another repo? I'm also pretty sure I'm missing more parties, since I haven't been adding them systematically. Do you have a more authoritative list @jamesturk ?

### DIFF
--- a/dbinit.sh
+++ b/dbinit.sh
@@ -10,6 +10,7 @@ Progressive
 Progressive/Democratic
 Democratic/Progressive
 Democratic-Farmer-Labor
+Nonpartisan
 "
 
 for party in ${parties}; do


showerst:
It would at least be nice if they were in a JSON file someplace.

Here's a list of the unique parties in my DB, which should be either from openstates or unitedstates/congress-legislators for some of the historical ones --

```
Democrat
Democratic
Democratic-Farmer-Labor
Democratic-Republican
Democratic/Progressive
Federalist
Independent
no party
No Party Specified
Non Affiliated
Nonpartisan
Other
Partido Independentista PuertorriqueÃ±o
Partido Nuevo Progresista
Partido Popular DemocrÃ¡tico
Progressive
Progressive/Democratic
Republican
Republican/Democratic
Unaffiliated
Unknown
Whig
Working Families
```

jamesturk:
Yeah this is a fair point, python-opencivicdata is supposed to remain fairly country-neutral which is why they aren't there, but like some of the other US-specific values we should find a place for them, maybe the best thing to do would be to add some country-specific settings.

Merging this for now- but feel free to open an issue there if you get to it before I do (which is likely at this point :/)

JasonWeakley:
Is the question where to find a list of political parties? 
If so Ballotpedia has a comprehensive list of national parties, with some local ones:
https://ballotpedia.org/List_of_political_parties_in_the_United_States

I also found and verified some of the states are accurate here:
https://votesmart.org/political-parties/ (includes national and all state parties - there are many parties at state level that only exist in one state)

PR #2403, opened 2018-07-01
NV: Fix session scraping and add 2019 session.

jmcarp:
Or should we disable the 2019 session until it starts?


showerst:
Oops looks like we overlapped here, fine to use yours.

showerst:
Does it return anything (those BDR prefiles) or nothing? 

If it comes back w/ data i say we keep it, otherwise let's disable.

PR #2407, opened 2018-07-03
WY: Restore bill version extras.

jmcarp:
Now that python-opencivicdata supports extras on bill versions, we can
restore extra properties on WY bill versions.


jamesturk:
note to myself to merge this once I've created a new py-ocd release

jmcarp:
Looks like we might already already be using the latest py-ocd, based on failures in bobsled.

PR #2415, opened 2018-07-06
Fix SD email address scraping

samtregar:
SD has removed email addresses from detail pages, fetch them from the Contact Legislators page instead.  Fixes issue 2414.


samtregar:
Thanks!

PR #2418, opened 2018-07-10
GA: Separate legislator names and districts in votes.

### LINKED ISSUE #1812: GA votes need name matching

jmcarp:
[Fixes #1812]

### DIFF
--- a/openstates/ga/bills.py
+++ b/openstates/ga/bills.py
@@ -1,3 +1,4 @@
+import re
 from collections import defaultdict
 
 from pupa.scrape import Scraper, Bill, VoteEvent
@@ -25,6 +26,8 @@ from .util import get_client, get_url, backoff, SESSION_SITE_IDS
 
 member_cache = {}
 SOURCE_URL = 'http://www.legis.ga.gov/Legislation/en-US/display/{session}/{bid}'
+
+vote_name_pattern = re.compile(r'(.*), (\d+(?:ST|ND|RD|TH))', re.IGNORECASE)
 
 
 class GABillScraper(Scraper):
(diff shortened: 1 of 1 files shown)


jmcarp:
Just noticed a similar fix in #1864. I see that @jamesturk was concerned about losing information about district, which I'm including here in the `note`; does that make sense? The Georgia API also includes legislator IDs along with votes, so we could expose those as well, but I don't see that pupa would make use of either the district number or the unique ID.

jmcarp:
Would it be useful to add an optional district to the `vote` method for resolution by pupa?

jamesturk:
i'm sort of torn on the right philosophy on these, on one hand- we could simply accept the strings as-is and then use the entity matching code to alias Jones-21st and Jones-4th to the proper Joneses.  The other approach is as you suggest, adding more custom code to support these cases.

The first is a more general solution, and better handles cases we've seen where they only provide the district disambiguator on one of the two (fun times), but I'm open to argument in favor of the latter over it

jmcarp:
Could you point me to the entity matching code? I poked around pupa a bit, but I'm not sure where that logic lives, or what it does.

jamesturk:
so the story in pupa is simply that if something is in other_names it can be resolved to the same entity- it is done here:

https://github.com/opencivicdata/pupa/blob/master/pupa/importers/base.py#L130
and 
https://github.com/opencivicdata/pupa/blob/master/pupa/importers/people.py#L37

billy had much more complex stuff going on, allowing manual resolution- which we're hoping to avoid as it became untenable, especially w/o dedicated intern resources

PR #2440, opened 2018-07-23
SC: remove http protocol downgrades

showerst:
This needs a full run-through test before merge, but i'm about to hop on a plane.


showerst:
Leaving in the HTTP downgrade functions (unused) for a bit, to save time if the server bug crops back up. I'll remove them later if things go smoothly.

PR #2474, opened 2018-08-16
WV: Fix typo in special session identifier.

jmcarp:


showerst:
The previous specials in this state used <digit>S -- Should we change it out to standardize, or try to keep it the same in each state?

jmcarp:
Good catch, I switched back to `2S` and made the identifiers consistent for WV.

PR #2478, opened 2018-08-16
Tim/2019/ky prefiles

showerst:
KY is a state that has its prefiles up on seperate pages and doesn't assign them true bill numbers yet.

Re-reading https://github.com/openstates/openstates/pull/1146 and https://github.com/openstates/openstates/issues/1205 -- I put prefiles behind a pupa argument, so that you have to pull them explicitly. I also made it so 2019 isn't the default session, since otherwise running scrape fl would throw an error.

My proposed syntax:
```
pupa update fl bills --scrape session=2019RS prefile=True
```

Any objections? I won't merge this one until ok'd.

### DIFF
--- a/openstates/ky/bills.py
+++ b/openstates/ky/bills.py
@@ -99,11 +99,11 @@ class KYBillScraper(Scraper, LXMLMixin):
         bill_url = 'http://www.lrc.ky.gov/record/{}/prefiled/prefiled_bills.htm'.format(
             abbr)
         if 'upper' == chamber:
-            bill_url = 'http://www.lrc.ky.gov/record/{}/prefiled/prefiled_sponsor_senate.htm'.format(
-                abbr)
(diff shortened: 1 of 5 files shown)


openstates/ky/bills.py
```diff
@@ -40,7 +40,7 @@ def scrape_subjects(self, session):
                 for bill in sdoc.xpath('//div[@id="bul"]/a/text()'):
                     self._subjects[bill.replace(' ', '')].append(subject)
 
-    def scrape(self, session=None, chamber=None):
+    def scrape(self, session=None, chamber=None, prefile=None):
```
jmcarp: If `prefile` is a flag, I would expect the default to be a boolean, probably `False`.

showerst: Good catch, i'll make this change later.

PR #2487, opened 2018-08-24
KS: Fail on bill scrape errors.

jmcarp:
I was going to fix the `KeyError` in the `except` block here, but taking a closer look, I'm not sure why we're ignoring scrape errors in the first place. I'm running the scraper now to see if this introduces any new problems.

### DIFF
--- a/openstates/ks/bills.py
+++ b/openstates/ks/bills.py
@@ -2,7 +2,6 @@ import re
 import json
 import datetime
 
-import scrapelib
 import lxml.html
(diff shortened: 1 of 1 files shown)


showerst:
Not 100% but I suspect this was to get around KS' aggressive rate limiting and constant server issues. I think i'm with you though, i'd rather bomb out than have it chug along without the versions.

jmcarp:
FWIW, I ran this locally and it finished without crashing. YMMV 😆

PR #2488, opened 2018-08-24
Use alpine docker image.

jmcarp:
I haven't run through every state with this image, so it's likely to break something, but I think we might as well switch over to the smaller alpine image and then clean things up. WDYT @jamesturk @showerst ?


showerst:
I"m heading out of town until after labor day, but i'll try running it through all 51 too when I get back. I should be around a bit each day to merge PRs though so feel free to keep requesting reviews, i'm more likely to see it that way.

PR #2511, opened 2018-09-11
Merge pull request #2510 from openstates/tim/2018/mo-special2

jmcarp:
Might be useful to configure the repo to prevent merges until the build passes. I don't have permission to do that--would you be open to either setting that up or giving me access @showerst @jamesturk ?


showerst:
(I don't seem to have access or i would) -- This must have happened when the integration was down, otherwise i'd have caught it. Any idea how that works if the remote service isn't running? 

My bad for not noticing and manually flake8-ing, thanks for the fix.

PR #2515, opened 2018-09-12
MO: Add billy metadata for 2018S2.

jmcarp:
Recommend testing scrapers with postgres locally 😄


showerst:
I go on vacation for ONE WEEK and forget how to program. Oof.

PR #2516, opened 2018-09-12
Pr add parties

jmcarp:
It looks like the at-large divisions we added for PR in #2507 don't exist in https://github.com/opencivicdata/ocd-division-ids. This patch replaces them with divisions that exist in our database. I also added some PR parties to `dbinit.sh` so that I can run the import locally.

@showerst @csnardi

### DIFF
--- a/dbinit.sh
+++ b/dbinit.sh
@@ -11,8 +11,12 @@ Progressive/Democratic
 Democratic/Progressive
 Democratic-Farmer-Labor
 Nonpartisan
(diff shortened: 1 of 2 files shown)


csnardi:
Hm, I assumed https://github.com/opencivicdata/ocd-division-ids/blob/master/identifiers/country-us/README.md#cd applied to all At-Large districts, but since I do see the At-Large division notices in that repo for PR, I guess it must not.

PR #2522, opened 2018-09-15
PR: Convert senator names into title case

csnardi:
Some senators have their names in all-caps. To avoid this, normalize names into title case to make them easier to read.


showerst:
We don't usually modify the names the state provides for purely stylistic reasons, although in this case since the state provides a mix of all caps and title case i'm inclined to merge.

@jamesturk @mileswwatkins @jmcarp any objections?

PR #2559, opened 2018-10-27
Use rawstrings for regexes with escape sequences.

jmcarp:
Looks like flake8 3.6 started checking for regexes with escape sequences.


showerst:
Big thanks for cleaning this up @jmcarp!

PR #2664, opened 2018-11-10
MT: Handle markup changes.

jmcarp:
I verified that this runs and produces reasonable-looking bills and votes, but it wouldn't hurt to have somebody more familiar with MT spot-check a few records. I'm also not sure what the `_versions_dict` and `add_other_versions` methods do, so I commented out calls to them for now.

### DIFF
--- a/openstates/mt/__init__.py
+++ b/openstates/mt/__init__.py
@@ -1,4 +1,6 @@
 from pupa.scrape import Jurisdiction, Organization
+
+from openstates.utils.lxmlize import url_xpath
 # from .people import MTPersonScraper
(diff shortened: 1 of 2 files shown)


jamesturk:
I think the identifier changes would create duplicates if those sessions were scraped again, need to look over this in a bit more detail

showerst:
@jmcarp I've pull requested your branch w/ an update that keeps the identifiers intact, and adds 2019

showerst:
I'm running this now for 2019 an can confirm the data looks good / matches the state, so i'm going to merge. Thanks @jmcarp

PR #2671, opened 2018-11-12
GA: Add 2018 special session.

jmcarp:
This doesn't actually work so far--looks like the GA API doesn't know about the new session yet. I'll check on this in a few days once the special starts.


showerst:
This data is in their system now and the scrape completes.

Issue #2708, opened 2018-12-10
start pulling seat/chamber data from openstates/people?


jamesturk:
the __init__.py have hard-coded districts, but we keep a better copy of that data in openstates/people

we should either have the __init__.py read from those or figure out a better way to get that data copied over & kept in sync

estaub:
>> the init.py have hard-coded districts, but we keep a better copy of that data in openstates/people

If other folks are looking too, it's in https://github.com/openstates/people/blob/master/settings.yml

PR #2710, opened 2018-12-12
migrate post information to openstates/people

jamesturk:
#2708


showerst:
Thanks James, this must've been super tedious.

PR #2750, opened 2019-01-06
CA: Switch back to root for MySQL user

csnardi:
It seems as if the username is still root, attempt to switch back to fix connection issues.


showerst:
@jamesturk just pinging you to point out that i'm merging this.

PR #2751, opened 2019-01-06
Updated docs regarding oregon API auth

nickoneill:
I emailed them for an API key but they said it does not require auth. Ran the scraper without it and it works great, but data has not been updated for 2019 yet.


nickoneill:
For sure, will update this PR

PR #2757, opened 2019-01-08
OH: Add 2019 session and adapt scraper for new website

showerst:


jamesturk:
hey- just noticed on this one, but we aren't using billy_metadata anymore (and hopefully never will again) unless you all are no need to update those anymore (figure I'll pull them soon if they aren't needed)

showerst:
Oh awesome, didn't know you'd migrated off of those!

PR #2777, opened 2019-01-12
Add pupa_settings step to instructions for import to database

jessemortenson:
While trying to figure out how to run openstates code, I ran into a barrier: [data from state scraper not being imported into the database](https://github.com/openstates/openstates/pull/2745). I finally figured out that pupa_settings.py was essentially telling pupa not to import these things. I'm not sure why those are default settings in the repo - maybe they should be changed? - but at least I think they should be mentioned in the set of instructions that lead up to telling users that the scraper can save data to the database.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,6 +40,11 @@ Then run database migrations and import jurisdictions::
 
     docker-compose run --rm dbinit
 
+You also need to enable the import of certain entities into the database. Change pupa_settings.py to::
+
+    ENABLE_PEOPLE_AND_ORGS = True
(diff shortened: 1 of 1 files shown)


jamesturk:
These are not really recommended for import for most people anymore, I'll be taking a pass at the documentation soon to address this & leave this open as a reminder

jamesturk:
docs are now updated with new instructions on using the database
file:///Users/james/code/os/documentation/_build/html/contributing/running-the-site.html#getting-a-working-database

PR #2852, opened 2019-02-11
MT: rewrite of people scraper

rsimoes:


rsimoes:
A site redesigned required a full rewrite. There is a CSV available, but there appears to only be one for the most recent session.

PR #2866, opened 2019-02-14
[IL] Temporary fix for committee resolution

estaub:
This hacks the source url for committees, used in pseudo-id resolution, to set the session back from 101 to 100, to match the urls in the openstates/people committees.  This is intended to provide time to consider support for committees when not under the gun.

Issue: https://github.com/openstates/openstates/issues/2821 .


estaub:
The build failed on an unused import in kentucky.

Issue #2868, opened 2019-02-14
Switch Docker Compose Postgres port from 5432 to 5433?


mileswwatkins:
cc @jmcarp @jamesturk 

Devs may have a local Postgres server already running at `5432`, the default Postgres port. This makes it a pain to connect to the Docker Compose `postgres` service, which tries to be available at that same `localhost:5432` port.

The solution I've used in the past, that I like, is to map the Docker Compose Postgres to another port, like `5433`. Thoughts on doing this here, or would it cause headaches that I'm not considering?

jamesturk:
I don't have any objection, I think as long as it got updated everywhere it'd be fine (I don't use the docker compose DB myself so I'm less likely to know if there'd be issues)

PR #2870, opened 2019-02-15
Switch Postgres port from 5432 to 5433 to improve usability from local psql; cc #2868

mileswwatkins:
@jmcarp, let me know if this looks good to you; I've used this solution on other projects and it solves some headaches


jamesturk:
external port is now 5405

PR #2871, opened 2019-02-15
CA: Upgrade MySQL container from version 5 to 8

mileswwatkins:
Running the CA download and import using MySQL 8 instead of 5 is working smoothly. I'm not very familiar with the MySQL community, though; is using 5.7 still standard/best-practice?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,7 +28,7 @@ services:
     depends_on:
     - postgres
   mysql:
-    image: mysql:5.7
+    image: mysql:8
     command: mysqld --max_allowed_packet=512M
     ports:
     - "3306:3306"


showerst:
MySQL 8 has been stable (GA) for about a year now, so should be fine, especially for this use case where we're just spinning up and then killing them. 

There are some indexing/behavior changes but they're small so if the import worked, i think we should be ok. I'm gonna merge this.

mileswwatkins:
Go for it. My only adjacent thought is that some of the shell commands used in the California task definition may need to be rejiggered – but they're _already_ not working, so that would be necessary anyway.

showerst:
Mark my words for next session i'm just gonna take a weekend and rewrite this stupid scraper to use the flatfiles they include in the download.

PR #2872, opened 2019-02-15
Switch Docker Compose to MariaDB container, to mimic Alpine server used in production

mileswwatkins:
Ugh, actually @showerst, the environment we should've been reproducing in Docker Compose is MariaDB. Alpine apparently switched from MySQL to MariaDB in their 3.2 release, and our Dockerfile uses MariaDB packages when installing (and then running in production).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,8 +28,8 @@ services:
     depends_on:
     - postgres
   mysql:
-    image: mysql:8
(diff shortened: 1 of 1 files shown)


mileswwatkins:
Again, tested the download and ingest locally myself. Merge if my thinking here makes sense.

PR #2906, opened 2019-03-05
CA: fix duplicate bills

estaub:
This addresses the ongoing corruptions reported in https://github.com/openstates/openstates/issues/2880, but someone will need to fix the database. 

(It is also a poster child for why functions that run on for pages and pages are a bad thing!)

### DIFF
--- a/openstates/ca/bills.py
+++ b/openstates/ca/bills.py
@@ -577,10 +577,10 @@ class CABillScraper(Scraper):
                 full_loc = vote.location.description
                 first_part = full_loc.split(' ')[0].lower()
                 if first_part in ['asm', 'assembly']:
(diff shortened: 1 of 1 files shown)


jamesturk:
thanks for this Ed!  I'll take a swing at the DB after this runs & I can verify they aren't being recreated

estaub:
@koryp123 This query works for me, can you share what's failing for you?

```graphql
{
  bill(jurisdiction: "California", session: "20192020", identifier: "SB 699") {
    id
    identifier
    title
  }
}
```

estaub:
@koryp123 See https://github.com/openstates/openstates/issues/2880#issuecomment-470272376

PR #2958, opened 2019-04-29
Updated regex Line 30

cultureclap:
Two forward slashes are needed to escape - maybe a Python 3 thing?


mileswwatkins:
Thanks, Canin! What issues is the existing code causing on your end? I believe that the raw string formatting (`r'…'`) allows single backslash characters to work fine on their own, which is why it's so often used with `re` syntax. But let me know if we're missing something here.

```
>>> print('\s{1,4}')
\s{1,4}
>>> print(r'\s{1,4}')
\s{1,4}

>>> print('\\s{1,4}')
\s{1,4}
>>> print(r'\\s{1,4}')
\\s{1,4}
```

PR #2995, opened 2019-06-23
HI: Issue 2957: fix duplicate vote pseudo-ids when vote reconsidered.

estaub:
If a line in a bill's action log includes "reconsider" after a vote, this patch prepends "Reconsider: " to the next vote's name.


showerst:
Thanks Ed!

Issue #3029, opened 2019-10-21
MI: Missing bills in 2019-2020


jessemortenson:
State: MI - Michigan

We found a bunch of bills missing from OpenStates for this jurisdiction and session, including:

[HB 4852](https://www.legislature.mi.gov/(S(4cewuxqvris3xuspgngsdh5t))/mileg.aspx?page=getObject&objectName=2019-HB-4852)
HB 4868
HB 4869
HB 4870
[SB 452](https://www.legislature.mi.gov/(S(qfybxzgtuheeke4q2byj2hkx))/mileg.aspx?page=getObject&objectName=2019-SB-0452)

Investigated by attempting to run the bill scrape locally:

```
jesse@greenbookx:~/repo/civiceagle/openstates$ docker-compose -f hybrid-compose.yml run --rm scrape mi bills --scrape
loaded Open States pupa settings...
mi (scrape)
  bills: {}
20:06:49 INFO pupa: save jurisdiction Michigan as jurisdiction_ocd-jurisdiction-country:us-state:mi-government.json
20:06:49 INFO pupa: save organization Michigan Legislature as organization_50bec3d2-f43e-11e9-b509-0242ac120002.json
20:06:49 INFO pupa: save organization Senate as organization_50bec97c-f43e-11e9-b509-0242ac120002.json
20:06:49 INFO pupa: save organization House as organization_50beccce-f43e-11e9-b509-0242ac120002.json
20:06:49 INFO pupa: no session specified, using 2019-2020
20:06:49 INFO scrapelib: GET - http://legislature.mi.gov/doc.aspx?2019-SB-0001
20:06:50 INFO scrapelib: GET - http://www.legislature.mi.gov/documents/2019-2020/Journal/Senate/htm/2019-SJ-05-07-043.htm
20:06:51 WARNING pupa: "In The Chair" indicator not found, unable to extract vote
Traceback (most recent call last):
  File "/opt/openstates/venv-pupa//bin/pupa", line 11, in <module>
    load_entry_point('pupa', 'console_scripts', 'pupa')()
  File "/opt/openstates/venv-pupa/src/pupa/pupa/cli/__main__.py", line 68, in main
    subcommands[args.subcommand].handle(args, other)
  File "/opt/openstates/venv-pupa/src/pupa/pupa/cli/commands/update.py", line 278, in handle
    return self.do_handle(args, other, juris)
[...]

PR #3030, opened 2019-10-21
MI: avoid error when vote roll call data is not as expected

jessemortenson:
I investigated #3029 and found problems. Hopefully this is an easy fix.

MI scrape was halting at an error NoneType object is not subscriptable when parse_roll_call() returned None. Check for None before calling the vote yielding logic. Also fixed the guard in parse_roll_call() so that it can find the text its looking for even when there are spaces/newlines/span tags between the words (which was causing None to be returned).


jessemortenson:
Is there any way for me to see why the build is failing, to see if I can address it?

jamesturk:
Thanks, I haven't had a chance to look at this, but if you click through to
travis it shows the results:

https://travis-ci.com/openstates/openstates/builds/132893869

jessemortenson:
Thanks, wow, I had to look 3 times to find the link :man_facepalming:

PR #3031, opened 2019-10-23
MI: Handle HTTP error when fetching roll call document

jessemortenson:
404 on request for a roll call document (vote details) would cause whole scrape to fail

Should fix additional issue identified on #3029


jamesturk:
thanks!

PR #3033, opened 2019-10-30
WI: Add 2019 November Special Session

showerst:
NOTE: Don't merge until November 11, after testing. Just queuing this up since i'll be out of town.


showerst:
Looks like they didn't end up adding this to the site.

PR #3041, opened 2019-11-20
NY: Make text url load full billtext

showerst:
By default NY HTML urls weren't pulling the bill's text, you had to click a button. This changes the URLs to include the billtext.


showerst:
Head's up @jessemortenson since I recall you using New York regularly.

PR #3054, opened 2019-12-03
Update ak bills scraper

schneidy:
Updated Alaska (AK) Bill Scraper. Vote details still don't currently work, but basic vote information is still added.

### DIFF
--- a/openstates/ak/bills.py
+++ b/openstates/ak/bills.py
@@ -235,22 +235,20 @@ class AKBillScraper(Scraper):
                 classification=atype)
 
         # Get subjects
-        for subj in doc.xpath('//a[contains(@href, "subject")]/text()'):
-            bill.add_subject(subj.strip())
+        # for subj in doc.xpath('//a[contains(@href, "subject")]/text()'):
+        #     bill.add_subject(subj.strip())
 
         # Get versions - to do
-        # text_list_url = (
-        #     "http://www.legis.state.ak.us/"
-        #     "basis/get_fulltext.asp?session=%s&bill=%s"
-        # ) % (session, bill_id)
-        # bill.add_source(text_list_url)
+        text_list_url = (
+            "https://www.akleg.gov/basis/Bill/Detail/%s?Root=%s#tab1_4") % (session, bill_id)
+        bill.add_source(text_list_url)
 
-        # text_doc = lxml.html.fromstring(self.get(text_list_url).text)
-        # text_doc.make_links_absolute(text_list_url)
-        # for link in text_doc.xpath('//a[contains(@href, "get_bill_text")]'):
-        #     name = link.xpath('../preceding-sibling::td/text()')[0].strip()
-        #     text_url = link.get('href')
-        #     bill.add_version_link(name, text_url, media_type="text/html")
+        text_doc = lxml.html.fromstring(self.get(text_list_url).text)
+        text_doc.make_links_absolute(text_list_url)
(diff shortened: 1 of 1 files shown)


openstates/ak/bills.py
```diff
@@ -71,39 +71,41 @@ def scrape(self, chamber=None, session=None):
...
-                '.asp?session=%s&bill1=%s1&bill2=%s9999'
-            ) % (session, abbr, abbr)
-            doc = lxml.html.fromstring(self.get(bill_list_url).text)
-            doc.make_links_absolute(bill_list_url)
-            for bill_link in doc.xpath('//table[@align="center"]//tr/td[1]//a'):
-                bill_url = bill_link.get('href')
-                bill_id = bill_link.text.replace(' ', '')
-                yield from self.scrape_bill(chamber, session, bill_id, bill_type,
-                                            bill_url)
+        bill_list_url = (
```
jamesturk: for new code, let's favor Python 3 style string formatting

either 'foo {}'.format(session) or f'foo {session}'  :)

openstates/ak/bills.py
```diff
@@ -198,109 +205,159 @@ def scrape_bill(self, chamber, session, bill_id, bill_type, url):
...
+                elif "E" in vote_count or "A" in vote_count:
+                    other += int(vote_count[1:])
+
+
+        vote = VoteEvent(
+                bill=bill,
+                start_date=act_date.strftime('%Y-%m-%d'),
+                chamber=act_chamber,
+                motion_text=action,
+                result='pass' if yes > no else 'fail',
```
jamesturk: is this our best proxy for passage?  it is often incorrect so let's double check that there's no passage indicator on the page

openstates/ak/bills.py
```diff
@@ -198,109 +205,159 @@ def scrape_bill(self, chamber, session, bill_id, bill_type, url):
...
-        votes_data = zip(votes_text[1::2], votes_text[2::2])
+        yield vote
 
-        iVoteOnPage = 0
+        
+        # print(action)
 
-        # Process each.
-        for motion, text in votes_data:
+        # Dan attempt at pulling out the vote information
```
jamesturk: could you post a few links to the pages this is scraping so I can review them and offer some suggestions?  (both the metadata and the journal entries)

schneidy: For [HB14](https://www.akleg.gov/basis/Bill/Detail/31?Root=HB14), here are 3 different votes:

- https://www.akleg.gov/basis/Journal/Pages/31?Chamber=H&Bill=HB%20%2014&Page=00727#0727
- https://www.akleg.gov/basis/Journal/Pages/31?Chamber=H&Bill=HB%20%2014&Page=00737#0737
- https://www.akleg.gov/basis/Journal/Pages/31?Chamber=H&Bill=HB%20%2014&Page=00739#0739

The last two of the above links lead to the same journal page with the detailed vote information, but are two different votes for the bill.

jamesturk:
Oof OK, yeah that's not a fun one.  Let's merge this in two pieces, bills
(with versions, actions, etc.) first, then we'll add votes in a second PR.

openstates/ak/bills.py
schneidy: Fixed.

openstates/ak/bills.py
schneidy: Fixed. Now checking action to see if passed or failed.

PR #3066, opened 2019-12-12
NJ: convert session year to int in _init_mdb

TheLocust3:
Fixes: #3046

In the NJ People scraper, just convert the year string (parsed from the session name) to an int when comparing in `_init_mdb`.

### DIFF
--- a/openstates/nj/utils.py
+++ b/openstates/nj/utils.py
@@ -26,7 +26,7 @@ def chamber_name(chamber):
 class MDBMixin(object):
 
     def _init_mdb(self, year):
-        if year < 2018:
+        if int(year) < 2018:
             self.mdbfile = 'DB%s.mdb' % year
             url = 'ftp://www.njleg.state.nj.us/ag/%sdata/DB%s.zip' % (year, year)
             fname, resp = self.urlretrieve(url)


jamesturk:
thanks for the contribution!  @schneidy mind taking a look at this?

schneidy:
@TheLocust3 Thanks for contributing! I'm not seeing a difference in any of the NJ scrapes with the added int(). Should it?

TheLocust3:
@schneidy Previously when I ran NJ People scrapes locally, it would fail with `TypeError: '<' not supported between instances of 'str' and 'int'` (I'm assuming that's what #3046 is referencing). But there shouldn't be any difference with the outputted JSON.

jamesturk:
Ah we aren't currently running people scrapes (instead using
openstates/people for those -- docs are being updated as I write this to
clarify more of this), but thanks for catching this!  The current failure
is in the bill/vote scrapes, the issues created by our automated tool will
also soon include a link to the logs of the failed build to help with this
:)

TheLocust3:
Ah alright, would love to see the logs from the automated tool in the issue itself. Would be a big help trying to figure out what's going on. Thanks for the quick replies!

PR #3069, opened 2019-12-16
NM: Add 2020 Session

showerst:
No data yet, but doing the legwork in advance for when they start prefiling.


showerst:
quick update, they started filing.

PR #3070, opened 2019-12-19
run black on everything, enforce going forward

jamesturk:


showerst:
@jamesturk do you have the ability to easily run the openstates.org scrapers on this branch for a day and make sure it comes back clean? No biggie if not, it would just be a nice to have a practice run at this to be _sure_ it's not breaking any xpath's or regexes.

If not, i'll do some spot checks with scrapers from my computer, eyeballing it, it looks fine, and black seems to have plenty of good tests to make sure the output is idempotent

jamesturk:
Yeah I was considering this, it is big enough that I could see us somehow
having found a black edge case.

I think I can create a docker image from this branch and switch bobsled to
run from there.  If it is a subtle change I'm not sure we'd catch it just
from that, but it'd catch breakages.

showerst:
If that's easy-ish I think it would be smart, like you said we'd at least catch any totally broken scrapers that way. I think it's relatively unlikely that it mangles something but doesn't break the scraper (if nothing else, broken xpaths and regexes should usually result in empty output variables throwing a list index error).

jamesturk:
This is running tonight, assuming nothing noticeably breaks I'll probably
merge over the holiday if that works for you, I figure merging while
minimal work is happening makes sense to avoid merge conflicts.

Issue #3103, opened 2020-01-14
CA: 240+ missing bills


jessemortenson:
State: CA (California)

Noticed that, for at least the last week, a couple hundred bills have been missing. Nothing appears wrong with the CA scraper code itself, as I was able to run locally and retrieve all bills. However it may be an operational issue related to other code changes - see discussion at the bottom.

## Evidence of problem behavior

List appears to be growing as new bills are being filed. California is in session. Examples:

[SB 797](http://leginfo.legislature.ca.gov/faces/billNavClient.xhtml?bill_id=201920200SB797)
[SB 800](http://leginfo.legislature.ca.gov/faces/billNavClient.xhtml?bill_id=201920200SB800)
[SB 848](http://leginfo.legislature.ca.gov/faces/billNavClient.xhtml?bill_id=201920200SB848)

Example query against OS public API:

```
{
  bill(identifier: "SB 797", jurisdiction: "California", session: "20192020") {
    id
    title
  }
}
```

## Evidence that CA scraper code is working

I was (eventually) able to run the California scraper locally. Local output concluded with a count of 3201 bills, which concurs with my own count of bills from the CA legislature website. Local output includes for example:

```
14:46:33 INFO pupa: save bill SB797 in 20192020 as bill_f2fd9264-370e-11ea-be21-0242ac170003.json
```

The output file bill_f2fd9264-370e-11ea-be21-0242ac170003.json looks OK.

## Possible operational problems due to other changes needed?

However in order to run the CA scraper, I had to make several code changes just to be able to run the CA scraper ([see those changes here](https://github.com/jessemortenson/openstates/commit/6478ebf4c7f25e32aa02a7460cb3e7313fa5cd37) - let me know if you want a PR). These were the steps I took:

* A mysql python module was missing, so `~/.poetry/bin/poetry add mysqlclient`
* The `mysql` service from `docker-compose.yaml` was stopping immediately after starting, so I commented out the overridden command from that file, in favor of the image's default command
* The `ca-download` service from `docker-compose.yaml` would not run due to missing packages in the `openstates/openstates` image, so I made changes to `Dockerfile`
* `docker build -t openstates/openstates .`
* `docker-compose up -d mysql`
* `docker-compose up ca-download`
* `docker-compose run --rm scrape ca bills --fastmode --scrape`

jamesturk:
a PR would be welcome, we hadn't gotten CA running after some recent breakage

jamesturk:
[...]

PR #3104, opened 2020-01-14
Changes required to make CA scraper runnable

jessemortenson:
See #3103 for context.


jamesturk:
this looks good, I'll work on getting CA running in production tomorrow

PR #3105, opened 2020-01-15
Illinois Archive Bill Data

schneidy:
Created a new method to add archived bill data for sessions 90-92 (1997-2002).


openstates/il/bills.py
```diff
@@ -2,14 +2,15 @@
...
 import datetime
-
 import pytz
 import scrapelib
 import lxml.html
 from pupa.scrape import Scraper, Bill, VoteEvent
 from pupa.utils import convert_pdf
 
-from ._utils import canonicalize_url
+eastern = pytz.timezone("US/Eastern")
```
jamesturk: IL is in Central time

openstates/il/bills.py
```diff
@@ -266,25 +283,169 @@ def scrape(self, session=None):
...
-                    yield from self.scrape_bill(chamber, session_id, doc_type, bill_url)
-
-        # special non-chamber cases
-        for bill_url in self.get_bill_urls(chamber, session_id, "AM"):
-            yield from self.scrape_bill(
-                chamber, session_id, "AM", bill_url, "appointment"
-            )
+        # Sessions that run from 1997 - 2002. Last few sessiosn before bills were PDFs
+        if session in ["90th", "91st", "92nd"]:
+            print("Within session", session)
```
jamesturk: let's remove these :)

openstates/il/bills.py
```diff
@@ -266,25 +283,169 @@ def scrape(self, session=None):
...
+                            action_date = eastern.localize(action_date)
+                            action_date = action_date.isoformat()
+
+                            action = " ".join(line[2:])
+                            if line[1] == "S":
+                                action_chamber = "upper"
+                            else:
+                                action_chamber = "lower"
+
+                            for pattern, atype in _archived_action_classifiers.items():
```
jamesturk: wouldn't this loop set every action to the last atype if there isn't a break?

look at using an else clause on for https://book.pythontips.com/en/latest/for_-_else.html

openstates/il/bills.py
schneidy: Removed!

openstates/il/bills.py
schneidy: I...should have caught this.

openstates/il/bills.py
schneidy: Fixed!

openstates/il/bills.py
jamesturk: haha, all good

PR #3108, opened 2020-01-18
ma: senate resolution handling

azban:
addresses https://github.com/openstates/openstates/issues/3058

changes:
- specify a list of expected bill types, and check against that. (this was verified against all bills that are coming back from most recent session. only failures were https://malegislature.gov/Bills/191/S2405,AppendixA and other appendix links for S2405. these links are broken anyway, and were already failing.. however https://malegislature.gov/Bills/191/S2405A works, do we want to make an issue for that?)
- non sequitur: sort by bill number when scraping chambers for better usability 

confirmed that this change fixes the import issue for MA/191/SR1


jamesturk:
this looks great! thank you

Issue #2129, opened 2018-02-06
Add transformers to strip whitespace for most string fields


mileswwatkins:
Now that transformers are in production (https://github.com/opencivicdata/pupa/pull/305), opening this issue to add the whitespace-stripping that you'd mentioned, for all fields where trailing spaces wouldn't make sense.

PR #3109, opened 2020-01-18
ma: strip whitespace on persons and committees

azban:
there was significant trailing whitespace on people names and committees, that would need to be stripped later, and also was preventing people name matching. not sure if this is the best way to handle this.


jamesturk:
for now, yes, but a more broad approach is suggested in #2129 which warrants some attention soon

Issue #1384, opened 2017-02-16
NY: Not getting roll calls


showerst:
State: NY  (be sure to include in ticket title)

The NY Senate site has some floor vote breakdowns that we're not collecting - IE [S981](https://www.nysenate.gov/legislation/bills/2017/S981) - [Scraper output ](https://openstates.org/api/v1/bills/ny/2017-2018/S%20981/)

TODO: Does the API offer these for the house side?

jmcarp:
Do you mean the floor vote at https://www.nysenate.gov/legislation/bills/2017/S981? I don't see it on http://legislation.nysenate.gov/api/3/bills/2017/S981. Should we file a ticket at https://github.com/nysenate/OpenLegislation? Grab the floor vote from the html?

PR #3117, opened 2020-01-22
NY Missing Votes

schneidy:
NYS Bill scraper now pulls votes from both Assembly and House. Fixes issue #1384.


openstates/ny/bills.py
```diff
@@ -64,10 +67,24 @@ def _parse_bill_details(self, bill):
...
         ).format(bill_session=bill["session"], bill_id=bill_id)
 
+        # assembly_url = (
+        #     "http://assembly.state.ny.us/leg/?default_fld=&bn={bill_id}"
+        #     "&Summary=Y&Actions=Y&Text=Y"
+        # ).format(bill_id=bill_id)
+        assembly_bill_id = bill_id
+        if (bill_id[-1] == "A") or (bill_id[-1] == "B"):
+            assembly_bill_id = bill_id[:-1]
+        if len(assembly_bill_id) == 3:
```
jamesturk: might be better to use string formatting on this

something like `"{}{:04d}".format(prefix, num)`

jamesturk:
looking good overall

PR #3147, opened 2020-02-03
MD: 2020 scraper updates

showerst:


showerst:
I've left in the old vote parsing code even though it's not currently running, in case someone wants to take on getting votes going.

jamesturk:
ah this is great, I'll work on votes

showerst:
@jamesturk cool. FYI the bill skipping code is in there because (i think) when they initially post a bill, the markup for the bill number is somehow different, but then they update it. Could also just be my bug, that might warrant a closer look.

PR #3148, opened 2020-02-04
IN: Switch title to bill description

csnardi:
Currently, the API's bill title is used as its title to store in the database. However, this title is not very helpful. For instance, this year's HB 1001 title is:

> A BILL FOR AN ACT concerning education.

Versus the description:

> School accountability.

This changes the stored title to be the bill description and not the API title.

Comments are welcome as to which is the preferred title (I presume as the more descriptive item the description is), and whether or not this will create any problems for already existing bills.

### DIFF
--- a/openstates/in/bills.py
+++ b/openstates/in/bills.py
@@ -284,10 +284,10 @@ class INBillScraper(Scraper):
                 self.logger.warning("Bill could not be accessed. Skipping.")
                 continue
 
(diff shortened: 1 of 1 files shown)


jamesturk:
Thanks for this, on the surface it makes sense but I'm going to take a look
at a results diff to make sure.  I'm out today but will try to get to that
when I'm back tomorrow.

Issue #3157, opened 2020-02-05
KY: Pdftotext error


showerst:
State: KY 

OS: Ubuntu 18LTS

When I run the KY bill scraper, I get a bunch of:
```
Internal Error: Cannot handle URI 'https://apps.legislature.ky.gov/record/20RS/sb2/vote_history.pdf'.
```

It looks like we're trying to run pdftotext on a URI, but it only runs on local files. I think this is related to https://github.com/openstates/openstates/commit/6cc3c2401a815aed84d68ded242131b48449eae3#diff-ade79e4caf2248ec6bad10123b04bb3e

@schneidy -- Do you get this error if you:

```pupa update ky bills --scrape``` ?

What about if you do a 

 ```pdftotext https://apps.legislature.ky.gov/record/20RS/sb6/vote_history.pdf``` ?

I'm wondering if it's an OS/pdftotext version difference. No biggie, we can always swap to downloading the file first, but i'm curious if it's just me.

schneidy:
I haven't run into that issue yet.

Here is the command that I've been running:
`docker-compose run --rm scrape ky bills --fast`

@jamesturk Any thoughts on why this issue might be popping up?

jamesturk:
I believe it only works on local files, you're getting complete vote
records as-is?

showerst:
Confirmed, i just ran it in docker and it works fine, including vote records. :man_shrugging: 

I'm going to do a merge later to move that to download the file first, so that it works on stock pdftotext. Interesting difference, though, the scraper python image is also based on ubuntu 18LTS.

showerst:
Resolved via https://github.com/openstates/openstates/pull/3163

PR #3163, opened 2020-02-07
KY: Scrape all versions, proposed amendments, and download pdfs locally first

showerst:


showerst:
This patch grabs proposed amendments and adds them as documents.

KY used to only post one version PDF but it looks like they've started sometimes posting multiple versions, so this update ensures we capture all of them.

Also resolves #3157 by downloading PDF files explicitly before conversion to text.

Head's up @schneidy just in case the proposed amendments are useful to you.

PR #3184, opened 2020-02-27
AZ: Fixes to bill scraper

csnardi:
The AZ bill scraper currently marks first readings as occurring in the wrong chamber due to the upper and lower chamber being flipped in `action_chamber_map`. It misses second readings because their dates have milliseconds whereas those for first readings do not. These bugs are fixed in the scraper.

Additionally, the scraper contains old code sorting the bill actions, seemingly from a time when it was important that actions be sorted according to their dates. Since the website should do this itself now, and the code had the side effect of lower-casing action descriptions, remove the sorting code entirely.

### DIFF
--- a/openstates/az/bills.py
+++ b/openstates/az/bills.py
@@ -56,7 +56,7 @@ class AZBillScraper(Scraper):
         )
         bill.add_source(bill_url)
 
-        self.sort_bill_actions(bill)
+        bill.actions = sorted(bill.actions, key=lambda action: action["date"])
 
         yield bill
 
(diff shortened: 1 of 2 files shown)


showerst:
I'll give this a spin tomorrow and merge when i'm back in the office, if nobody else beats me to it

jamesturk:
Thanks for this!

I'm not sure 100% sure about the sorting issue, the site does do that, but
I'm hoping that we can get the data sorted too for others to use via the
API/etc.

@showerst does this affect you?

showerst:
@jamesturk I'm sorting actions on the frontend, so not directly. 

I agree re:having them sorted for the API / cleaner data though. It also makes debugging the scraper output a little easier, and having them in consistent order makes it easier to cache scraper output, although the state API's order could at least be consistent, i don't remember.

Regardless of what we do with sort, we should pull the chamber fix and stop lowercasing them though.

jamesturk:
yeah I'm inclined to leave (or replace) the sort, but the other changes look good to me

csnardi:
I can back out the changes to sorting then, but it does seem like that code can be simplified. I think this is relatively unique to the AZ scraper(?) so I wonder if there would be any way of sorting objects on database insert for more portability.

jamesturk:
Yeah I was considering that too, I think that'd be an improvement we could
make at the import level

csnardi:
Instead of adding back the old code, I replaced it with a call to `sorted()` which seems to be more reliable and much simpler.

PR #3199, opened 2020-03-04
VA: Fix vote scraping

csnardi:
Vote scraping in VA has been broken, as the method erased any `cached_vote`s after a continue line was commented out. So that actions will continue to be scraped for votes, this change moves the erasing of `cached_vote` and `cached_action` to when a VoteEvent is saved (and no additional data has been scraped for future processing).

This change also fixes scraping of votes which have links come second in sequence. Previously, individual votes were only scraped if the link came first in the history sequence. We will now scrape individual votes no matter when the link comes, and the placeholder link will be erased upon finding the second link.

### DIFF
--- a/openstates/va/bills.py
+++ b/openstates/va/bills.py
@@ -4,6 +4,7 @@ import datetime
 import collections
 import logging
 
+from itertools import tee
 from spatula import Page, Spatula
 from pupa.scrape import Scraper, Bill, VoteEvent
 from .common import SESSION_SITE_IDS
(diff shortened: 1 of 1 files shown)


csnardi:
The code here is somewhat unwieldy, so I'm going to try and clean it up (since my current solution would create other problems).

csnardi:
I pushed a new commit which overhauls this code and should hopefully be more understandable.

showerst:
I'll run this tonight and give it a lookover, thanks!

PR #3205, opened 2020-03-05
MO, NH, NJ, NM, SC, UT: Clean legislator names

csnardi:
Scrapers in these states were often not able to match legislators since they were not formatted as expected. In every state except for NH, the names were formatted as last, first. This change flips the formatting for names in these states to be first, last. In NH, additional spaces were accidentally added to names, which this change prevents.

### DIFF
--- a/openstates/mo/bills.py
+++ b/openstates/mo/bills.py
@@ -498,6 +498,7 @@ class MOBillScraper(Scraper, LXMLMixin):
         bill.add_source(url)
 
         bill_sponsor = clean_text(table_rows[0][1].text_content())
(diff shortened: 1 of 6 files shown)


jamesturk:
historically we haven't done this, favoring what the state has so we can avoid introducing our own error

I'm going to think about this one a bit though, we could adopt a policy of re-ordering names across the board and I don't think it'd be that tricky to implement 

I'm worried about edge cases still, where we introduce error.  If we're doing this I tend to think we should settle on a function that re-orders names in openstates.utils that can be called in these states

jamesturk:
going to close this one, but appreciate it nonetheless

Issue #3187, opened 2020-02-27
CA: adjust downloads


jamesturk:
right now CA downloads grab the latest year, then layer days on top, it looks like the daily downloads might be sufficient

we also want to add support to the download.py script to grab prior years

- [x] limit current-year download to one file (if possible)
- [x] add download.py support for prior years
- [x] test scraper on prior years (however far back we can go without major changes)

schneidy:
Archive data can now go back to 1989 with bills, and 1999 with bills and votes.

Issue #2755, opened 2019-01-07
CA: actions.py need to be updated for false and missing actions


mhedblom:
**CA/actions.py** -- Several of the existing bill action for CA return incorrect results, and many are missing. After reviewing the 2015-2016 session data, I identified about 330 unique descriptions that CA uses, and from those developed this list of unique regex search text and their mappings.

As part of the review I also identified some new Action Type that I have proposed (see https://discourse.openstates.org/t/request-for-new-action-types/227). This table includes mappings for both existing and the proposed actions.

Generally the expression can be a simple case-insensitive search. However, there are some matches that require a period (.), comma (,), or trailing space character. These are noted.

I don't know Python, but can take a stab at updating the source file. However, I do not have an environment to test the change. I do have a file of the CA actions and expected results.

Action   text | Action | Proposed Action | Regex note
-- | -- | -- | --
Amend, | amendment:introduced |   | coma
Adopted. | amendment:passed |   | period
amended | amendment:passed |   |  
tabled | amendment:tabled |   |  
died | bill:failed |   |  
Refused adoption. | bill:failed |   | period
refused passage | bill:failed |   |  
returned to Secretary | bill:failed |   |  
without further action | bill:failed |   |  
Introduced. | bill:introduced |   | period
Adopted and to | bill:passed |   |  
adopted Conference Committee | bill:passed |   | 
adopts Conference Committee | bill:passed |   | 
passed | bill:passed |   |  
to Engrossing and Enrolling | bill:passed |   |  
Read first time | bill:reading:1 |   |  
Read second time | bill:reading:2 |   |  
Read third time | bill:reading:3 |   |  
Read. | bill:reading:3 |   | period
veto stricken from file | bill:veto_override:failed |   |  
failed passage | committee:failed |   |  
refused to concur | committee:failed | concurrence:failed |  
"from committee with   " | committee:passed |   | trailing space
from committee: | committee:passed |   | colon
from committee chair | committee:passed |   |  
from conference committee: | committee:passed |   |  
amendments concurred in | committee:passed:favorable | concurrence:passed |  
from committee without | committee:passed:unfavorable |   |  
to com. | committee:referred |   | period
to coms. | committee:referred |   | period
Ordered to Conference   Committee | committee:referred | concurrence:referred |  
enrolled | governor:received |   |  
approved by the Governor. | governor:signed |   | period
vetoed | governor:vetoed |   |  
Approved by the Governor with   item veto | governor:vetoed:line-item |   |  
[...]

PR #3207, opened 2020-03-06
CA Archive Data

schneidy:
Updated CA to scrape archive data. You can now scrape bills and votes all the way back to 1999, and just bills back to 1989.

Additionally, the number of files has been limited to either the year or the most recent files.

Related to issue #3187


schneidy:
Added actions per issue #2755

PR #3212, opened 2020-03-09
CA: Updates to action parser for joint rule 56

showerst:
The CA parser was marking bills that had died as 'passed committee' due to the actions starting with 'From Committee:' -- this fixes that.

Ex: http://leginfo.legislature.ca.gov/faces/billHistoryClient.xhtml?bill_id=201920200AB700

@schneidy I don't think this conflicts w/ your CA work but let me know if so.


schneidy:
It shouldn't cause any issues and should be an easy merge.

PR #3224, opened 2020-03-17
CA: Added in additional long titles

schneidy:
Added in additional long titles by grabbing the title from the status page on the bills' pages.

Should fix issue 1380.


jamesturk:
OK going to merge this, but might revert if there's a major regression in terms of time

Issue #2998, opened 2019-06-25
OH: Bill vote history titles are broken


pierrefortier:
State: Ohio

Titles in the "Recent votes" section seem broken.

![Screen Shot 2019-06-25 at 8 53 35 AM](https://user-images.githubusercontent.com/37414904/60100336-89a01800-9727-11e9-933f-1494d74c6bc1.png)

jamesturk:
can you provide some example URLs?

pierrefortier:
Sure! For instance, [SB 6](https://openstates.org/oh/bills/133/SB6/) or [SB 23](https://openstates.org/oh/bills/133/SB23/).

schneidy:
Started to fix with branch oh_bill_vote_history_titles. Can't fully update the action dictionary for votes until [Ohio API documentation](http://docs.api.lis.state.oh.us/) is fixed.

jamesturk:
openstatesorg=# select distinct(motion_text) from opencivicdata_voteevent ve join opencivicdata_legislativesession ls on ve.legislative_session_id=ls.id join opencivicdata_jurisdiction j on ls.jurisdiction_id=j.id where j.name='Ohio';

  motion_text
----------------
 confer_713
 confer_712
 msg_506
 motion_913
 motion_909
 final_510
 imm_consid_360
 pass_300
 msg_reso_503
 intro_103
 intro_108
 intro_101
 concur_606
 crpt_301
 adopt_reso_110
 intro_102
 concur_608
 msg_507
 pass_301
(19 rows)

likely possible to map these

jamesturk:
further investigation:

select bs.url from opencivicdata_voteevent ve join opencivicdata_bill b on ve.bill_id=b.id join opencivicdata_billsource bs on bs.bill_id=b.id where motion_text='confer_713'; 

 lets us inspect the source, but https://www.legislature.ohio.gov/legislation/legislation-votes?id=GA132-HB-26 doesn't have motions :(

actions do line up well:
- confer_713: Conference report agreed to

looks like if we do this ~18 more times we'd have a fix

schneidy:
I followed @jamesturk lead and went through all 18 of the motion_texts that he provided. I ended up adding in one more when testing the script (third_407).

PR #3225, opened 2020-03-17
OH: Added in a vote motion dictionary

schneidy:
Added in a vote motion dictionary to try and clear up what the vote motions actually were.

Relates to Issue #2998.


jamesturk:
mind adding a comment explaining a good way to populate this dictionary for us in the future?

Issue #3022, opened 2019-10-08
ME : Missing additional bill versions/document attachments.


mohit-complianceai:
State: Maine 

Eg : - 
```
open states bill id - ocd-bill/bf8036dc-ec1f-46a1-9cdd-d7f941369153
source url - http://legislature.maine.gov/LawMakerWeb/summary.asp?ID=280071262
bill identifier - LD 459
```

Open states provides only one bill version - initial version, whereas the source has lot more document attachments present.

PR #3230, opened 2020-03-18
ME: additional bill versions

schneidy:
Added code to grab all additional Bill Versions and Documents for ME.

Should fix issue #3022 

Some of the code can be rewritten and made less verbose.

### DIFF
--- a/openstates/me/bills.py
+++ b/openstates/me/bills.py
@@ -250,6 +250,7 @@ class MEBillScraper(Scraper):
 
                     # various versions: billtexts, billdocs, billpdfs
                     v_links = []
+
                     for v in range(
                         0, len(vdoc.xpath('//span[@class="story_heading"]')) - 1
                     ):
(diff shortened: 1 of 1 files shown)


openstates/me/bills.py
```diff
@@ -249,37 +249,84 @@ def scrape_bill(self, bill, chamber):
...
-                    bill_text_pdf = vdoc.xpath(
-                        'string(//a[contains(@href, \
-                        "getPDF.asp")][1]/@href)'
-                    )
-
-                    if bill_text_html:
-                        bill.add_version_link(
-                            "Initial Version", bill_text_html, media_type="text/html"
+                    v_links = []
+                    for v in range(
```
jamesturk: would it make more sense to iterate over the result of the vdoc.xpath(...story_heading) instead of using this?

openstates/me/bills.py
```diff
@@ -249,37 +249,84 @@ def scrape_bill(self, bill, chamber):
...
+                            f"//span[@class='story_heading']/following::a[contains(@href, 'rtf')]/@href"
+                        )
+                        version_fiscal_pdf = vdoc.xpath(
+                            f"//span[@class='story_heading']/following::a[contains(@href, 'fiscalpdfs')]/@href"
                         )
+                        version_fiscal_html = vdoc.xpath(
+                            f"//span[@class='story_heading']/following::a[contains(@href, 'fiscalnotes')]/@href"
+                        )
+
+                        if len(version_pdf) > v:
```
jamesturk: I think this definitely needs some comments to be more clear, it is unclear what these checks are doing (what does it mean for the len(pdf) > v and why are the version_pdf[v] in v_links checks there?

openstates/me/bills.py
schneidy: I tried to iterate over the result of story_heading but kept running into xpath issues where it kept picking up all of the version types urls during each loop through (eg. all the pdf urls were returned up at once).

openstates/me/bills.py
schneidy: Added some additional comments to clear it up.

PR #3250, opened 2020-03-30
TN: Avoid adding older actions

showerst:
We've seen a few times where due to a (presumed) site bug, TN has displayed previous session actions and votes on current session bill pages. So e.g. 2020 HR 5 will show actions from 2017 HR 5, a totally different bill.

This patch skips actions older than 2 months before the session start date (to allow for prefiles).

### DIFF
--- a/openstates/tn/bills.py
+++ b/openstates/tn/bills.py
@@ -158,7 +158,7 @@ def categorize_action(action):
     return list(types), attrs
 
 
-def actions_from_table(bill, actions_table):
+def actions_from_table(self, bill, actions_table):
     """
     """
     action_rows = actions_table.xpath("tr")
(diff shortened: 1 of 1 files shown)


openstates/tn/bills.py
```diff
@@ -158,7 +158,7 @@ def categorize_action(action):
     return list(types), attrs
 
 
-def actions_from_table(bill, actions_table):
+def actions_from_table(self, bill, actions_table):
```
jamesturk, line 161: looks good to me, but curious if it'd make sense to just move this into the class body now that it takes self?

showerst: Yeah this makes sense. I also see a stray print statement I left in --  I'll take another look this afternoon or tomorrow and maybe refactor it to be a little cleaner.

Issue #2200, opened 2018-03-12
WI: Scrape additional documents


mileswwatkins:
Wisconsin exposes a lot of nice documents/links on bill pages, that we don't currently capture. Eg:

<img width="555" alt="screen shot 2018-03-12 at 14 17 53" src="https://user-images.githubusercontent.com/4959135/37301857-3c240df0-2600-11e8-8615-acdb967389bf.png">

Most of these are HTML, but some are also PDF. Regardless, it'd be cool to have these scraped as additional documents, in the long term.

PR #3251, opened 2020-03-30
WI: Added documents that were originally skipped.

schneidy:
Added documents that were originally skipped.

Related to issue #2200

### DIFF
--- a/openstates/wi/bills.py
+++ b/openstates/wi/bills.py
@@ -209,7 +209,7 @@ class WIBillScraper(Scraper):
                     if extra_a.text:
                         bill.add_document_link(extra_a.text, extra_a.get("href"))
             else:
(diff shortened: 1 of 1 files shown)


jamesturk:
couple of questions:

- what types of documents are added?  are they all HTML as the media type says? (the issue said otherwise but maybe things changed?)
- did you get a sense of what the documents were from scraping/looking at the results?  can you list the types of documents that might be included by this?

schneidy:
Some of the docs included committee meeting minutes, other resolutions, bill hearing materials, and links to other bills.

I've only run the most recent session and didn't find any PDFs. I can run the script through past sessions.

jamesturk:
alright, sounds good, not too worried about prior sessions w/ this for now

Issue #3258, opened 2020-03-31
KS: Sponsor name issues


schneidy:
Kansas has additional words in the sponsor name columns before the name, such as "Senator" and "Representative".
_Example_: Representative Francis

Additionally, committee names can be found within the sponsor name columns. They will mostly appear with the words "Committee on" within the voter name.
_Example_: Committee on Judiciary

Sessions affected: 2017-2018; 2019-2020

PR #3266, opened 2020-04-01
KS: Removed committees from being sponsors. Removed 'Representative' …

schneidy:
KS: Removed committees from being sponsors. Additionally removed 'Representative' and "Senator" from before sponsor's names.

Related to issue #3258


jamesturk:
sorry for confusion- committees can be sponsors, we want to leave them be

schneidy:
> sorry for confusion- committees can be sponsors, we want to leave them be

Easy fix. Will push updates in a second.

PR #3272, opened 2020-04-02
WIP: switch to openstates-core

jamesturk:


jamesturk:
openstates/openstates-core#3

scrapers/nh/bills.py
```diff
@@ -3,9 +3,9 @@
 from collections import defaultdict
 import pytz
 
-from pupa.scrape import Scraper, Bill, VoteEvent as Vote
+from openstates.scrape import Scraper, Bill, VoteEvent as Vote
 
-from openstates.nh.legacyBills import NHLegacyBillScraper
+from .nh.legacyBills import NHLegacyBillScraper
```
doubleswirve, line 8: Just a heads up in case someone else runs into it, I hit the following error running locally:
```txt
ModuleNotFoundError: No module named 'nh.nh'
```

If I switch to `from nh.legacyBills` or `from .legacyBills` it seems to work fine. Thanks.

jamesturk:
ah thanks for catching this, I'll push a fix

Issue #3255, opened 2020-03-31
FL: Voter name issues


schneidy:
Flordia has voter names with extra spaces before their names. 

Committee names can additionally be found in place of voter names. This affects around 30 votes.

Sessions affected: 2018, 2019, 2020

PR #3273, opened 2020-04-02
FL Voter Name Issues

schneidy:
Cleaned up voter namers that had extra spaces and the letters VA at the end.

Removed strip() to voter names to clean up extra spaces.

Resolves issue #3255


openstates/fl/bills.py
```diff
@@ -461,7 +463,11 @@ def handle_page(self):
         # set voters
         for vtype, voters in votes.items():
             for voter in voters:
-                vote.vote(vtype, voter)
+                voter = voter.strip()
+                if "  VA" in voter:
```
jamesturk, line 467: this seems like it could cause a weird error later, can we just replace the "VA" or at least comment what is going on here?

schneidy: Added in a comment explaining why I remove the VA and the 2 extra spaces. Here is an example name that this if statement takes care of:
"Cruz                                                               VA"

Issue #3259, opened 2020-03-31
MI: Sponsorship name issues


schneidy:
Michigan has just district numbers as sponsors.
Example: 42
Sessions affected: 2017-2018

Occasionally, full sentences will be in place of the voter name.
Example: "Was read a third time and passed, a majority of the members serving voting therefor, by yeas and nays, as follows:"
Sessions affected: 2019-2020

PR #3275, opened 2020-04-02
MI Sponsorship and Voter Name Issues

schneidy:
Cleaned voter names that included extra spaces, multiple names in one line, or were sentences that should not have been recorded as a voter.

Resolves issue #3259

### DIFF
--- a/openstates/mi/bills.py
+++ b/openstates/mi/bills.py
@@ -111,7 +111,7 @@ class MIBillScraper(Scraper):
             else:
                 classification = "primary"
             bill.add_sponsorship(
-                name=name,
+                name=name.strip(),
                 chamber=chamber,
                 entity_type="person",
                 primary=classification == "primary",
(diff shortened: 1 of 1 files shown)


openstates/mi/bills.py
```diff
@@ -193,13 +193,27 @@ def scrape_bill(self, chamber, session, bill_id):
...
-                            vote.no(name)
+                            if len(name.split()) < 5:
+                                vote.no(name.strip())
+                            if session == "2017-2018":
+                                names = name.split("\t")
+                                for n in names:
+                                    vote.no(name.strip())
                         for name in results["other"]:
-                            vote.vote("other", name)
+                            if len(name.split()) < 5:
```
jamesturk, line 211: this will need comments to be clear for sure

openstates/mi/bills.py
```diff
@@ -193,13 +193,27 @@ def scrape_bill(self, chamber, session, bill_id):
...
+                                vote.yes(name.strip())
+                            if session == "2017-2018":
+                                names = name.split("\t")
+                                for n in names:
+                                    vote.yes(name.strip())
                         for name in results["no"]:
-                            vote.no(name)
+                            if len(name.split()) < 5:
+                                vote.no(name.strip())
+                            if session == "2017-2018":
```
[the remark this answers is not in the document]
schneidy: Refactored the code to share the same cleaning function.

openstates/mi/bills.py
schneidy: Comment added. It's to prevent sentences like "House Bill No. 4451, entitled" being entered in the voter name column.

Issue #3254, opened 2020-03-31
WY: Voter name issues (historical only)


schneidy:
Wyoming has things in the voter name column that is not voters, such as:
* "1/11/2018 Bill Number Assigned 2/9/2018 S Received for Introduction 2/12/2018 S Introduced"
* "Do Pass 7-1-1-0-0 ROLL CALL"
* "Considered 2/20/2018"
* "Joint Corporations, Elections & Political Subdivisions Interim Committee"
* "Referred to S01 - Judiciary 24-6-0-0-0 ROLL CALL"
* "renumber as necessary. "

Sessions affected: 2017, 2018

PR #3277, opened 2020-04-02
WY Voter Name Issues

schneidy:
Added a function to remove the extra spaces in a voter name for Wyoming.

I haven't been able to replicate the voter name issues that were mentioned in issue #3254. 

@jamesturk, we should re-run the 2017 and 2018 sessions to see if issues are still arising with voter and sponsorship names.


openstates/wy/bills.py
```diff
@@ -262,6 +267,11 @@ def scrape_vote(self, bill, vote_json, session):
 
         yield v
 
+    def clean_voter_name(self, name):
```
jamesturk, line 270: let's use .strip() for cases like this as it'll work even if extra whitespace is added

schneidy: @jamesturk Just pushed a fix.

PR #3285, opened 2020-04-06
sd: get version link for pdf

azban:
the scraper was only grabbing the html link. modified to grab the pdf link and also added the dates to versions.

PR #3286, opened 2020-04-06
or: encode whitespace in urls

azban:
the Oregon site has URL's with whitespace in them that were not being encoded, causing issues for people following the link outside of a browser.

i included a utility function to fix the URL for this specific issue, based off of a [werkzeug util](https://github.com/pallets/werkzeug/blob/51860032cc780fa4a120b4a3c0457a36bbe91715/src/werkzeug/urls.py#L608). don't know if it makes sense to integrate this more generally.. it seems like this is fairly isolated.

i also made a [PR](https://github.com/opencivicdata/pupa/pull/333) to tighten up validation in pupa to catch this.

### DIFF
--- a/openstates/or/bills.py
+++ b/openstates/or/bills.py
@@ -1,10 +1,15 @@
 import datetime
-import re
 import logging
+import re
 
 from pupa.scrape import Scraper, Bill
 from .apiclient import OregonLegislatorODataClient
-from .utils import index_legislators, get_timezone, SESSION_KEYS
+from .utils import (
+    index_legislators,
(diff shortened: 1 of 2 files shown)


azban:
i'm not seeing these changes reflected in API requests (or for the SD pr here https://github.com/openstates/openstates-scrapers/pull/3285). is there a delayed release or should we expect to see scrapes after merges include new changes?

jamesturk:
Looks like there was a hiccup that kept this from being deployed to the
scrape runner.  Fixing that now, and it should run today

PR #3288, opened 2020-04-07
VA: Initial commit for csv_bills - Work in Progress

schneidy:
Work in Progress for creating a script to scrape the VA CSVs.

### DIFF
--- a/openstates/va/csv_bills.py
+++ b/openstates/va/csv_bills.py
@@ -1,0 +1,18 @@
+# from pupa.scrape import Scraper, Bill, VoteEvent
+
+
+# Load members of legislative
+def load_members():
+    # Grab Members.csv and save all of them locally
+    return True
+
+
+# Start scrapping Bills
+def scrape_bills():
+    # Grab BILLS.csv
+    return True
+
+
+if __name__ == "__main__":
+    load_members()
+    scrape_bills()


openstates/va/csv_bills.py
```diff
@@ -0,0 +1,368 @@
...
+                            vote_result = "abstain"
+                        self._votes[history_refid].append(
+                            {"member_id": member, "vote_result": vote_result}
+                        )
+        self.warning("Total Votes Loaded: " + str(len(self._votes)))
+
+    def load_bills(self):
+        resp = self.get(self._url_base + "BILLS.CSV").text
+        reader = csv.reader(resp.splitlines(), delimiter=",")
+        # ["Bill_id" (0), "Bill_description (1)", "Patron_id (2)", "Patron_name (3)", "Last_house_committee_id (4)",
```
jamesturk, line 145: let's make sure to comment where these come from

for this many fields thinking it might make the code more clear overall if we use DictReader & pass in the field list

https://docs.python.org/3/library/csv.html#csv.DictReader

docker-compose.yml
```diff
@@ -13,6 +13,8 @@ services:
     - PYTHONPATH=./openstates
     - NEW_YORK_API_KEY
     - INDIANA_API_KEY
+    - VA_USER
+    - VA_PASSWD
```
jamesturk, line 17: to bring these in line with the other environment variables for clarity can we rename these to

VIRGINIA_FTP_USER
VIRGINIA_FTP_PASSWORD

schneidy: Resolved.

openstates/va/csv_bills.py
schneidy: Switched over to using csv.DictReader()

jamesturk:
looking good

realized I want to give a heads up to @showerst about this one, as it requires an FTP login 

but I think this'll get us past the flakiness of VA's site nicely

schneidy:
> looking good
> 
> realized I want to give a heads up to @showerst about this one, as it requires an FTP login
> 
> but I think this'll get us past the flakiness of VA's site nicely

It filled in the actions that were missing and only took a fraction of the time.

showerst:
ooh sweet! 

Can you document somewhere who to reach out to for an FTP login? Looks like it's here -- https://lis.virginia.gov/SiteInformation/ftp.html

jamesturk:
@schniedy mind adding that to the docs page here:
https://docs.openstates.org/en/latest/contributing/state-specific.html

schneidy:
@jamesturk Added PR for documentation update here: https://github.com/openstates/documentation/pull/63

PR #3295, opened 2020-04-14
ME: Voter Name Issues

schneidy:
Maine:

Voter name issues:

- [x] Voter names fully capitalized.
- [x] Voter names include " Of {district name}"

Sessions to rerun after merging: 128, 129

### DIFF
--- a/openstates/me/bills.py
+++ b/openstates/me/bills.py
@@ -408,6 +408,8 @@ class MEBillScraper(Scraper):
             if ("'" in name) and (name[2] != " "):
                 name = "".join(name[:2]) + name[2].capitalize() + "".join(name[3:])
             vtype = row.xpath("string(td[4])")
+            if " Of " in name:
+                name = name.split(" Of ")[0]
             if vtype == "Y":
                 vote.vote("yes", name)
             elif vtype == "N":
--- a/openstates/me/bills.py
+++ b/openstates/me/bills.py
@@ -2,7 +2,7 @@ import re
 import html
 import socket
 import datetime
-
+import string
 import requests
 import lxml.html
 import scrapelib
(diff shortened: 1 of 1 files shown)


openstates/me/bills.py
```diff
@@ -397,10 +403,15 @@ def scrape_vote(self, bill, motion, url):
...
         for row in member_cell.xpath("../../tr")[1:]:
-            name = row.xpath("string(td[2])")
-            # name = name.split(" of ")[0]
-
+            name = string.capwords(row.xpath("string(td[2])"))
+            # Handles last names like O'Connell
+            if ("'" in name) and (name[2] != " "):
+                name = "".join(name[:2]) + name[2].capitalize() + "".join(name[3:])
             vtype = row.xpath("string(td[4])")
+            # Removes district name from voter name
```
jamesturk, line 411: we need to leave these for disambiguation

openstates/me/bills.py
```diff
@@ -397,10 +403,15 @@ def scrape_vote(self, bill, motion, url):
 
         member_cell = page.xpath("//td[text() = 'Member']")[0]
         for row in member_cell.xpath("../../tr")[1:]:
-            name = row.xpath("string(td[2])")
-            # name = name.split(" of ")[0]
-
+            name = string.capwords(row.xpath("string(td[2])"))
```
jamesturk, line 406: I think instead of this level of name mangling we'll prefer case insensitive comparisons in the import code

jamesturk:
figuring out the right balance between cleaning (white space, weird punctuation, etc.) and potentially introducing error is a tough one, but I'm gonna say that this one is probably past the point where we want to be manipulating names.  gonna close this one as we definitely don't want to change the "of" portion (similar to the decision on the districts being included) and I don't think the other change to handle O' is worth the complexity either

Issue #3296, opened 2020-04-14
AR: People scrapper no longer functions with updated website.


ColbyReed:
State: Arkansas

Since the legislature site update the people scraper can no longer grab new data.

PR #3297, opened 2020-04-14
AR: Updated paths for people scrapping

ColbyReed:
Paths are updated to scrap for people data on new website.

### DIFF
--- a/openstates/ar/people.py
+++ b/openstates/ar/people.py
@@ -30,7 +30,7 @@ class ARLegislatorScraper(Scraper):
             '/tbody'
             '/tr'
             '/td[not(text()[contains(.,"(Deceased)")])'
-            'and not(text()[contains(.,"(Removed)")])' 
+            'and not(text()[contains(.,"(Removed)")])'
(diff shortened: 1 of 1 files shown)


ColbyReed:
Fixes #3296

ColbyReed:
Should the people scraper skip persons who have resigned, are removed, or are deceased? The updated list on the website includes 3 members with these tags. The fix I implemented currently scans these individuals, increasing the number of persons to 138 rather than 135.

List for reference:
https://www.arkleg.state.ar.us/Legislators/List

jamesturk:
Hi Colby, welcome!

Thanks for the contribution.  We don't typically include retired/etc. members as we typically use these scrapers now to update github.com/openstates/people which has historical info as well.

Right now the process of running this scraper and getting the data into openstates/people is more complex than we'd like, so there's a bit of a delay between this and updating there just to let you know (in case you were hoping to see the data live immediately).

ColbyReed:
Happy to help! I'll add a check to skip names with those tags. 

Also no problem about a delay, just noticed this scrapper was failing after the site updated.

PR #3315, opened 2020-04-29
UT: Add 2020 4th Special

showerst:


scrapers/ut/__init__.py
```diff
@@ -231,6 +231,14 @@ class Utah(State):
             "start_date": "2020-04-15",
             "end_date": "2020-04-17",
         },
+        {
+            "_scraped_name": "2020 4th Special Session",
+            "classification": "special",
+            "identifier": "2020S4",
+            "name": "2020 4th Special Session",
+            "start_date": "2020-04-27",
+            "end_date": "2019-05-01",
```
azban, line 240: is it an issue that this is 2019?

showerst: Apologies, that's a typo on my end. Shouldn't cause any errors for openstates but it's still bad data. I'll do a fix PR tomorrow.

PR #3324, opened 2020-05-09
AK: edit document links, add conferees

brjezierski:
Add related document format as pdf, strip .pdf from doc names, and add conferees from House and Senate as related docs


jamesturk:
looks good to me!

PR #3335, opened 2020-05-14
TX: add some missing info for resolutions, add PDF version links for all bill types

azban:
i found that TX resolutions did not contain versions, and looks like were also missing fiscal notes, witnesses and analyses.

the issue is that the file paths for resolutions omit the 'R' in the identifier, causing issues when we look up associated files by bill ID. i modified the function used to extract bill id to fix this when we create the lookup mapping.

i also added PDF version links for all documents. let me know if you concerns about that or want me to separate that out to a different PR.

### DIFF
--- a/scrapers/tx/bills.py
+++ b/scrapers/tx/bills.py
@@ -58,6 +58,17 @@ class TXBillScraper(Scraper, LXMLMixin):
             else:
                 yield "/".join(["ftp://" + root, dir_, name])
 
+    @staticmethod
+    def _get_bill_id_from_file_path(file_path):
+        bill_id = file_path.split("/")[-1].split(".")[0]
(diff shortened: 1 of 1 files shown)


azban:
i attempted to get rid of the need to crawl for file names, and use the history XML instead. this works for everything, except for witness lists. see https://github.com/azban/openstates/pull/1/files and let me know what you think and if you want to include those changes here. functionally isn't much different, other than grabbing version descriptions for some links that previously were just using the bill type in the note. but seems a little cleaner to me.

showerst:
@jamesturk this looks good to me -- ok to merge or do you want to review?

@azban sorry for the delay on this, thanks for submitting!

jamesturk:
LGTM 👍 -- sorry hadn't had a chance to review yet

PR #3343, opened 2020-05-20
DC: New API

schneidy:
Rewrote the DC bill scraper to work with the [new API](https://lims.dccouncil.us/api/help/index.html).

Should resolve issue #3341


scrapers/dc/bills.py
```diff
@@ -26,546 +27,269 @@ class DCBillScraper(Scraper):
...
+                                    yes_count = no_count = other_count = 0
+                                    for leg_vote in act["voteDetails"]["votes"]:
+                                        mem_name = leg_vote["councilMember"]
+                                        if leg_vote["vote"] == "Yes":
+                                            yes_count += 1
+                                            v.yes(mem_name)
+                                        elif leg_vote["vote"] == "No":
+                                            no_count += 1
+                                            v.no(mem_name)
+                                        else:
```
jamesturk, line 277: what kinds of other votes were you seeing?  we might want to categorize them now

scrapers/dc/bills.py
```diff
@@ -26,546 +27,269 @@ class DCBillScraper(Scraper):
...
+
+                                    v.set_count("yes", yes_count)
+                                    v.set_count("no", no_count)
+                                    v.set_count("other", other_count)
+                                    yield v
 
-            bill.add_document_link(doc_type, doc_url, media_type=mimetype)
+                yield bill
 
     def date_format(self, d):
```
jamesturk, line 288: is this still used?  I wasn't seeing it

schneidy: Good catch! Removed it.

scrapers/dc/bills.py
schneidy: Added in absent, recused (abstain), and Present (other) as vote categories.

jamesturk:
looking good, cc @showerst if you want to take a look too

jamesturk:
also Dan can you add docs to the doc repo about the API key in the same style as the NY/etc. ones?

showerst:
Are you guys getting a zillion insecureRequestWarning -s or is that just my env? If it's a problem for y'all too i can PR the code to disable that warning, DC is bad about certificates so it's a good idea to keep the verify=False

schneidy:
> also Dan can you add docs to the doc repo about the API key in the same style as the NY/etc. ones?

schneidy:
> also Dan can you add docs to the doc repo about the API key in the same style as the NY/etc. ones?

Adding now.

schneidy:
> Are you guys getting a zillion insecureRequestWarning -s or is that just my env? If it's a problem for y'all too i can PR the code to disable that warning, DC is bad about certificates so it's a good idea to keep the verify=False

I'm also getting all of those warnings.

showerst:
> I'm also getting all of those warnings.

Ok, i'll PR (back to this branch not master) a fix for that.

showerst:
I added a separate MR for warning suppression, and grabbing the law number. Otherwise good to merge from my end!

showerst:
Oops, link just in case it doesn't notify -- https://github.com/openstates/openstates-scrapers/pull/3344

PR #3353, opened 2020-05-27
tx: use history xml for links when available

azban:
i attempted to get rid of the need to crawl for file names, and use the history XML instead. this works for everything, except for witness lists.

functionally isn't much different, other than grabbing version descriptions for some links that previously were just using the bill type in the note. but seems a little cleaner to me.


showerst:
Cool, this code is more readable if nothing else. I'll do a test run (which might take a while, TX isn't fast as you've probably noticed lol) and merge if it all comes back clean.

PR #3368, opened 2020-06-08
OH: Do not error when an unrecognized vote code is encountered

jessemortenson:
instead fallback to 'results' field and warn

Attempted to solve #3363 This is similar to the behavior now implemented for the `action_dict`. As mentioned in the issue, sounds like these codes are not documented anywhere. In the case of the error encountered, it's not clear what relationship the code has to any text that the end user would see when [viewing the vote in the legislature.](https://www.legislature.ohio.gov/legislation/legislation-votes?id=GA133-HB-425). Decided to fall back to the "passed" value on the vote data because that matches the string found on the leg page ("Passed")

### DIFF
--- a/scrapers/oh/bills.py
+++ b/scrapers/oh/bills.py
@@ -501,6 +501,12 @@ class OHBillScraper(Scraper):
             except KeyError:
                 motion = v["motiontype"]
 
+            if motion in self._vote_motion_dict:
+                motion_text = self._vote_motion_dict[motion]
+            else:
+                self.warning('Unknown vote code {}, please add to _vote_motion_dict'.format(motion))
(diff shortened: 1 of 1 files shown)


jessemortenson:
I'm running the full OH bills scrape locally to test this, but it hasn't cmopleted yet. It worked fine for fastmode scrape.

showerst:
I know oh takes forever so just ping me when your test is done and i'll merge.

Thanks, this is clearly a better solution than playing whack-a-mole with motion codes.

jessemortenson:
Full scrape test ran successfully :+1:

jamesturk:
thanks for this!

PR #3403, opened 2020-07-02
NV: Add 81st 2021 session to list to fix scraper runtime error

jessemortenson:
This fixes the "missing session" error currently causing the NV scraper to fail.

However, the session is not listed as the ultimate in the array of sessions, because doing so causes a different error. The new session [appears on the NV website](https://www.leg.state.nv.us/Session/81st2021/) but there is a [related URL that scraping depends on](https://www.leg.state.nv.us/Session/81st2021/Reports/TablesAndIndex/2021_81-index.html) that does not work (404). I'm guessing the new session is only partially implemented at this point.

What's the protocol here? On one hand, the scraper is unable to run if the new session isn't added. On the other hand, adding the session as the penultimate may make it harder to remember to change this once again, closer to the actual session (so the new session is ingested).

### DIFF
--- a/scrapers/nv/__init__.py
+++ b/scrapers/nv/__init__.py
@@ -95,6 +95,14 @@ class Nevada(State):
             "end_date": "2017-06-06",
         },
         {
+            "_scraped_name": "81st (2021) Session",
+            "classification": "primary",
+            "identifier": "81",
+            "name": "2021 Regular Session",
+            "start_date": "2021-02-01",
+            "end_date": "2021-06-01",
+        },
+        {
             "_scraped_name": "80th (2019) Session",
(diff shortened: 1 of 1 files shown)


showerst:
In a case like this where they add a partial or 'empty' session, you can add it to the ```ignored_scraped_sessions``` variable. This will skip the check. Once the session is scrapeable, we'll just remove it from that variable and restore it into the sessions list.

If you have the data handy it's nice to add the actual session's metadata just commented out so make life easier when it does start up, but by no means a requirement.

I'm going to close this patch and do one that ignores it instead.

PS: Just noticed you dropped a few PRs last week, apologies for the late responses!

PR #3404, opened 2020-07-03
VT: Fix invalid date string for bill action

jessemortenson:
VT scraper failed the last two days on a date string that failed validation. Looks like VT has typos in the dates of actions once in a while. Tried to fix here a little more generally than the previous fix (finding/replacing a specific typo string). Seems better to have a small number of stupid dates on bill actions (the year 200) rather than a scraper that totally stops if one action on one bill doesn't validate.

Interestingly `strftime` formats dates differently on different platforms! so `%Y` doesn't necessarily give you 4 digits back. This might point to a need for a more general error-correction method across scrapers.

### DIFF
--- a/scrapers/vt/bills.py
+++ b/scrapers/vt/bills.py
@@ -232,13 +232,19 @@ class VTBillScraper(Scraper, LXMLMixin):
                 # Manual fix for data error in
                 # https://legislature.vermont.gov/bill/status/2020/H.511
                 action["StatusDate"] = action["StatusDate"].replace("/0209", "/2019")
+                action_date = datetime.datetime.strftime(
+                    datetime.datetime.strptime(action["StatusDate"], "%m/%d/%Y"),
+                    "%Y-%m-%d"
+                )
+                # strftime doesn't always pad year value (%Y)  (https://bugs.python.org/issue32195)
+                # and sometimes this state has typos in year part of the StatusDate value
+                # which can cause validation errors, so fix leading zeroes if they are missing
+                if action_date.find("-") < 4:
+                    action_date = ("0" * (4 - action_date.find("-"))) + action_date
 
                 bill.add_action(
                     description=re.sub(HTML_TAGS_RE, "", action["FullStatus"]),
-                    date=datetime.datetime.strftime(
-                        datetime.datetime.strptime(action["StatusDate"], "%m/%d/%Y"),
-                        "%Y-%m-%d",
-                    ),
+                    date=action_date,
                     chamber=actor,
                     classification=action_type,
                 )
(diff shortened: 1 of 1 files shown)


jessemortenson:
Looks like someone else put in the fix for the specific case that was erroring-out on the scraper. I believe my proposed changes are complementary with this (should be placed after the specific-case-fix)

showerst:
Trying to follow the logic here w/out running the code -- so the idea is if they do like "-202" to convert it to "-2020" ?

jessemortenson:
Sorry I missed this comment a couple weeks back! Not quite: in summary:

- strftime() doesn't always return leading zeroes for the `Y` value, even though docs suggest it should
- therefore a little extra logic to pad the beginning of the year string with however many zeroes are necessary to get it back to a 4-digit number (in this state where we know at least two cases where a typo in the year).
- resulting in: "-202" would be converted to "0202". 

That value is likely wrong, but it at least won't throw an error that stops bill processing. As stated above, it seems better to have the occasional bad year value than have to make a manual fix every once in a while to individual typos.

jessemortenson:
Been a week - just want to check: is there a concern/question with this PR?

jamesturk:
do we know of cases where this has happened besides the -202 issue?  I'm inclined to suggest

re.sub("-202$", "-2020", broken_dt)

jessemortenson:
It's already happened twice and been fixed twice in the way you suggest in this file. What I'm saying about strftime() is that it leaves an obvious category of data validation errors open where "Y" doesn't always return 4 digits. We don't control the occurrence of typos. Why keep fixing typos as they come up (at the cost of attention, stoppage to data ingestion, operational investigation) instead of closing that category of data validation errors?

showerst:
I'd rather have the scraper break and we have to fix it, than have it output bad data, or assume all downstream consumers know to check for magic numbers. 

@jamesturk re:your fix, what if it happens in 2021?

Since these are actions, another way we could do it is to just skip adding the action if the date isn't valid, and throw a warning. 
I don't love that because we're a little less inclined to notice and fix warnings, but it would keep the pipeline cranking along.

jessemortenson:
IMO downstream consumers should know that the occasional bad value is going to come through, because this project scrapes source websites. Does OS take on the responsibility for assuring that data is valid in terms of real world truth, or valid in terms of representing what is available on the source website? 

Totally valid to prefer a stoppage in data flow to force a fix. I just prefer the other side of that tradeoff :)

jamesturk:
Spent some time thinking through this and I think I've come around on 0202

Here's my justification:

- When a year is just wrong (say they wrote 1965) we certainly don't correct it.
- When a year is wrong enough that we can't import it (say 19100 for 2000) but in a predictable way we have to do *something* to import the data (or leave it off)- but we can figure out intent pretty easily, so we should.
- Omitting the data altogether is an option, but a last resort when we just can't import it at all, it makes us unlikely to get it fixed, be aware of it, etc. Whereas data showing up with bad dates is at least a bit more understandable.  Most users will be able to figure out from context that it isn't the right year.

I initially thought that 202- fell into the second camp, we could figure out intent, but as Tim correctly pointed out, that's only because we're in 2020.  If it were 2023 we'd need complicated logic to figure out if the year was likely 2022 or 2023 or other options. 

I think that means it falls into the same camp as when states (looking at you DC) publish data for 2200 instead of 2020.  A clearly incorrect year that we won't fix.

PR #3425, opened 2020-08-11
Issue 100: bounce out of versions if node is empty

dlamoureaux:
Fix for https://github.com/openstates/issues/issues/100. The versions node was empty in the xml causing the scraper to fail when trying to parse the non-existant versionDescription child node. Updated to bounce out if the versions node is empty.


jamesturk:
Thank you for this fix!

PR #3431, opened 2020-08-24
WIP: VT: Add 2020 first special

showerst:
Don't merge until data is posted so we can test


showerst:
Note: they ended up just continuing the regular session here instead of adding a new special to the site.

PR #3433, opened 2020-08-24
Fix `bill_list_url` query string in NV bill scrape

serixscorpio:
Fixes openstates/issues#103 

## Issue & Investigation

Recent NV bill are duplicated.

I think the `bill_list_url` is intended to be a multi-line string, but it's effectively truncated due to missing multi-line string syntax.
This likely led to multiple passes of querying the same bill and resulted in certain NV bills being duplicated.

## Fix

Use parenthesis to ensure multiple lines of the url's query string
are accounted for.  Use f-string for the entirety of the multi-line string for consistency.
Note: as mentioned in https://github.com/openstates/issues/issues/103#issuecomment-672020483, some manual bill removal may be needed to fix preexisting data.

### DIFF
--- a/scrapers/nv/bills.py
+++ b/scrapers/nv/bills.py
@@ -89,8 +89,11 @@ class NVBillScraper(Scraper, LXMLMixin):
         }
 
         for doc_type in doc_types[chamber]:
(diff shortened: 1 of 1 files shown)


jamesturk:
nice catch!

PR #3434, opened 2020-08-28
MA: add optional page_limit to allow a limited-latest bill scrape

jessemortenson:
Hoping to provide a workaround/mitigation for https://github.com/openstates/issues/issues/107

My goals are:

- have a smaller daily task that would be *very likely* to complete and yield *most* bill data updates (ensure some data is updated frequently)
- could keep the existing full run (maybe less frequently) to *hopefully* yield *all* bill data updates

So in addition to merging this code, y'all would need to add an MA-scrape-lite job that did something like:

```
docker-compose run --rm scrape ma bills sort=latest page_limit=20 --scrape
```

## other options I considered

- I can imagine a structure wherein bill scrape yields bills and jobs. In MA, there are several related entities that need to be scraped in association with each bill. Instead of making these additional scrape requests as part of the bill-scrape procedure, the MA bill scraper would instead yield jobs. Each job would represent the context necessary to perform the additional, bill-related scrape/transform. Another worker task slowly picks up these jobs from a queue or database table and processes them atomically, yielding updates to the existing bills that cause add the related entities (actions, sponsors, votes). This would introduce some significant infrastructure changes, I imagine. 
-  Apply bill-related scrapes only if it looks like a Bill has been updated since (last update). Use some proxy available on the bill page (perhaps actions text) to decide whether it's worth performing the related scrapes or not. Looking at the MA pages, I'm guessing this method would not be reliable enough.

### DIFF
--- a/scrapers/ma/bills.py
+++ b/scrapers/ma/bills.py
@@ -54,9 +54,13 @@ class MABillScraper(Scraper):
     # os-update ma bills --scrape sort=latest
     # bill_no can be set to a specific bill (no spaces) to scrape only one
     # os-update ma bills --scrape bill_no=H2
-    def scrape(self, chamber=None, session=None, bill_no=None, sort=None):
+    # page_limit can be set to stop scraping after a certain number of pages (for each chamber)
+    def scrape(self, chamber=None, session=None, bill_no=None, sort=None, page_limit=None):
         if not session:
             session = self.latest_session()
+
+        if page_limit:
+            page_limit = int(page_limit)
 
         if bill_no:
             single_bill_chamber = False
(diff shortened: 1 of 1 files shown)


showerst:
@jessemortenson this makes sense to me -- i'll ask james about the linter error because i'm seeing it too before i merge.

jessemortenson:
Thanks for the merge! Are you planning to add the `docker-compose run --rm scrape ma bills sort=latest page_limit=20 --scrape` task?

jamesturk:
Yes, going to update the normal task to do that, and then play with adding
a longer one too that runs less often.

PR #3440, opened 2020-09-16
Add 2 sessions to MN to fix breakage

mollycode:
Add support for the 5th special session and 92nd regular session as they are both now showing up on the leg website and causing the scraper to fail.

I also updated previous special sessions to indicate they are special not primary, I'm not sure if going back and changing those will break anything so let me know if it's better to just leave them as is.


showerst:
Thanks @mollycode ! 

FYI we generally wait to merge until there's live data for the session, so we can do a quick test run and check for errors. 

I'll merge this once the 5th special starts filing.

mollycode:
Ah gotcha, ok thanks!

mollycode:
Actually would it make more sense to move those to ignored for now then? That way the scraper doesnt have to keep failing?
https://bobsled.openstates.org/run/04547f37b72b46f8a132125c135e331b

showerst:
That can kinda go either way -- 

- If the state is still in action in the 'current' session then we definitely ignore new ones, so that we can keep getting current session updated data.
- If it's a far future session in a state that doesn't do prefiling, like 2021, we generally ignore.

But when a session is just about to start it's nice to keep the scraper failing as opposed to ignored, because the error messages and failure logs provide a reminder to keep an eye on it every day, and the API isn't missing any data by not scraping, since the state is out.

I see they're holding subcommittee meetings, but they're not updating any bills are they?

If the fourth special is not going to update any more bills i'd go with "ignore 2021, make 2020s5 the current and let it fail for a few days, because the API isn't missing any updates anyway". 

Otherwise we should just add them both to the ignore list. In cases like that I like to just comment out entries for the the upcoming ones to save work later when they come live later.

showerst:
Just in case it's non-obvious, I like to wait to merge new sessions (particularly specials) until real bills are posted because sometimes you get some goofy unexpected errors. 

We've seen states post old bills from the wrong session, dummy test bills, actions from previous sessions, etc. etc, and it's a lot easier to not scrape than to untangle bad data.

mollycode:
Ok, that makes sense, thanks for the background! I dont think session 4 is updating any more bills so i'll go with the 2020s5 session as current and ignore 2021.

mollycode:
Just a heads up that the 5th special session starts on Monday!

showerst:
Thanks! I've left myself a note to check monday/tuesday and once there's data in we'll get this merged.

PR #3449, opened 2020-10-13
Catch FTP error in VA load_members

mollycode:
FTP flakiness was causing the error below and leading to semi-frequent failed [VA scraper runs](https://bobsled.openstates.org/task/VA-scrape). This PR just catches that error so the scraper can continue.

`scrapelib.FTPError: error while retrieving ftp://**SCRAPERS/VIRGINIA_FTP_USER**:**SCRAPERS/VIRGINIA_FTP_PASSWORD**@legis.virginia.gov/fromdlas/csv202/Members.csv
`

### DIFF
--- a/scrapers/va/csv_bills.py
+++ b/scrapers/va/csv_bills.py
@@ -3,6 +3,7 @@ import csv
 import re
 import pytz
 import datetime
+import scrapelib
 from openstates.scrape import Scraper, Bill, VoteEvent
 from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


jamesturk:
if self._members isn't set doesn't that cause problems in load_votes?  maybe we could retry in this case if it seems intermittent?

mollycode:
Its failing after 3 retries right now (with the 10, 20, 40s waits), are you thinking we do a longer retry? And it might cause issues in load_votes, is there a way to test the full run without having a username and password? I wasn't able to figure that out. Thanks for your help!

jamesturk:
I see that now, hadn't taken a close look at the failure logs.  I remember
that when we switched over to FTP it was looking like their FTP is down at
certain hours.  I believe we moved the scrape earlier to avoid those
mostly.

I don't think the scrape can continue without the FTP being up- so maybe
the solution is to find other times to scrape (it is fast enough we could
try multiple times per day).

As for the API key question, I'd recommend getting one for the team, but
also glad to discuss (on Slack) sharing one if you want to run a few times
to try something out.

mollycode:
Ok I'll work on getting a key for our team! I think shifting when we scrape or doing multiple retries seems like a good solution to me. I can close this PR, but should I file an issue for it/is that something I can update?

jamesturk:
just pushed a config change that'll try it a few times a day: https://github.com/openstates/task-definitions/commit/7047de601fa4be7cc8976d60ddc91402fbd3ebec

we can monitor that & revise it as needed.  if you'd like to open an issue that'd be great too

mollycode:
awesome, thanks! and will open an issue as well

PR #3467, opened 2020-11-30
ME: Add 130th session which starts Dec 2

jessemortenson:
Appears that no bills are available yet in session 130, so impossible to verify this is working. However, the bill advanced search interface does not seem to have changed. It does now list session 130:

http://www.mainelegislature.org/LawMakerWeb/advancedsearch.asp

### DIFF
--- a/scrapers/me/__init__.py
+++ b/scrapers/me/__init__.py
@@ -80,7 +80,7 @@ class Maine(State):
         {
             "_scraped_name": "130th Legislature",
             "identifier": "130",
-            "name": "130th Legislature (2020-2021)",
+            "name": "130th Legislature (2021-2022)",
             "start_date": "2020-12-02",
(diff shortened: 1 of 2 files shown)


showerst:
:+1:  i'll merge after bills populate, if you catch them populated before i do just @ me.

showerst:
@jessemortenson rather than me PR'ing your PR, can you change line 78 of bills.py from:

```python
if bill_id in BLACKLISTED_BILL_IDS[session]:
```

```python
if session in BLACKLISTED_BILL_IDS and bill_id in BLACKLISTED_BILL_IDS[session]:
```

If we do that I think we'll be good to go.

jessemortenson:
Got it - tested locally and looks good to me. (started making fixes before I saw your comment).

PR #3471, opened 2020-12-02
NE: Add events scraper

showerst:


showerst:
This needs a bit more testing, i'll merge once it's finished.

PR #3472, opened 2020-12-03
MN: ignore broken SSL/TLS verification, account for 7th special session

jessemortenson:
SSL verification is now failing, so turning that off in the get_session_list() call. BTW: is there a way to add intermediate certs to the scraping stack? I have the missing certs (we add these in our nodejs stack).

Also accounts for the 7th special session, which is now listed on the website. That isn't announced yet, but is almost certain to occur in the next 5-10 days (it's a constitutional requirement for our declaration of emergency by the gov)


jamesturk:
figuring out a workflow for adding intermediate certs could be a new issue if you want to file it, I hadn't considered that option previously

PR #3476, opened 2020-12-07
Don't change Massachusetts House district names

bhrutledge:
Fixes https://github.com/openstates/issues/issues/170.

This may be premature and and presumptuous, but it seemed liked a quick thing to do. I haven't tested it locally; happy to go through the [setup](https://docs.openstates.org/en/latest/contributing/scrapers.html) if requested.

### DIFF
--- a/scrapers/ma/people.py
+++ b/scrapers/ma/people.py
@@ -3,46 +3,6 @@ import lxml.html
 from openstates.scrape import Person, Scraper
 
 
-def clean_district(district):
-    mappings = {
-        1: "First",
-        2: "Second",
-        3: "Third",
-        4: "Fourth",
-        5: "Fifth",
-        6: "Sixth",
-        7: "Seventh",
-        8: "Eighth",
-        9: "Ninth",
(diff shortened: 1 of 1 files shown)


jamesturk:
👍 i'll pull this in, the actual change will take us doing this in openstates/people

bhrutledge:
Thanks @jamesturk! Out of curiosity, can you point to where the parallel change needs to be made?

jamesturk:
In the data itself here:
https://github.com/openstates/people/tree/main/data/ma

Doing so will avoid merge issues/etc.  when we do update MA legislators

bhrutledge:
@jamesturk Gotcha. I see some instructions at https://docs.openstates.org/en/latest/contributing/people.html. Continuing with my curiosity, is this a regular/automated maintenance task? Does it warrant a separate issue and/or PR?

jamesturk:
This change will have to be done manually to the old legislators,
essentially a find and replace across those YAML files.

PR #3484, opened 2020-12-11
MT: Enable 2021 session as there is at least one bill

jessemortenson:
One bill is available on http://laws.leg.mt.gov/legprd/LAW0217W$BAIV.return_all_bills?P_SESS=20211


showerst:
Ran this and it looks good to me. @jessemortenson did james give you merge privs yet? If so feel free to merge, otherwise let me know and i'll hit the button.

PR #3487, opened 2020-12-15
MT: Ingest proposed bills (LCXXXX) aka unintroduced

jessemortenson:
Start ingesting LCXXXX bills in MT under the proposed bills classification.

https://github.com/openstates/issues/issues/193

One wrinkle in here is that some "unintroduced" bills are attributed to organizations, eg the "Economic Affairs Interim Committee". In that case I omit any sponsor, and arbitrarily assign chamber="lower". Is that horrible?

### DIFF
--- a/scrapers/mt/bills.py
+++ b/scrapers/mt/bills.py
@@ -245,7 +245,7 @@ class MTBillScraper(Scraper, LXMLMixin):
                 chamber = "upper"
             else:
                 # a true fallback: some sponsors are organizations eg "Economic Affairs Interim Committee"
-                chamber = "lower"
+                chamber = "legislature"
 
         bill = Bill(
             bill_id,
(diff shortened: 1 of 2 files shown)


jamesturk:
Typically we allow org sponsors as they are, it'll affect us when we get to measuring metrics on matched sponsors, but there are plenty of states that allow committees (and in some cases non-gov entities) to be listed as formal sponsor for bills.

scrapers/mt/bills.py
```diff
@@ -203,8 +228,24 @@ def parse_bill_status_page(self, url, page, session):
...
+        elif _bill_id[0] == "s":
+            chamber = "upper"
+        else:
+            # fall back to using the sponsor's chamber
+            # used for proposed bills aka unintroducd aka LC bills
+            if " HD " in list_sponsor:
+                chamber = "lower"
+            if " SD " in list_sponsor:
+                chamber = "upper"
+            else:
```
jamesturk, line 246: ah in reviewing the code itself so I could suggest an alternative, I think I missed part of your point on the issue itself.  If I read this right, you're using the sponsor to determine bill chamber, and these can't be determined cleanly?

jessemortenson: Yes, if an organization is the sponsor, and there does not appear to be a chamber assignment, should I instead default to "legislature" ?

jamesturk: yes, let's go with legislature if the chamber isn't known.  sponsor would be the committee 👍

jessemortenson: Cool, pushed changes - please take a peak and make sure I got it

jamesturk: LGTM

PR #3489, opened 2020-12-17
VA: 2021 session metadata [HOLD]

jamesturk:
Working to verify that the bills/etc. that are present here are actually carryover, I believe they are but want to confirm there aren't data issues.


jamesturk:
found a summary of the carryover and also new bills from a blog (https://bluevirginia.us/2020/11/75-of-the-best-new-bills-and-bills-carried-over-from-2020-for-the-2021-virginia-general-assembly-session) and the numbers match up

also checked a dozen and they were all either new prefiles or had carryover actions

PR #3497, opened 2021-01-04
OH: 2021 session and adjust SSL verification for some requests

jamesturk:
openstates/issues#216 and openstates/issues#161


jamesturk:
just saw @showerst note that he's working on this scraper, probably won't need this but leaving here as a reminder

showerst:
my branch of this is the same, so we can use this one. Sorry for the dupe effort, I should've just PR'd with a hold.

PR #3502, opened 2021-01-04
CT: FTP listing isn't working, switch metadata to use HTTP, add 2021

jamesturk:
waiting on actual 2021 data


jamesturk:
data is live but empty now, will verify before turning on job

PR #3506, opened 2021-01-04
NE: 2021 session

showerst:
Waiting on data, do not merge yet.


jamesturk:
checked this morning & session is up now, but no bills yet

showerst:
Bills look good, merging.

FYI if you check the logs you'll see a few errors about bill titles, that's happened before -- sometimes they post incomplete info to the website, and fill it in later.

PR #3507, opened 2021-01-05
IN: Version Name Check

NewAgeAirbender:
Some versions on the API are misnamed, so added a sanity check to make sure that the name matches the chamber of the version url and printed name. Should fix for most versions.


jamesturk:
this is a complex one, but based on my understanding makes sense-- extra thanks for the detailed comments, I imagine they'll be helpful in the future

PR #3509, opened 2021-01-05
MT: switch PDF links to HTTPS

jessemortenson:
MT server is now redirecting PDF urls from HTTP to HTTPS (302) so this seems more correct

Don't know if this causes duplicate versions the way that OS logic tries to merge in versions? (eg will it create a new version based on the URL being different?) If that causes a headache we could solve this downstream by parsing OS data.


jamesturk:
OS will update the versions here without any intervention 👍

PR #3510, opened 2021-01-05
Add federal scrapers

showerst:
US Federal Scrapers that output in Openstates format.

As previously discussed, adding these back here in case it's useful to anyone, and for license compliance purposes.

This data won't show up in the OS API, if you're interested in an API for this check out Propublica's congress api.


showerst:
@jessemortenson FYI.

jamesturk:
ah this is great, really glad to have this!

PR #3514, opened 2021-01-05
MN: enable 2021-2022 session scraping

jessemortenson:
MN 2021 bills are now appearing at https://www.revisor.mn.gov/bills/reference.php

running `docker-compose run --rm scrape mn bills --fastmode --scrape` worked locally for me with this change


showerst:
@jamesturk i ran and it's good so i'm approving, but just a ping to run the scrape.

PR #3521, opened 2021-01-12
SD: 2021 WIP

NewAgeAirbender:
Works for 2021 but run into a few issues: 

- HCR 6001 pulls as bill_id "HCR HCR6001" and I don't know why.
- Need to come up with a full action text like they have on the webpage that lists the status, chamber, committees, etc. but not sure the best way to go about doing that. Need to append api values into a string, but considering none of the 2021 bills have actions available yet, it's not as time sensitive. Was going to take a look at the site and see if I could find their logic for formulating that and replicate it on the scraper.

Still need to test for previous sessions to make sure actions and votes are accurate.


showerst:
@NewAgeAirbender is this something you're comfortable with us testing and merging in its current state, or should I hold off on merge until there are actions and votes to test?

NewAgeAirbender:
> @NewAgeAirbender is this something you're comfortable with us testing and merging in its current state, or should I hold off on merge until there are actions and votes to test?

Yup, can test and merge. James is also taking a look at it.

jamesturk:
taking a closer look at the HCR thing, but looking good to me and I'm ok merging soon

showerst:
@jamesturk just ran it and checked over some bills and it looks good to me.

PR #3523, opened 2021-01-13
CO: 2021 Session Metadata (WIP)

jamesturk:


jamesturk:
there is a test bill but we aren't getting it, there may be a site change-- but waiting until there's an actual bill to diagnose

jamesturk:
bill posted, scraper needs a small fix I think

jamesturk:
OK, there is a edge case where there aren't bills yet in the house.  Instead of writing a workaround, given that the only bill is a SJR organizational bill, I'm going to wait to push this until there's at least one house bill ideally (or more Senate bills if need be)

jamesturk:
looks like they're filing senate bills now, will just ignore missing chamber with a warning

PR #3533, opened 2021-01-16
AR: fix vote scraper

ColbyReed:
A previous update to the Arkansas Legislature website broke scraping of vote data for bills. I fixed the xpath logic and changed the parsing of names to instead parse legislature urls to allow the "Mr. Speaker" vote to be properly recorded.

This is related to this issue I posted a few months ago:
https://github.com/openstates/issues/issues/144

However, data from previous sessions will still be missing vote data as scrapes are dependent on:
ftp://www.arkleg.state.ar.us/SessionInformation/LegislativeMeasures.txt
But this file is cleared at the end of each session.

To collect older vote data there will still need to be a rewrite that would allow webpage only scraping.


jamesturk:
thanks for this!

PR #3538, opened 2021-01-21
SD: Enable action scrape

showerst:
Now that SD is posting actions, enable the scrape

https://github.com/openstates/issues/issues/238


showerst:
Note: don't merge this yet, still working out some failures.

PR #3540, opened 2021-01-21
SD: Actions & Votes

NewAgeAirbender:
Fixes issue with versions not having dates or documents, pulls in information for full_action to be accurate, and gets Votes working. Full_action gets transformed into the action description and has all the critical information in the order it looks like on the site, but the it might have an extra space or chamber/committee in the beginning just because the logic is funky on the SD site. Would rather have more info than not enough, right?


scrapers/sd/bills.py
```diff
@@ -163,66 +159,97 @@ def scrape_action(self, bill, url, chamber):
                     self.error("Unknown vote code: {}".format(action["Result"]))
                 atypes.append("%s%s" % (first, second))
 
+            if (
+                action_text == "Do Pass"
+                or action_text == "Tabled"
+                or "ShowCommitteeName" in action
+            ):
+                full_action.insert(0, f'{action["ActionCommittee"]["Name"]}')
```
showerst, line 167: Let's filter out "House of Representatives" and "Senate" from this, since the chamber is mentioned in the action anyway.

NewAgeAirbender: Switching `or "ShowCommitteeName" in action` to `or action["ShowCommitteeName"]` does this, I just mis-interpreted converting the site logic. Should be fixed now.

jamesturk: thanks for this fix, the rest looks good from here!

PR #3541, opened 2021-01-22
SD: Scrape hearings

showerst:


showerst:
Hold off merging this -- i think the start times on hearings are wrong, so I'm thinking about the right way to grab them.

PR #3548, opened 2021-01-25
AR: fix people scraper xpaths

ColbyReed:
Fixed the xpaths for people scraping. Should be able to correctly grab phone numbers, emails, and occupations fields.


ColbyReed:
Seems the district xpath was already fixed. This PR should only change the xpaths mentioned in the original comment.

PR #3556, opened 2021-01-26
IA Prefiles, openstates/issues#252

jamesturk:
openstates/issues#252


jamesturk:
the scrape looks OK, we had a note about duplicate bills but it looks like this might be the only way to get both

whether or not we need a flag might require some reconsideration in a future session

@showerst do you know of any issues with continuing to collect prefiles & regular bills side by side here?

jamesturk:
going to merge this for now, we can revert/change if needed

PR #3560, opened 2021-01-27
Add federal scraping 

rmcarthur:
Reinstates federal scraping and updates it to the tied Jurisdiction IDs from this PR: https://github.com/openstates/openstates-core/pull/19


rmcarthur:
This is a WIP, but the goal is to get federal bill ingested into the OS database in an acceptable format.

jamesturk:
Thanks for this.

I'll need to think through what changes OS will require to accomodate this, for instance we don't want the US showing up on openstates.org for now, but I can see the value in having it in the API now that the work is there.

PR #3566, opened 2021-01-28
AR: add scraping of amendments

ColbyReed:
Added scraping for amendments on bills and removed leftover print statements.


jamesturk:
thanks for all this great work on Arkansas!

PR #3613, opened 2021-02-24
VA: fix source URLs in weird 2021 special+regular session

jessemortenson:
Attempts to fix https://github.com/openstates/issues/issues/328

However I don't have the Virginia FTP credentials - could somebody else test or DM me credentials?


showerst:
I'll test shortly -- FYI if you call them at the number here https://lis.virginia.gov/SiteInformation/ftp.html and ask they'll set you up with creds.

showerst:
This looks good, but since it changes the URLs of all the versions, @jamesturk would you mind checking it out -- not sure if this should be done in tandem with a DB url string replacement to avoid dupes.

jamesturk:
sorry for leaving this hanging, thought I hit merge couple of days ago.  no dupes will be created, versions will update

PR #3637, opened 2021-03-24
WIP: Enable US scrapers

jamesturk:


jamesturk:
Question for you @showerst : part of integrating this with the openstates-core code requires renaming the module from us to usa (to avoid a conflict with the python-us package we depend upon) I wanted to flag that change and make sure it wouldn't cause problems for you.

showerst:
Thanks for the flag, that shouldn't be a problem @jamesturk -- I can update my jobs when we merge all of this. The main thing I need to keep an eye on is if the OCD identifier for the emitted bills changes. Still not a problem, i'll just need to update some things on my end.

jamesturk:
The Jurisdiction identifier?  If so, that is ``ocd-jurisdiction/country:us/government`` with this.

PR #3639, opened 2021-03-24
MI: fix some actor classification errors

showerst:
In MI, we're classifying the action chambers (actors) via a reference to the chamber's journal page, which appears next to the action on the website. (HJ vs SJ).

Lately, they've been posting actions with 'Expected Soon' as the journal cite, then adding it later. This causes us to incorrectly classify the action due to an overly loose else clause.

**edit** looks like we can fall back to using the casing, because house lowercases and senate uppercases. Yuck.

@jamesturk until they post the reference, we may not know the chamber. Will it cause the OS ingestion any problems if we save those actors as 'legislature', then it gets updated on subsequent scrapes?

Here's a screenshot of what i'm talking about, to help clarify:

https://www.dropbox.com/s/itzv7ru8s3ku3kr/Screenshot_2021-03-24-17%3A06.png?dl=0

### DIFF
--- a/scrapers/mi/bills.py
+++ b/scrapers/mi/bills.py
@@ -137,7 +137,15 @@ class MIBillScraper(Scraper):
                     )
                     continue
             # instead of trusting upper/lower case, use journal for actor
-            actor = "upper" if "SJ" in journal else "lower"
+            # Journal entries are often posted with 'Expected Soon' as the cite,
+            # then changed to the journal entry.
+            if "SJ" in journal.upper():
+                actor = "upper"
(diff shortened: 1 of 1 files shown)


jamesturk:
is the question about 'legislature' still relevant?  I need to investigate if so, if case is reliable I'm OK to merge

showerst:
I think so, to try to phrase it better --

If we don't know the chamber of an action, then that data might come in later, should it start with 'legislature' ?

jamesturk:
Yes, that works and shouldn’t cause any issues.

PR #3642, opened 2021-03-29
AR: fixed vote event scraper after website update removed screenreader class

ColbyReed:
Seems a website update removed the screenreader class for most bill pages, which the page scraper relies on for searching for vote links. I changed the xpaths to point to the divs in the grid role instead.


showerst:
Thanks @ColbyReed -- testing this now.

PR #3646, opened 2021-03-30
AR: Scrape date and time of day for bill actions

ColbyReed:
I noticed that bill actions were only scraping dates, making it difficult to chronologically sort actions that occurred the same day. This change allows time of day to be scraped as well.

### DIFF
--- a/scrapers/ar/bills.py
+++ b/scrapers/ar/bills.py
@@ -110,7 +110,7 @@ class ARBillScraper(Scraper):
             date = TIMEZONE.localize(
                 datetime.datetime.strptime(row[5], "%Y-%m-%d %H:%M:%S.%f")
             )
-            date = "{:%Y-%m-%d}".format(date)
+
             action = row[6]
 
             action_type = []


ColbyReed:
I just noticed the timezone conversion might be giving incorrect times? All times recorded for actions and vote events are 6 hours ahead. Is this intentional?

jamesturk:
Yeah, if things are showing up strangely in the API/etc. let me know and I’ll take a closer look.

ColbyReed:
Yes the emitted data. Central is UTC-6 so that does make sense. Just wanted to verify.

ColbyReed:
Just as an example this is how its formated in the .json for HB 1263

"date": "2021-01-25T20:21:08.050000+00:00"

showerst:
Cool, that's good then. The +00:00 denotes no timezone offset (so UTC). FYI we convert all dates to UTC to make things more sane when comparing across jurisdictions or searching. 

On openstates.org we then convert it back to local time for display.

I'll test/merge your actual PR later today or tomorrow if nobody beats me to it.

showerst:
@jamesturk @jessemortenson this code looks good, but just wanted to flag that a merge would invalidate the cache for all AR bills. Is that an issue here?

jamesturk:
no worries, OK to merge if it looks good, thanks!

PR #3659, opened 2021-04-14
IA: Enable vote scraper with recent session ID

ColbyReed:


jamesturk:
also enabled this and TX in task-definitions

PR #3660, opened 2021-04-14
MN: add vote urls for 2020 special sessions and 2021 regular session

ColbyReed:
Looks like Minnesota didn't have vote urls for the 2020 Second Special Sesssion to the 2021-2022 Regular session. I added them, but the special sessions between 2 and 7 will need to be scraped as well.


showerst:
@jamesturk just leaving a note that those will need one-off scrapes.

PR #3661, opened 2021-04-14
TX: Bills: Clean up action classifier

drvander:
Addresses [openstates/issues/#375](https://github.com/openstates/issues/issues/375)


showerst:
Thanks @drvander !

PR #3673, opened 2021-05-14
CO: Skip votes without counts

showerst:


showerst:
Should fix https://bobsled.openstates.org/run/4c3b1136c3cb48b2ae299cc80944fa8c

PR #3675, opened 2021-05-18
CA: Adding Bill Analyses to 

AliAbstract:
Unfortunately, like California PDFs, the link to the bill analyses doesn't actually grab the bill analyses, so this URL is mostly as a data store. I'm using it in a fork of the project to actually download the CA bill PDFs and analyses PDFs, but I'm not sure where to put that code within the openstates ecosystem.


showerst:
Even providing the proper link to the page is still helpful, thanks! I'll give this a test run later today and merge or comment.

PR #3696, opened 2021-06-21
Wa scrape

morden35:
Wrote a new scraper to convert old wa/people.py to use spatula. Used some pre-existing XPath selectors as well as some new/modified CSS selectors. Works for both Senators (SenList) and Representatives (RepList). @jamesturk to review.

### DIFF
--- a/scrapers_next/wa/people.py
+++ b/scrapers_next/wa/people.py
@@ -11,11 +11,8 @@ class PartyAugmentation(HtmlListPage):
     """
 
     source = URL("https://app.leg.wa.gov/memberemail/Default.aspx")
-    # selector = CSS("#membertable tbody tr", num_items=147)
-    # dist_pos_dict = {}
 
     def find_rows(self):
-        # the first table on the page that has a bunch of rows
         # print(CSS("html body").match(self.root)[0].getchildren()[0].text_content())
         for table in CSS("#membertable").match(self.root):
             rows = CSS("tbody tr").match(table)
@@ -23,7 +20,6 @@ class PartyAugmentation(HtmlListPage):
                 return rows
 
     def process_page(self):
-        # Retrieve the member's position from the email link.
         # We need it to find the member's email address.
         # These positions are enough to discriminate the chamber too (0 = upper, 1,2 = lower)
         mapping = {}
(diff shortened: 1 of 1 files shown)


jamesturk:
I'll take this one!

scrapers_next/wa/people.py
```diff
@@ -0,0 +1,167 @@
+import re
+from spatula import HtmlListPage, CSS, XPath, URL
+from ..common.people import ScrapePerson
+
+
+class PartyAugmentation(HtmlListPage):
```
jamesturk, line 6: let's rename this to something appropriate for Washington's usage here

scrapers_next/wa/people.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    WA Email addresses are listed on a separate page.
+    """
+
+    source = URL("https://app.leg.wa.gov/memberemail/Default.aspx")
+
+    def find_rows(self):
```
jamesturk, line 13: might be better expressed as        

 return CSS("#membertable tbody tr").match(self.root, num_items=147)

scrapers_next/wa/people.py
```diff
@@ -0,0 +1,167 @@
...
+        (name, party) = re.search(
+            r"^(?:Senator|Representative)\s(.+)\s\(([RD])\)$", title_name_party
+        ).groups()
+        if party == "R":
+            party = "Republican"
+        elif party == "D":
+            party = "Democratic"
+
+        (district_name, _district_name) = XPath(
+            ".//a[contains(text()," ' " Legislative District")]/text()'
```
jamesturk, line 56: this is a bit tough to read, and can be cleaned up by using 
`
            './/a[contains(text(), "Legislative District")]/text()'
`

instead of the two split strings

scrapers_next/wa/people.py
```diff
@@ -0,0 +1,167 @@
...
+        p.capitol_office.address = capitol_address
+        p.capitol_office.voice = capitol_phone
+
+        if len(_has_district_office) > 0:
+            p.district_office.address = district_address
+            if district_phone:
+                p.district_office.voice = district_phone
+
+        p.add_link(XPath('.//a[contains(text(), "Home Page")]/@href').match(item)[0])
+        p.add_source(str(self.source))
```
jamesturk, line 134: nit: let's use self.source.url for these

str(source) can return something that isn't a URL if the __str__ method of Source ever changes, `url` is guaranteed to be a valid URL

scrapers_next/wa/people.py
```diff
@@ -0,0 +1,167 @@
...
+            email = tds[1].text_content().strip()
+            dist = tds[2].text_content().strip()
+            position = tds[3].text_content().strip()
+            party = tds[4].text_content().strip()
+            mapping[name] = (email, party, dist, position)
+        return mapping
+
+
+class LegList(HtmlListPage):
+    dependencies = {"party_mapping": PartyAugmentation()}
```
jamesturk, line 42: party_mapping should probably be renamed to something like email_mapping here

jamesturk (changes requested):
on the whole this looks great!

 a few small stylistic fixes, but a quick skim of the scraper output looks good to me

PR #3699, opened 2021-06-24
restore linting

jamesturk:
black & flake8 linting accidentally got disabled a couple of months ago

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,9 +2,9 @@ name: Lint Python
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
   lint:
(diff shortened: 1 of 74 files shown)


jamesturk:
this is kind of a big one since it had been disabled for a while, it changes a bunch of event.py files too so I wanted to make sure this wouldn't create any major merge conflicts with upcoming work @showerst

showerst:
Thanks for the h/u -- my bad on the linting, i'll try to get my pre-commit pipeline configured to the same config that OS uses so that we stop needing these larger passes.

jamesturk:
sounds good, my bad on not noticing it was disabled

if you use `pre-commit install` most things should just work, let me know if it gives you any trouble

jamesturk:
wanted to confirm, ok to merge or better to hold off?

showerst:
Cool with me to merge, thanks!

PR #3709, opened 2021-06-30
Ri spatula scrape

morden35:
This was a tough one. Converted old RI people scraper to spatula framework. Note that for each person, there is an html source and an excel file source. Also note that images are coming from another web page (bio page).


jamesturk:
this looks great!

PR #3717, opened 2021-07-06
AZ: resume scraping regular session to catch missing HR 2017

jessemortenson:
Re-orders sessions so regular session will be scraped again. [HR 2017](https://apps.azleg.gov/BillStatus/BillOverview/76200?SessionId=123) is missing from [Open States data](https://openstates.org/search/?query=HR+2017&state=az).

(@jamesturk is this still the preferred way to handle this - vs. changing the task definition to specify the regular session?)

### DIFF
--- a/scrapers/az/__init__.py
+++ b/scrapers/az/__init__.py
@@ -250,20 +250,20 @@ class Arizona(State):
             "end_date": "2020-05-27",
         },
         {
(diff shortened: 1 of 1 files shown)


jamesturk:
👍 and yeah I think generally this is preferred, but admittedly a little inconsistent there

PR #3718, opened 2021-07-06
MN committee spatula scrape

judgejudes:
- scraped both the Senate and House committees
- added extra information like meeting schedules, and rooms, along with extra links
- I was a bit unsure of capitalization of positions, so please let me know if there should be any changes with those!


jamesturk:
👏 looks great, MN committees are now added to the people repo too!

PR #3719, opened 2021-07-07
Ga committee spatula

morden35:
Scraping Georgia House and Senate Committees using spatula framework. Note that Georgia is using an api that requires authorization. James created the get_key() and get_token() functions to generate the authorization key for the api. Also note that this does not scrape GA special/study committee pages. @jamesturk


scrapers_next/ga/committees.py
```diff
@@ -0,0 +1,107 @@
...
+
+        com = ScrapeCommittee(
+            name=item["name"],
+            parent=chamber,
+        )
+
+        com.add_source(
+            self.source.url, note="Initial list page (requires authorization token)"
+        )
+        com.add_link(
```
jamesturk, line 100: let's drop this link

scrapers_next/ga/committees.py
```diff
@@ -0,0 +1,107 @@
...
+    return "Bearer " + requests.get(token_url).json()
+
+
+class CommitteeDetail(JsonPage):
+    def process_page(self):
+        com = self.input
+        com.add_source(
+            self.source.url, note="Detail page (requires authorization token)"
+        )
+        com.add_link(self.source.url, note="Detail page (requires authorization token)")
```
jamesturk, line 44: I think for this we could build the human-friendly link that looks like

https://www.legis.ga.gov/committees/house/90

I think the necessary data is in the response (90 is the ID)

jamesturk (changes requested):
one small change for this, I'd like to get the link URLs to be something that a user could click in the browser

everything else looks great!

scrapers_next/ga/committees.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if com.parent == "upper":
+            link_chamber = "senate"
+        elif com.parent == "lower":
+            link_chamber = "house"
+
+        source_url_list = self.source.url.split("/")
+        item_id = source_url_list[-2]
+
+        link = URL(f"https://www.legis.ga.gov/committees/{link_chamber}/{item_id}")
```
jamesturk, line 53: nit: I'd drop the URL from this, and just use the string directly since we're not requesting it

jamesturk: change looks good, but no real advantage to URL here since we aren't requesting the source page

PR #3721, opened 2021-07-07
TX committees scraper

judgejudes:
- note/potential issue: accented characters don't show up correctly in the data
- other than that, i tried my best to hardcode the fields as little as possible!

### DIFF
--- a/scrapers_next/tx/committees.py
+++ b/scrapers_next/tx/committees.py
@@ -24,7 +24,10 @@ class CommitteeDetail(HtmlPage):
                             label = label[:-1]
                         if label in positions:
                             role = label
+                        elif label == "members":
+                            role = "member"
                         else:
+                            # self.warn(f"unknown role {label}")
                             role = "member"
                     name = person.text_content().split(":")[1]
                 except SelectorError:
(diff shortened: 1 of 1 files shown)


scrapers_next/tx/committees.py
```diff
@@ -0,0 +1,103 @@
...
+            if name:
+                try:
+                    label_text = XPath("./td[1]//text()").match(person)
+                    for label in label_text:
+                        label = label.lower()
+                        if label.endswith(":"):
+                            label = label[:-1]
+                        if label in positions:
+                            role = label
+                        else:
```
jamesturk, line 27: I'd maybe use a `self.warn(f"unknown role {label}")` or similar here so we can know what roles are skipped by looking at the output

scrapers_next/tx/committees.py
```diff
@@ -0,0 +1,103 @@
...
+        table = XPath("//div[@id='content']/table[1]/tr/td//text()").match(self.root)
+        table = [
+            thing.replace("\r", "")
+            .replace("\n", "")
+            .replace("\t", "")
+            .replace(": ", "")
+            for thing in table
+        ]
+
+        fields = table[1::2]
```
jamesturk, line 49: since ::2 is an unusual construction some might not be familiar with right away, could you add comment to explain why you're skipping every other?

scrapers_next/tx/committees.py
```diff
@@ -0,0 +1,103 @@
...
+        extra = table[2::2]
+
+        for i in range(5):
```
jamesturk, line 52: a good rule of thumb is to always explain "magic numbers" like this in a comment or variable name, why is it 5?

jamesturk (changes requested):
looking very good, a couple of small nits (mainly comments)

scrapers_next/tx/committees.py
judgejudes: adjusted it a bit to only raise the warning when the label is not "members", and to set the role to "members" after sending the warning (just to be extra safe!)

scrapers_next/tx/committees.py
```diff
@@ -0,0 +1,105 @@
...
+                    for label in label_text:
+                        label = label.lower()
+                        if label.endswith(":"):
+                            label = label[:-1]
+                        if label in positions:
+                            role = label
+                        elif label == "members":
+                            role = "member"
+                        else:
+                            # self.warn(f"unknown role {label}")
```
jamesturk, line 30: looks good but this got commented out

jamesturk (commented):
one last thing, then all good!

scrapers_next/tx/committees.py
judgejudes: ah sorry about that!

PR #3732, opened 2021-07-14
Ca committee scrape

morden35:
Converting ca committee scraper over to spatula. Currently have 2 files. committees.py is organized by committee type. committees_test.py is organized by page type. committees_test.py should work for all senate committees at the time of this pr.


scrapers_next/ca/committees_test.py
```diff
@@ -0,0 +1,335 @@
...
+            return Type_Two(self.input, source=self.source)
+        elif page_type == "type_three":
+            # return None
+            return Type_Three(self.input, source=self.source)
+        else:
+            # return None
+            return Type_Four(self.input, source=self.source)
+
+
+class Type_One(HtmlPage):
```
jamesturk, line 41: nit: TypeOne is the preferred way to CamelCase class names :)

scrapers_next/ca/committees_test.py
```diff
@@ -0,0 +1,335 @@
...
+class AssemblyCommitteeList(HtmlListPage):
+    source = URL("https://www.assembly.ca.gov/committees")
+
+    selector = CSS("div .block.block-views ul li", num_items=98)
+
+    def process_item(self, item):
+        comm_name = CSS("a").match_one(item).text_content()
+        comm_url = CSS("a").match_one(item).get("href")
+
+        if (
```
jamesturk, line 306: I wonder if there is a better way to do this, to make it more readable

if not I'd at least add a comment sort of documenting the page structure around this that made the many parent calls required

scrapers_next/ca/committees_test.py
```diff
@@ -0,0 +1,243 @@
+from spatula import HtmlListPage, URL, XPath, CSS, HtmlPage, SelectorError
+from openstates.models import ScrapeCommittee
+import re
+
+
+class ChooseType(HtmlPage):
+    def process_page(self):
+
+        xpaths = {
```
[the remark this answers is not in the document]
jamesturk: since the others only differ on process_page but get the exact same input, I'm debating whether it makes more sense to do this by dispatching to process_page_type1, process_page_type2 functions vs. different classes, but I'll leave that up to you & let you decide which feels more natural to you while working on it.

jamesturk (commented):
added some initial thoughts, glad to chat through these later

scrapers_next/ca/committees_test.py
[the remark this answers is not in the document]
morden35: I tried to turn the classes into functions but could not get it to work properly (would call the function within the dictionary)

jamesturk:
wanted to check if this was ready to review again, if so could you move the actual scrapers into committees.py (& drop the _test file 👍 )

morden35:
@jamesturk yes! ready for review

scrapers_next/ca/committees.py
```diff
@@ -0,0 +1,343 @@
+from spatula import HtmlListPage, URL, XPath, CSS, HtmlPage, SelectorError
+from openstates.models import ScrapeCommittee
+import re
+
+
+class ChooseType(HtmlPage):
+    def process_page(self):
+
+        # this link was being considered type one, but should be type four
+        if self.source.url == "https://ajed.assembly.ca.gov":
```
jamesturk, line 10: OK I've been thinking about this and I think I have an alternate solution that I'd recommend that'll be some small code refactoring but fortunately no changes to the actual functions below

instead of TypeOne through TypeFour being their own classes, could we go with

def process_page_type1(root, committee, source, ...):
      # existing code from TypeOne.process_item with `self` usage replaced with parameters above
      return com 

(and so on)

and then instead of returning TypeOne/TypeTwo/TypeThree/etc. you could return the result of these functions

if this doesn't make sense, we can talk it through, as it is a bit unorthodox, but will reduce requests by 50%

jamesturk (commented):
this looks great overall!  thanks so much for diving in and figuring out these varied pages and documenting so much!

scrapers_next/ca/committees.py
morden35: I'm not sure I understand this. Could we meet at some point and talk this through?

morden35: I think I was able to figure this out. Let me know what you think. Only needed to pass source for process_page_type4()

scrapers_next/ca/committees.py
```diff
@@ -0,0 +1,356 @@
...
+            return process_page_type4(self.root, self.input, self.source)
+
+        if process_page_type1(self.root, self.input):
```
jamesturk, line 13: if data := function():
   return data
elif data := function2():
  return data

jamesturk:
👍 looking great!

PR #3738, opened 2021-07-19
WIP LA: add 2021 veto session

showerst:
Don't merge until there are bills to check.


showerst:
They didn't end up posting any bills here, they put the updates in the regular.

PR #3740, opened 2021-07-21
In committees clean

morden35:
Having trouble accessing the leg detail pages via api (getting 500 response). See lines 45-66 @jamesturk


jamesturk:
OK, I played with this on & off most of the day, different timeouts, retries, etc. and it is incredibly flaky.  I think their API just isn't suitable for this purpose, not sure why it is so much worse than their bill API -- but I think we've spent enough time trying to make it work.  Let's go with the approach you found initially, sorry for the diversion!

PR #3743, opened 2021-07-26
VA: Ignore 2021 s2 for now

showerst:


showerst:
I'll revisit this when they start next week -- last time they did a weird thing where they posted across sessions.

PR #3745, opened 2021-07-27
Il spatula people

morden35:
Re-writing the IL people scraper to use spatula framework. Currently scraping 59 Senators and 118 Representatives. Skipping over "Former Senate Members". process_addrs() handles pages with varying number of capitol/district address lines.

10 senators and 5 representatives have a link to a new page with "Additional District Addresses". I tried to write a class to capture these additional addresses, but could not get it to work. 
example: https://ilga.gov/house/Rep.asp?GA=102&MemberID=2879

@jamesturk

### DIFF
--- a/scrapers_next/il/people.py
+++ b/scrapers_next/il/people.py
@@ -3,6 +3,21 @@ from openstates.models import ScrapePerson
 import re
 
 
+class AdditionalAddresses(HtmlPage):
+    # print("we get here")
+    def process_page(self):
+        p = self.input
+        additional_office = (
+            CSS("td.notranslate.member").match_one(self.root).text_content()
+        )
+        additional_office, additional_phone = additional_office.split("(")
+        # print(additional_office)
+        # print(additional_phone)
+
+        p.additional_offices = additional_office
+        p.extras["extra phone"] = additional_phone
+
+
 class LegDetail(HtmlPage):
     def process_page(self):
         p = self.input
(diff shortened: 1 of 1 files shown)


scrapers_next/il/people.py
```diff
@@ -0,0 +1,206 @@
...
+
+class AdditionalAddresses(HtmlPage):
+    # print("we get here")
+    def process_page(self):
+        p = self.input
+        additional_office = (
+            CSS("td.notranslate.member").match_one(self.root).text_content()
+        )
+        additional_office, additional_phone = additional_office.split("(")
+        # print(additional_office)
```
jamesturk, line 14: let's remove these

scrapers_next/il/people.py
```diff
@@ -0,0 +1,206 @@
...
+            if (
+                line == "Springfield Office:"
+                or line == "District Office:"
+                or line == ""
+                or line == "Additional District Addresses"
+                or line.startswith("Senator")
+                or line.startswith("Representative")
+                or line.startswith("Years served:")
+            ):
+                line_number += 1
```
jamesturk, line 96: does line_number get incremented no matter what?  if so maybe we do it at the top instead of each if conditional needing one 

also check out `enumerate`: https://realpython.com/python-enumerate/ potentially

scrapers_next/il/people.py
```diff
@@ -0,0 +1,206 @@
...
+            or CSS("td").match(item)[0].get("class") == "header"
+        ):
+            self.skip()
+
+        first_link = CSS("td a").match(item)[0]
+        name = first_link.text_content()
+        detail_link = first_link.get("href")
+
+        # these are "Former Senate Members"
+        former_members = [
```
jamesturk, line 169: is there a way to detect these from the page?  the hard coded list will get out of date quickly

jamesturk (changes requested):
some small changes

PR #3751, opened 2021-08-03
AL people spatula scrape

judgejudes:
I parsed through the table full of information in a pretty tedious way just because I wanted to make sure that every piece of information matched its category (rather than assuming that td[1] was the legislator's party, I checked that the td[0] equalled "Affiliation", and etc.). There are a few vacant seats, so just for the sake of being able to scrape everyone (including the vacant seats), I just put a warning and assigned it a random party (guessing that this will be caught later that the seat is vacant), but please let me know if I should treat this differently!


scrapers_next/al/people.py
```diff
@@ -0,0 +1,144 @@
...
+                        # will assign it a random party for scraping purposes, but the warning should be addressed:
+                        party = "Democrat"
+
+            elif type == "District:":
+                district = info.split(" ")[2]
+            elif type == "County:":
+                county = info
+            elif type == "Phone Number:":
+                phone = info
+            elif type == "Fax Number:":
```
jamesturk, line 56: if any of these conditionals are never met in a loop, the fax/etc. from the prior iteration will be used

at the top of the for loop it'd be good to initialize all of them to None/empty strings accordingly, so that no information can bleed over from the prior iteration

scrapers_next/al/people.py
```diff
@@ -0,0 +1,144 @@
...
+
+            if type == "Affiliation:":
+                party = ""
+                if info == "(R)":
+                    party = "Republican"
+                elif info == "(D)":
+                    party = "Democrat"
+                else:
+                    party = info
+                    if party == "( )":
```
jamesturk, line 45: the scraper can skip vacancies, the data being missing will be a hint that the seat is vacant

scrapers_next/al/people.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class LegList(HtmlListPage):
+    def process_item(self, item):
+        last_name = re.split("Pictures/|_", item.get("src"))[1]
+        oid_person = item.get("alt")
+
+        # specific numbers in the img tag make up parts of the URL
+
+        if self.chamber == "upper":
```
jamesturk, line 119: generally it is best practice to not have code in a parent class that depends on knowing things from the child class

applied here that'd mean that instead of having one process_item function, you'd have two (one on RepList and one on SenList) and put the appropriate code there instead of in this conditional

jamesturk (changes requested):

jamesturk:
looking good, a few small issues- and you can skip vacancies instead of assigning them fake data 👍

PR #3757, opened 2021-08-05
Dc people spatula

morden35:
Writing dc people scraper to use spatula framework.

Notes:
1. Councilmember Kenyan R. McDuffie is listed twice. I skip him the first time.

2. Is there a way to use placeholders when I first assign p = ScrapePerson()? Right now I either use empty strings, or "Democratic" for the party (and later overwrite). See line 86.

3. I assign p.district as either 'chariman' (all lower), 'At-Large', or 'Ward \d'. Should I change any of this?

4. No district office info available.

5. What would I assign chamber for these people?

6. I am assigning the detail pages as 'homepage.' Some people also have websites, which I am adding to p.extras. ex: https://dccouncil.us/council/councilmember-robert-c-white-jr/ vs http://www.RobertWhiteAtLarge.com/

@jamesturk

### DIFF
--- a/scrapers_next/dc/people.py
+++ b/scrapers_next/dc/people.py
@@ -1,24 +1,46 @@
 from spatula import CSS, HtmlListPage, URL, HtmlPage, SelectorError
 from openstates.models import ScrapePerson
 import re
+from dataclasses import dataclass
+
+
+@dataclass
+class PartialPerson:
+    state: str
+    chamber: str
+    image: str
+    source1: str
+    source2: str
+    link: str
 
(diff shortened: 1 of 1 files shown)


jamesturk:
1. 👍 
2. this is where you'd use an intermediate model potentially (https://jamesturk.github.io/spatula/data-models/) some of the other scrapers do this if you want examples 
3. Chairman with a capital C is how we have it in our data 👍 
4. no district offices here AFAIK
5. chamber=legislature
6. 👍

PR #3761, opened 2021-08-06
AR people spatula scrape

judgejudes:
I tried something a bit different to avoid a bunch of if and elif statements to try to be more efficient, but I'm not sure if I wrote more lines in the end 😅  Ideally extra info shouldn't have to be checked with a matching string in the table dictionary, but for the sake of time and finishing goals for the sprint, I thought this would be okay for now.


jamesturk:
this makes sense to me, looks good!

PR #3764, opened 2021-08-12
us public law

jamesturk:
- add dates to versions
- US: add scrape_public_law_version method
- US: skip public law check if not enrolled


jamesturk:
this adds one missing version (the "Public Law" version) to the versions dict, the extra request is only done for the bills which already have an enrolled status, since in my test that was the only time that one exists & saves us several thousand requests

PR #3765, opened 2021-08-13
NJ: Get committee scraper working again

everettttt:
I stole code from `bills.py` to get `committees.py` working again. I tested it for the 2020-2021 session only, so far.

I ran into a new committee member role code (or "position on committee" code). "C" mean chair, "V" is vice-chair, "" is member, and now there's "O", which I assume means member of a commission. See, for example, one commission where all members have role code "O": [SELECT COMMISSION ON EMERGENCY COVID-19 BORROWING (SCEB)](https://www.njleg.state.nj.us/committees/sceb.asp); they're only called "members" on this page.

For now, I assign members with role code "O" to role "member", same as for code "". **Should members with role code "O" get a different role name instead?**

### DIFF
--- a/scrapers/nj/committees.py
+++ b/scrapers/nj/committees.py
@@ -5,12 +5,11 @@ from .utils import MDBMixin
 
 class NJCommitteeScraper(Scraper, MDBMixin):
     def scrape(self, session=None):
-        if not session:
-            session = self.jurisdiction.legislative_sessions[-1]["name"]
+        if session is None:
+            session = self.latest_session()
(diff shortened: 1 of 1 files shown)


jamesturk:
thanks for this!  we're currently porting committee scrapers to a new format (see scrapers_next for examples) but fixing up old ones is very helpful so that when we port them they are working 

it'll be a little bit until this one gets to run in production as we work to restore committees, but the changes look good and your decision on "O" makes sense to me based on what you cited!

PR #3766, opened 2021-08-13
CT people spatula scrape

judgejudes:
- the old scraper listed the reps' home address as their district office address so I did the same, but wanted to check if this was still okay
- listed the committees each rep is on only as extra information, not as an Organization object like the old scraper
- I've never used `assert` for the scrapers but just included the line because the old version had it
- I usually do some sort of party checking, but figured that this is already done during scraper, but wanted to make sure this is the case!
- apologies for the few commits, will commit more often next time!

### DIFF
--- a/scrapers_next/ct/people.py
+++ b/scrapers_next/ct/people.py
@@ -60,6 +60,8 @@ class LegList(CsvListPage):
         print("legislator url", legislator_url)
         p.add_link(legislator_url)
 
+        p.add_source(self.source.url)
+
         # the spacing of the address?
         office_address = "%s Room %s; %s" % (
             row["capitol street address"],
@@ -74,7 +76,7 @@ class LegList(CsvListPage):
 
         # spacing?
         if row["home street address"] and row["home city"]:
-            home_address = "{}\n{}, {} {}".format(
+            home_address = "{}; {}, {} {}".format(
                 row["home street address"],
                 row["home city"],
                 row["home state"],
--- a/scrapers_next/ct/people.py
+++ b/scrapers_next/ct/people.py
@@ -1,0 +1,54 @@
+# import attr
+from spatula import HtmlListPage, XPath, CSS
+
+# from openstates.models import ScrapePerson
+
+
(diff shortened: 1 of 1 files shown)


scrapers_next/ct/people.py
```diff
@@ -0,0 +1,91 @@
...
+from openstates.models import ScrapePerson
+
+
+class LegList(CsvListPage):
+    source = "ftp://ftp.cga.ct.gov/pub/data/LegislatorDatabase.csv"
+
+    def process_item(self, row):
+        name = row["first name"]
+        mid = row["middle initial"].strip()
+        if mid:
```
jamesturk, line 11: if you could replace this old style string formatting with new that'd be great for clarity

either

name += " " + row["last name"]

or name += f" {row['last name']}"

scrapers_next/ct/people.py
```diff
@@ -0,0 +1,91 @@
...
+        legislator_url = row["URL"].replace("\\", "//").strip()
+        if legislator_url != "":
+            if not legislator_url.startswith("http"):
+                legislator_url = "http://"
+        p.add_link(legislator_url)
+
+        p.add_source(self.source.url)
+
+        # the spacing of the address?
+        office_address = "%s Room %s; %s" % (
```
jamesturk, line 52: same here, {} would be better

scrapers_next/ct/people.py
```diff
@@ -0,0 +1,91 @@
...
+            row["capitol city"].replace("  ", " "),
+        )
+        p.capitol_office.address = office_address
+
+        if row["capitol phone"].strip():
+            p.capitol_office.voice = row["capitol phone"]
+
+        # spacing?
+        if row["home street address"] and row["home city"]:
+            home_address = "{}; {}, {} {}".format(
```
jamesturk, line 64: this is a good example of newer string formatting 👍

jamesturk (changes requested):
one small nit, looking good otherwise & all good w/ the questions raised, nothing to change on those!

PR #3767, opened 2021-08-13
TX: check names for spaces or newlines

NewAgeAirbender:
With some of the special session bills, new lines got pulled in as names since they weren't just empty strings. Done some test scrapes and the rest of the bills look okay, feel like the logic could be prettier so happy to take critiques.


jamesturk:
looks good to me

PR #3770, opened 2021-08-18
GA: gets token for events scraper to work

NewAgeAirbender:
Mostly duplicate code from the spatula scraper, should unify whenever that conversion happens for the event scrapers.
@jamesturk

### DIFF
--- a/scrapers/ga/events.py
+++ b/scrapers/ga/events.py
@@ -2,6 +2,9 @@ import pytz
 import datetime
 import dateutil.parser
 import re
+import time
+import requests
+from hashlib import sha512
 from openstates.scrape import Scraper, Event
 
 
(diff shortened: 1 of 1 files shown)


scrapers/ga/events.py
```diff
@@ -10,6 +13,19 @@ class GAEventScraper(Scraper):
     #  PYTHONPATH=scrapers poetry run os-update ga events --scrape start=YYYY-mm-dd
     tz = pytz.timezone("US/Eastern")
 
+    def get_key(timestamp):
```
jamesturk, line 16: nit: typically these would live outside the class so they didn't need the unusual calling pattern

scrapers/ga/events.py
```diff
@@ -20,8 +36,16 @@ def scrape(self, start=None):
         date_slug = start.strftime(date_format)
 
         url = f"https://www.legis.ga.gov/api/meetings?startDate={date_slug}"
-
-        page = self.get(url).json()
+        token = GAEventScraper.get_token()
```
jamesturk, line 39: this is a bit unusual for Python, typically either a classmethod or just a separate function outside the class is preferable

jamesturk (changes requested):
small tweak for now, will also make the eventual refactor a tiny bit easier

jamesturk:
There is now a file in scrapers_next/ga/utils.py that could be used to import this code from

jamesturk:
got this in

PR #3771, opened 2021-08-19
Mt os actions

morden35:
Appending committees string to action_name string ("Action - Committee"). Not sure if I should keep (S|H) in the committee name string (for example "(S) Finance and Claims" vs "Finance and Claims").

@jamesturk 
@jessemortenson


jamesturk:
looks good to me, @jessemortenson wanted to double check w/ you that this info being in the action text would resolve the issue at hand

jessemortenson:
Yes, I believe that will solve the issue for Enview (allow Enview code to grab the committee name instead of having to re-scrape). I vote for keeping the full string from source, eg `(S) Finance and Claims` as that gives downstream users the most information

PR #3776, opened 2021-08-24
MO committee spatula scraper

judgejudes:
for Senate:
- we're only scraping statutory and standing committees right now, so I debated deleting SenateCommitteeTypeList and instead directly scraping committees from the statutory and standing committee webpages. But just in case we wanted to scrape more committee types other than statutory and standing, going through one more [webpage](https://www.senate.mo.gov/committee/) was best?
- I didn't specify which ones were joint committees (the old scraper removed that distinction too)
- I didn't add members that were listed as "House Vacancy" or "Senate Vacancy"

for House:
- I'm a little unsure what the parents of the subcommittees are [here](https://www.house.mo.gov/CommitteeHierarchy.aspx). Right now, the subcommittee of the parent is the sub-bullet underneath. For ex. "Appropriations - Agriculture, Conservation, Natural Resources, and Economic Development" has the parent "Budget" and not "Appropriations" (there's no "Appropriations" committee), or "Rules". And "Mental Health Policy Research" has the parent "Health and Mental Health Policy".
- The ["Missouri Works Job Training Joint Legislative Oversight Committee"](https://www.house.mo.gov/Committees.aspx?category=all&committee=2571&year=2021&code=R) is empty but scraped anyway, just in case it's updated later.
- I didn't check for the legitimacy of the positions for each member (like "Vice Chair", etc.) because all positions were under the "Position" header in the table
- I scraped individual committee pages by making a URL that's different from the one that [this page](https://www.house.mo.gov/CommitteeHierarchy.aspx) directs you to. For ex., instead of https://www.house.mo.gov/Committees.aspx?category=all&committee=2582&year=2021&code=R, I scraped from https://www.house.mo.gov/MemberGridCluster.aspx?filter=compage&category=committee&Committees.aspx?category=all&committee=2582&year=2021&code=R. The old scraper did this too, but wanted to make sure this was still fine (it made scraping a lot easier :) ).

- I kept getting Timeout Errors, so I kept scraping using `--fastmode --rpm 10`

### DIFF
--- a/scrapers_next/mo/committees.py
+++ b/scrapers_next/mo/committees.py
@@ -144,7 +144,8 @@ class HouseCommitteeList(HtmlListPage):
         committee_name = item.text_content()
 
         committee_name = (
-            committee_name.replace("Joint Committee on", "")
+            committee_name.replace("Joint Committee on the", "")
+            .replace("Joint Committee on", "")
             .replace("Special Committee on", "")
             .replace("Special Interim Committee on", "")
             .replace(", Standing", "")
--- a/scrapers_next/mo/committees.py
+++ b/scrapers_next/mo/committees.py
@@ -1,5 +1,5 @@
 # import re
-from spatula import HtmlPage, HtmlListPage, CSS, XPath
+from spatula import HtmlPage, HtmlListPage, CSS, XPath, SelectorError
 from openstates.models import ScrapeCommittee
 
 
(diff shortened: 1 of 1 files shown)


scrapers_next/mo/committees.py
```diff
@@ -0,0 +1,187 @@
...
+
+class HouseCommitteeList(HtmlListPage):
+    source = "https://www.house.mo.gov/CommitteeHierarchy.aspx"
+    selector = CSS("a")
+    chamber = "lower"
+
+    def process_item(self, item):
+        committee_name = item.text_content()
+
+        committee_name = (
```
jamesturk, line 146: since this replace-chain is the same in both places, let's make a small utility function to clean up these names and call it in both places (so that future modifications only need to update one place and not keep them in sync)

scrapers_next/mo/committees.py
```diff
@@ -0,0 +1,187 @@
...
+            .replace(", Special Standing", "")
+        )
+        committee_name = committee_name.strip()
+
+        if "Subcommittee" in committee_name:
+            name = committee_name.replace("Subcommittee on ", "").replace(
+                ", Subcommittee", ""
+            )
+
+            parent = (
```
jamesturk, line 163: code is same here

scrapers_next/mo/committees.py
```diff
@@ -0,0 +1,187 @@
...
+        com.add_source(self.source.url)
+        com.add_link(self.source.url, note="homepage")
+
+        members = CSS(".gallery .desc").match(self.root)
+
+        positions = ["Chairman", "Vice-Chairman"]
+        for member in members:
+            member_position = member.text_content().replace("Senator", "").split(", ")
+
+            if (
```
jamesturk, line 20: 👍

scrapers_next/mo/committees.py
```diff
@@ -0,0 +1,187 @@
...
+            com = ScrapeCommittee(
+                name=name,
+                chamber=self.chamber,
+                classification="subcommittee",
+                parent=parent,
+            )
+        else:
+            com = ScrapeCommittee(name=committee_name, chamber=self.chamber)
+
+        # We can construct a URL that would make scraping easier, as opposed to the link that is directly given
```
jamesturk, line 184: 👍

jamesturk (changes requested):
one small change, everything else seems to look good

if you can submit a committees pr against openstates/people once this has run locally that'd be great

scrapers_next/mo/committees.py
judgejudes: I think I have to keep this here so that I don't replace the "Subcommittee" in the original string before I check for it and then replace it

judgejudes:
> if you can submit a committees pr against openstates/people once this has run locally that'd be great

for some reason I'm getting this error when I run a full scrape: 
`spatula.pages.MissingSourceError: SenateTypeCommitteeList has no source or get_source_from_input` at the very end (I'm pretty sure everything is scraped at this point). Everything's fine when I just scrape the Senate committees independently from the House, though. 
When I run the full scrape a few times, it doesn't scrape anything before throwing the error. Not sure what's going on!

jamesturk:
Ah that is likely an issue with having ListPage subclasses that aren't the final step, I'll take a look.

I think it need not block things though, if you run the scrapers independently you can run the merge step on each output directory (once for house and once for senate).

PR #3783, opened 2021-08-31
Sc people spatula

morden35:
Writing SC people scraper using spatula framework. Notes:

1. Emails are not available for any legislators.

2. Currently assigning 'Home Address' to p.district_office.address. Could alternatively assign to p.additional_offices.

3. The code to grab phone numbers could be simplified if we don't care about their labels. For example, a legislator can have 1-4 phones listed. Currently avoiding/checking for duplicates.
ex: https://www.scstatehouse.gov/member.php?code=1640908894
- Columbia Address Business Phone assigning to p.capitol_office.voice
- Columbia Address Home/Cell Phone assigning to p.extras
- Home Address Home Phone assigning to p.extras
- Home Address Business Phone assigning to p.district_office.voice

@jamesturk

### DIFF
--- a/scrapers_next/sc/people.py
+++ b/scrapers_next/sc/people.py
@@ -1,0 +1,40 @@
+from spatula import URL, CSS, HtmlListPage
+from openstates.models import ScrapePerson
+import re
+
+
+class Legislators(HtmlListPage):
+    selector = CSS("div.member")
+
+    def process_item(self, item):
+        name = CSS("a.membername").match_one(item).text_content()
+        name = re.search(r"(Senator|Representative)\s(.+)", name).groups()[1]
+
+        party = CSS("a.membername").match_one(item).tail.strip()
+        if party == "(D)":
+            party = "Democratic"
+        elif party == "(R)":
+            party = "Republican"
+
+        district = CSS("div.district a").match_one(item).text_content().strip()
+        district = re.search(r"District\s(.+)", district).groups()[0]
+
+        p = ScrapePerson(
+            name=name,
+            state="sc",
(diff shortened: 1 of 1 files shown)


scrapers_next/sc/people.py
```diff
@@ -0,0 +1,148 @@
...
+            home_line2 = home_addr_path.getchildren()[0].tail
+            if not re.search(r"SC", home_line2):
+                city, h_zip = re.search(r"(.+),?\s(\d{5})", home_line2).groups()
+                home_line2 = city + ", SC " + h_zip
+            home_addr += home_line2
+            p.district_office.address = home_addr
+        except SelectorError:
+            pass
+
+        cap_phones = XPath(
```
jamesturk, line 45: phew, I see what you mean about being able to simplify phones a bit

I wonder what you think about this approach (I haven't gone deep into the site/data, so maybe this isn't right)

it seems like there is a way to extract each type of phone # and each label, so could you do something where the code builds a dictionary like

phone_numbers = {
   "Capitol Phone": "123-456-7980",
   "Home Phone": "222-222-2222",
   "Business Phone": "444-444-444",
}

which would help detect/avoid dupes, and then we could just read out of that dictionary at the end when assigning to the various offices?

my thinking is that'd separate the code that pulls the various phones from the code that figures out where they should go, and would also help avoid duplicates and all of the checks within nested if statements

glad to talk more if this doesn't seem like it'll work or could use clarification

morden35: I see what you're saying. I took a crack at this. Looks a little better than previous code in my opinion.

PR #3784, opened 2021-08-31
USA: skip S 164

NewAgeAirbender:
Scraper is [failing due to a 503 error](https://github.com/openstates/issues/issues/482), this addition just skips that bill for now and we can come back to it later.


jamesturk:
good w/ me, let's add a comment though explaining it and maybe linking to the ticket?

& not necessary, but generally when doing these I tend to the exceptional case its own conditional (as opposed to using or) just so that it stands out & the other condition doesn't accidentally get wiped out when we clean this up.

PR #3793, opened 2021-09-08
Pr people spatula

morden35:
Writing PR people scraper using spatula framework. Notes:

1. For Senators, translating "Independiente" to "Independent" for party.

2. If district is "Senador/Representante por Acumulación", assigning district = "At-Large".

3. Almost all senators with district "Senadora por Distrito" are missing a district number. Old code does not produce anything for current pages. Not sure what to do with these.

4. All senators have the same phone, capitol office address, and mailing address. Should I hard-code these?

5. 3 representatives do not have party listed. Currently assigning to Independent.

6. Representatives have addresses listed, but they don't look like full/complete addresses.
(examples: Presidencia,  1er.PISO ANEXO, Sótano Capitolio)

7. Representative phones, fax and tty (what is this?) are being added to p.extras since I don't know if they belong to capitol or district offices.

@jamesturk


jamesturk:
1 & 2 & 5.  looks good
3. hmm, leave these scraping as-is, these pages are definitely broken but we can either hard code fixes later or rely on manual data for the broken ones
4. is it possible to scrape them?  if so I'd do that, otherwise hard coding is fine
6. capture what is there for now
7. let's add them as district for now, TTY is a phone for the deaf & can be in extras

morden35:
@jamesturk ready for re-review

3. sounds good. I added back the old code. all but 1 Senator are missing this info.
4. yes, I am currently scraping senate phone, capitol addresses, and mailing addresses
6 and 7. sounds good

PR #3796, opened 2021-09-09
Ut people spatula

morden35:
Writing UT people scraper using spatula framework. Notes:

1. No capitol office information available.

2. There is a comment in the old code about phone numbers:
"Work phone seems to be the person's non-legislative office phone, and thus a last option. For example, we called one and got the firm where he's a lawyer. We're picking them in order of how likely we think they are to actually get us to the person we care about."
I am currently assigning 'workPhone' to p.district_office.address, but maybe this should not be the case based on this comment. Cell and home phones are going in p.extras.

@jamesturk


jamesturk:
this looks good, I'd lean towards the work phone too-

PR #3797, opened 2021-09-10
Vt people spatula

morden35:
Writing VT people scraper using spatula framework. Notes:

1. Districts are names rather than numbers (ex: Windham-4, Rutland)

2. Each legislator has a Work Address, Home Address, and Mailing Address. Not sure which should be assigned to capitol office and which should be assigned to district. Currently putting all in p.extras. Old code assigned Mailing address to district office and hard-coded a capitol address.

3. Each legislator has 4 different types of emails: Email, Work Email, Home Email, and State Email. Not sure which one should be assigned p.email and which should go in extras. Currently adding all to p.extras['emails'] dictionary.

4. Each legislator has 'vTown' and 'Town' attributes. Not sure what these are (maybe home town?) or if they are needed. Not currently storing anywhere.

@jamesturk


jamesturk:
2. I'd keep the address logic the same as it was for continuity.
3. Probably the one that is just 'Email' in p.email 👍 
4. No need to capture these.

morden35:
@jamesturk ready for re-review

PR #3802, opened 2021-09-14
OR people spatula

morden35:
Writing OR people scraper using spatula framework. Notes:

1. The old code used apiclient.py and utils.py to connect to the api/url. I'm not sure this is necessary since we're only scraping the current session ('2021R1') and legislators. I left these files out of the new code, but please let me know if they should be used.

2. The old code was also able to index into a legislator object using strings ('WebSiteUrl', 'Chamber', etc). I couldn't figure out how to get this legislator object from the xml. Instead, I was able to index using integers, which is less clear to a future reader. Maybe this is related to how I'm connecting to the api? Please let me know if there is a better way to do this.

3. I was not able to scrape this link (not sure why): https://www.oregonlegislature.gov/bonham. For this person, I am just hard-coding his image instead of scraping.

4. This link has a different image: https://www.oregonlegislature.gov/wagner. I am just hard-coding his image.

5. Some legislators have district office addresses and phone numbers on their websites. This code to scraper this is a bit ugly since there is varied html, but it works.

@jamesturk


morden35:
@jamesturk 2 above is now taken care of. Ready for re-review

PR #3804, opened 2021-09-15
Wy people spatula

morden35:
Writing WY people scraper using spatula library. Notes:

1. I assumed item["phone"] is a district office phone. There is a field 'phoneType' ('Cell', 'Home', or 'Work'). Should I assign 'Work' as district office phone and the others as extras? Only 5 out of 90 phones are labeled 'Work'.

2. Note that "https://wyoleg.gov/LsoService/api/legislator/2021/(S|H)" and "https://wyoleg.gov/LsoService/api/legislator/{item['legID']}" are being added to sources while "http://wyoleg.gov/Legislators/2021/{item['party']}/{item['legID']}" is being added to links (homepage).

3. I assumed that self.data['address'] is a district office address.

4. No capitol office information available.

@jamesturk 
We did it! 52 jurisdictions now in spatula!!

### DIFF
--- a/scrapers_next/wy/people.py
+++ b/scrapers_next/wy/people.py
@@ -13,7 +13,8 @@ class LegDetail(JsonPage):
         distr_addr = f"{self.data['address']}, {self.data['city']}, {self.data['state']} {self.data['zip']}"
         p.district_office.address = distr_addr
 
-        # self.data['phoneType']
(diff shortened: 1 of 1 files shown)


jamesturk:
The district office assumptions makes sense to me, looks good from here!

🎆 52! 🎆

PR #3813, opened 2021-09-21
Enable EmptyScrape for Events scrapers

showerst:
Raise EmptyScrape for event scrapers when there are no events scheduled, as opposed to the scrapers emitting no results due to an error.


showerst:
@jamesturk just a quick check before I do the rest of these, is this the way you want this implemented? Import at top, raise check during the scrape?

jamesturk:
yeah this seems good to me, thanks for tackling!

PR #3821, opened 2021-09-28
AL: correct session for version urls

NewAgeAirbender:
AL has different url structure for 2021S1 bills, so just updates the version url to not be 404

Ex: 
http://alisondb.legislature.state.al.us/ALISON/SearchableInstruments/2021s1/PrintFiles/HR1-int.pdf should actually be
http://alisondb.legislature.state.al.us/ALISON/SearchableInstruments/2021FS/PrintFiles/HR1-int.pdf


showerst:
Mea culpa, I'd checked the output URLs here but not the versions before merging.

PR #3825, opened 2021-10-05
Co committee

dcliner:


dcliner:
The committee section for colorado was added in the co_committee branch

scrapers_next/al/committees.py
```diff
@@ -31,7 +31,7 @@ def process_page(self):
                 idx, val = info.text_content().strip().split(":")
                 com.extras[idx.strip()] = val.strip()
             else:
-                com.extras["Room"] = info.text_content().strip()
```
jamesturk, line 34: this change was probably an accident, let's revert it

scrapers_next/co/committees.py
```diff
@@ -0,0 +1,36 @@
...
+        return com
+
+class CommitteeList(HtmlListPage):
+    source = "https://leg.colorado.gov/content/committees"
+    selector = CSS("div.view-content tbody a[href]")
+
+    def process_item(self, item):
+        name = item.text
+        return CommitteeDetails(
+            name, source="https://leg.colorado.gov/content/committees")
```
jamesturk, line 36: this line is setting the CommitteeDetails source to the list page, when you want the `href` value

jamesturk (changes requested):
looking better, a couple of issues

make sure you run the entire scraper like this:

`poetry run spatula scrape scrapers_next.co.committes` that'll use the list page to get the URLs for the detail page (see the inline note on one issue there)

you'll also find that you need to handle other committee chambers (Joint Committee Year Round Committee comes up quickly for instance)  if it isn't upper or lower it should be `"legislature"` which means that the committee is just part of the legislature, not one chamber or the other

jamesturk:
Just gave this a run, it finishes if you pass --timeout 30 as an option, the pages are just slow.

`spatula scrape scrapers_next.co.committees --timeout 30`

The data however is missing legislators, it looks like only the first legislator is grabbed.

PR #3839, opened 2021-10-29
Ok committee

dcliner:
Inital code. still Fixing bugs


jamesturk:
unfortunately this work isn't on a clean branch, each time you start a new scraper you'll want to take the following steps:

1. checkout 'main'
2. pull from remote
3. make a new branch just for the state you're working on
4. push, making sure only the files you've changed are included

if you can copy the actual file to a new branch I can take a better look at it (similar to what you did with the others)

in the future, please be sure to use a clean branch (as you can see this PR changes 7 files instead of just one)

jamesturk:
still need a clean branch version of OK committee scraper, just wanted to remind you

PR #3840, opened 2021-10-29
FL: Added session for 2022 prefiled bills

nickrecchi:
openstates/issues#491
Prefiled bills are now being introduced for the Florida 2022 session. Added the 2022 session to the Florida scraper.

Source of the FL 2022 session dates:
https://www.myfloridahouse.gov/FileStores/Web/HouseContent/Approved/ClerksOffice/ImportantLegislativeDates.pdf


jamesturk:
Thanks for this!

 @NewAgeAirbender mind taking a look at this & adding the active flag if it is ready to go?

PR #3854, opened 2021-11-09
Added the inital code the rest are commented out to add the detail pa…

dcliner:
…ges individually.


jamesturk:
Looking good, it seems like there are some cases that aren't accounted for like http://www.kslegislature.org/li/b2021_22/committees/ctte_ot_criminal_justice_reform_1/

I'd recommend actually checking for role titles using CSS selectors instead of just using len(rolez), as it seems like there are different numbers & checking for the existence of the titles themselves will be more robust.

jamesturk:
@dcliner I've pushed a small change to this branch to highlight when it misses scraping members altogether: https://github.com/openstates/openstates-scrapers/pull/3854/commits/c932a095410c9c2f754c4fe0db02edaa10714be1

PR #3865, opened 2021-11-16
NH: Bills Fix

NewAgeAirbender:
FYI @jamesturk @showerst 
So now that NH added data back to the dynamicdatadump pages, we can scrape again. Most of the issues were just url tweaks, but the major issue is that the [LsrsOnly](http://www.gencourt.state.nh.us/dynamicdatadump/LsrsOnly.txt?x=2021161620081637095205) page that we scrape for versions has 2022 bill information, while the rest of the pages are still 2021 data. I moved to ignore scraping new versions for now since there isn't a way to set the session year and session is adjourned, but definitely open to ideas.

Also, the resolution check was just causing a bunch of redirects since looks like they've removed anything about old resolutions from the site and there aren't any prefiled yet.


NewAgeAirbender:
Those should be updated to the proper legacy urls now, actually spot checked this time.

showerst:
I just reran and it worked! Awesome.

PR #3868, opened 2021-11-18
LA: Ignore 2022 for now, add session commented out

showerst:


showerst:
No bills yet; http://www.legis.la.gov/legis/BillSearch.aspx?sid=22RS

PR #3871, opened 2021-11-22
PA_Committee_Scraper

dcliner:
Has some data issues as it get the members but repeats the names and gives them the members tag.


jessemortenson:
Hi Derefaa - I'm about to start looking at this code to give you more feedback, but I want to first note that the python linting check failed here on your last commit. When you push code that you want to be ready for review, please check the list of "checks" that github does, and if one fails, please try to address that. In this case it looks like you need to run flake8 and address code style

scrapers_next/pa/committees.py
```diff
@@ -0,0 +1,171 @@
...
+        try:
+            # This section has the chair memebers the regular, democratic and minority and the roles
+            # main chair
+            Chair_Member = (
+                CSS("div.MemberInfoList-MemberWrapper.ChairWrapper div.ChairNameText a")
+                .match(self.root)[0]
+                .text.strip()
+            )
+            # main chair role
+            Chair_Member_rolez = (
```
jessemortenson, line 23: We talked about variable naming conventions last week, and you have only implemented a couple of the pieces of feedback I gave you about variables. Two pieces that are missing:

- we talked about `rolez` and how that is not an English word. If you mean `roles` it needs to be `roles`
- Python style guide says that variable names in functions use lowercase letters. So this variable needs to be `chair_member_roles` (and other variables throughout need to follow this style). I pointed out the [PEP 8 python style guide](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names): that needs to be followed throughout, please make changes.

scrapers_next/pa/committees.py
```diff
@@ -0,0 +1,171 @@
...
+                    "div.MemberInfoList-MemberWrapper.ChairWrapper div.ChairNameText div"
+                )
+                .match(self.root)[1]
+                .text.strip()
+            )
+            com.add_member(Demo_Chair_Member, Demo_Chair_Member_rolez)
+        except IndexError:
+            pass
+        # Regular majority members and their roles
+        Majority_Members = XPath("//div[12]/div/div/div/div[2]").match(self.root)
```
jessemortenson, line 49: We talked about how there is a CSS selector that looks like it would be more appropriate to use to find the majority members, as well as another CSS selector for the minority members. Did you try that?

dcliner:
I tried the CSS selectors but the site rejected it for some reason so I
used the XPath.

jessemortenson:
Can you please share the specific CSS selector(s) you tried? Ideally as a series of commands/code you tried from a spatula shell session (with URL specified) so that I can understand what you mean by "the site rejected it". If you run into a bug/barrier, a good development habit is to report the specific behavior you tried and output/error you saw (with necessary details for another developer to be able to reproduce).

dcliner:
I'll get on, I'll start sharing in more detail.

scrapers_next/pa/committees.py
dcliner: The CSS selector (CSS("div.Widget CteeInfo-MajortityList")) didn't work it through back an error but as I started trying more selectors the only CSS selectors that worked were (CSS("div:nth-child(13) div div div div:nth-child(2)") and ("div a") for name and ("div.position") for the role/position. This one is much to similar to the XPath so i just left the XPath there.

jessemortenson: OK, so when you say "it threw back an error" that's an example of a lack of specifics - when I asked for "a series of commands/code you tried from a spatula shell session (with URL specified)" I'm looking for something like this:

```
(openstates-scrapers) jesse@codepc:~/repo/civiceagle/openstates$ spatula shell "https://www.legis.state.pa.us/cfdocs/CteeInfo/index.cfm?Code=32&CteeBody=H&SessYear=2021"
spatula 0.8.7 shell
available selectors: CSS, SimilarLink, XPath
local variables
---------------
url: https://www.legis.state.pa.us/cfdocs/CteeInfo/index.cfm?Code=32&CteeBody=H&SessYear=2021
resp: requests Response instance
root: `lxml HTML element` <html>
Python 3.9.5 (default, Jun  8 2021, 14:08:29) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.30.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: maj_list = CSS(".Widget.CteeInfo-MajorityList").match_one(root)
---------------------------------------------------------------------------
SelectorError                             Traceback (most recent call last)
~/.pyenv/versions/3.9.5/envs/openstates-scrapers/lib/python3.9/site-packages/spatula/cli.py in <module>
----> 1 maj_list = CSS(".Widget.CteeInfo-MajorityList").match_one(root)

~/.pyenv/versions/3.9.5/envs/openstates-scrapers/lib/python3.9/site-packages/spatula/selectors.py in match_one(self, element)
     78         :param element: Element to search within.
     79         """
---> 80         return self.match(element, num_items=1)[0]
     81 
     82     def get_items(self, element: _Element) -> Iterator[_Element]:  # pragma: no cover

~/.pyenv/versions/3.9.5/envs/openstates-scrapers/lib/python3.9/site-packages/spatula/selectors.py in match(self, element, min_items, max_items, num_items)
     55 
     56         if num_items is not None and len(items) != num_items:
---> 57             raise SelectorError(
     58                 f"{self} on {_display(element)} got {len(items)} results, "
     59                 f"expected {num_items}"

SelectorError: CSS(.Widget.CteeInfo-MajorityList) on <html lang='en'> @ line 2 got 2 results, expected 1
```

Generally when working with other developers it is better to provide too much detail than not enough. Plus, the task of documenting your own problem in detail often helps in your own troubleshooting process (see for example [rubber duck debugging](https://en.wikipedia.org/wiki/Rubber_duck_debugging).

The CSS selector you have above contains two errors: 
- `CteeInfo-MajortityList` is misspelled: the class from the site is `.CteeInfo-MajorityList` Please be sure to copy-paste strings from the website rather than retype, so you always get exactly the same string
- A space between selectors indicates that you are selecting a *child* of the first selector. So `div.Widget .CteeInfo-MajorityList` will look for a child element of `div.Widget` that has the class `.CteeInfo-MajorityList`. In this case what you want is an element that has both classes .Widget and .CteeInfo-MajorityList, so instead it should be `div.Widget.CteeInfo-MajorityList` (no space in between - they are chained together).

There is a working solution here using CSS selectors. Please try again :)

dcliner:
I will, I'll get back to you on the matter. Thanks for the back and forth.

dcliner (commented):
The selectors were changed and the code reshaped to include them

scrapers_next/pa/committees.py
```diff
@@ -0,0 +1,171 @@
...
+    example_input = ScrapeCommittee(
+        name=example_name, classification="committee", chamber="lower"
+    )
+
+    def process_page(self):
+        com = self.input
+        try:
+            # This section has the chair memebers the regular, democratic and minority and the roles
+            # main chair
+            Chair_Member = (
```
jessemortenson, line 17: Still seeing a lot of variables with capital letters in them. I need you to follow [python style guide](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names), as we talked about previously. That means variable names are "lowercase, with words separated by underscores as necessary to improve readability".

scrapers_next/pa/committees.py
```diff
@@ -0,0 +1,171 @@
...
+                .text.strip()
+            )
+            # Democratic Chair member and or the minority chair member role
+            Demo_Chair_Member_role = (
+                CSS(
+                    "div.MemberInfoList-MemberWrapper.ChairWrapper div.ChairNameText div"
+                )
+                .match(self.root)[1]
+                .text.strip()
+            )
```
jessemortenson, line 44: I notice in this section you are repeating (most of) a selector string four times. This is a risky code pattern because a) it is not clear to a later developer if these are actually the same four strings or not, and b) someone could try to fix a bug and only change one or two of the strings, not realizing that all four need to be changed. This needs to be refactored (changed) to use a different pattern. We talked about this previously.

One option is to use [f-strings](https://realpython.com/python-f-strings/) so that you can do 

``` 
chair_details_selector = "div.MemberInfoList-MemberWrapper.ChairWrapper div.ChairNameText"
chair_member = CSS(f"{chair_details_selector} a")
```

But an even better refactor would be to first select the container for each chair, then loop through and perform operations on each sub-selection to get individual pieces of data (rather than a list of four seemingly unrelated selections):

```
In [7]: chair_elements = CSS(".MemberInfoList-MemberWrapper.ChairWrapper").match(root)
for chair_element in chair_elements:
In [14]: for chair_element in chair_elements:
    ...:     member_name = CSS("a").match_one(chair_element).text.strip()
             ...etc.
             com.add_member(member_name, member_role)
```

scrapers_next/pa/committees.py
```diff
@@ -0,0 +1,171 @@
...
+                minor_mem_position = "Member"
+            com.add_member(minor_member_name, minor_mem_position)
+        return com
+
+
+class SenateCommitteeDetail(HtmlPage):
+    source = "https://www.legis.state.pa.us/cfdocs/CteeInfo/index.cfm?Code=32&CteeBody=H&SessYear=2021"
+    input = "Aging & Older Adult Services"
+
+    def process_page(self):
```
jessemortenson, line 75: Is the code in `SenateCommitteeDetail`'s `process_page` method exactly the same as `HouseCommitteeDetail`'s `process_page` method? It looks like it is. There are times when it makes sense to repeat a few lines of code in multiple places, however repeating two long functions that are exactly the same is a risky code pattern. Someone could find a bug in one and not know that they need to fix the other. It's not obvious to a later developer that these are exactly the same segments of code (they are long, which makes it harder to compare). If the HTML structure of the Senate and House committee pages are fundamentally the same, then it makes sense to use a single function to process both.

You can define a function outside of the classes, that can be called inside of the `process_page` method in both classes, see [this example](https://github.com/openstates/openstates-scrapers/blob/455cf66c6460a819a4e7f8bd31df59a5c7d4016e/scrapers_next/fl/people.py#L7).

scrapers_next/pa/committees.py
```diff
@@ -0,0 +1,171 @@
...
+        com.add_link(detail_link, "homepage")
+        return SenateCommitteeDetail(com, source=detail_link)
+
+
+class HouseCommitteeList(HtmlListPage):
+    source = "https://www.legis.state.pa.us/cfdocs/CteeInfo/StandingCommittees.cfm?CteeBody=H"
+    chamber = "lower"
+    selector = CSS("table tbody tr td:nth-child(1) a")
+
+    def process_item(self, item):
```
jessemortenson, line 155: Similarly, I note that  the `process_item` method of `HouseCommitteeList` and `SenateCommitteeList` are *almost* identical. This code could be refactored out to a shared function. However, because the repeated code is brief, it is more obvious that this code is being repeated. So it's a judgement call but I personally would not bother refactoring these out to a single function.

There is something called the [DRY principle - Don't Repeat Yourself](https://thevaluable.dev/dry-principle-cost-benefit-example/). It's a good thing to remember, but as that article states, it's not an absolute. It's part of the learning process to understand when to apply it by refactoring code to avoid repeating yourself, and when a little bit of repeating is OK.

jessemortenson (changes requested):

dcliner (commented):
Reduced the redundant code and tried to change all the variables to lower case.

PR #3891, opened 2021-12-22
FL: Avoid SSL Error

NewAgeAirbender:
So the scraper has been failing due to SSL errors like [this one](https://bobsled.openstates.org/run/fcb1cb8f3c164324a11068fd8b698684), so figured out how to get around it for the initial session list, but still getting the error on bill scrapes trying to hit `https://flsenate.gov/Session/Bills/2022?chamber=both`. Wasn't sure how to get around it using spatula, @jamesturk any thoughts?

Error text:
`requests.exceptions.SSLError: HTTPSConnectionPool(host='flsenate.gov', port=443): Max retries exceeded with url: /Session/Bills/2022?chamber=both (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)')))`

Also updating current sessions.

### DIFF
--- a/scrapers/fl/__init__.py
+++ b/scrapers/fl/__init__.py
@@ -15,7 +15,7 @@ class Florida(State):
         # "events": FlEventScraper,
     }
     # Full session list through 2019:
-    # https://www.flsenate.gov/PublishedContent/OFFICES/SECRETARY/SessionsoftheFloridaSenateFromStatehood.pdf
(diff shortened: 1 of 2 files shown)


jamesturk:
Ah you can pass verify=False into the URL() source object to skip verification.

https://jamesturk.github.io/spatula/reference/

If the source is a string you’ll need to turn it into URL(“https://whatever”, verify=False)

James

PR #3898, opened 2022-01-03
MS 2022

jamesturk:
Don't merge until MS has posted bills.


NewAgeAirbender:
[263 bills posted](http://billstatus.ls.state.ms.us/2022/pdf/all_measures/allmsrs.xml)

PR #3930, opened 2022-01-11
NJ: 2022 Session

NewAgeAirbender:
[2022 Bills are posted](https://www.njleg.state.nj.us/bill-search), adds new session and updates error expected when there isn't a vote at the url so the scrape doesn't fail.


jamesturk:
ah good catch, thanks for checking on that w/ the new session!

PR #3932, opened 2022-01-12
WIP: CO 2022 & new get_session_list

jamesturk:
old get_session_list was broken


showerst:
FYI I ran this and it looks good. Minor addition - the base url and ajax URL can now be moved to https. I can do that after merge as a separate PR, or you can add it here.

jamesturk:
thanks, updated the URLs here

PR #4069, opened 2022-06-21
VA: 2023 prefiles posted

NewAgeAirbender:
VA source is a mess. People are adding 2023 bills under the 2022 file right now and there aren't all the needed files posted yet so will be checking back before session start to get those.


showerst:
Haha i was just about to post an issue about this. 

I support this temp fix -- i'm not sure that 2022 should be False though, if nothing else there should be some enacted actions in July.

NewAgeAirbender:
Cool, will make that change & merge!

PR #4074, opened 2022-06-22
stop using external volume for local scraper db

johnseekins:

This should address the problem brought up in slack back in April: https://open-states.slack.com/archives/CQ1CUSRFG/p1649357266865389

It's not really clear why we had a volume for the local postgres instance, and simply using a local folder (that we also ignore in git) should be way easier to manage for people.


bfossen-ce (commented):
Tested this out on my computer off of this branch. I have an M1 Pro chip, ran the following successfully: docker-compose run --rm scrape nc bills --fastmode --scrape  Looks good to me.

bfossen-ce (approved):

PR #4094, opened 2022-08-08
AR: 2022 Third Special

showerst:
Pending, don't merge until there are bills to test.

### DIFF
--- a/scrapers/ar/__init__.py
+++ b/scrapers/ar/__init__.py
@@ -201,7 +201,7 @@ class Arkansas(State):
         {
             "_scraped_name": "Third Extraordinary Session, 2022",
             "classification": "special",
-            "identifier": "2021S3",
+            "identifier": "2022S3",
             "name": "2022 Third Extraordinary Session",
             "start_date": "2022-08-08",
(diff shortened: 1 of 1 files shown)


scrapers/ar/__init__.py
```diff
@@ -196,6 +196,15 @@ class Arkansas(State):
             "name": "2022 Fiscal Session",
             "start_date": "2022-02-14",
             "end_date": "2022-03-15",
+            "active": False,
+        },
+        {
+            "_scraped_name": "Third Extraordinary Session, 2022",
+            "classification": "special",
+            "identifier": "2021S3",
```
NewAgeAirbender, line 204: Need to swap this to 2022S3, but bills posted now

showerst: I left it at 2021 because it's the third special of the 2021 biennium, and that's the coding we've use for the other specials in that state.

showerst: Oh I see the issue now, that won't scrape correctly =P. I'll change that and merge in a bit.

PR #4109, opened 2022-08-08
Add NY Senate Collection

johnseekins:


scrapers_next/ny/people.py
```diff
@@ -158,3 +158,95 @@ def process_item(self, item):
...
+        )
+        party = (
+            CSS(
+                "div.nys-senator--info span.nys-senator--district span.nys-senator--party"
+            )
+            .match_one(item)
+            .text_content()
+            .strip()
+        )
+        district = (
```
NewAgeAirbender, line 225: District needs to just be the number value, but the rest looks good!

NewAgeAirbender (changes requested):

scrapers_next/ny/people.py
johnseekins: So our example should be `52` and not `52nd District`?

NewAgeAirbender: Yup

NewAgeAirbender (approved):

PR #4120, opened 2022-08-22
FL: Make previous sessions active so they can be re-scraped to fix sponsorships

nickrecchi:
openstates/issues#745
Inactive Florida sessions were changed to be active temporarily so they can be re-scraped. The re-scraping is needed to correct the bill sponsorships issue that was fixed in this pull request:
openstates/openstates-scrapers/pull/4068

Here are the sessions that were made active:
- 2017A
- 2018
- 2019
- 2020
- 2021
- 2021A
- 2021B

### DIFF
--- a/scrapers/fl/__init__.py
+++ b/scrapers/fl/__init__.py
@@ -108,6 +108,7 @@ class Florida(State):
             "classification": "special",
             "start_date": "2017-06-07",
             "end_date": "2017-06-09",
+            "active": True,
         },
         {
             "name": "2018 Regular Session",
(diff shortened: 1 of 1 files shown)


NewAgeAirbender:
Typically if we have sessions we need to rescrape for accuracy, we can pass them in as specific years in [task-definitions](https://github.com/openstates/task-definitions/blob/main/tasks/fl.yml) instead of setting them to "active" when they aren't actually. It's more accurate and is helpful for people who have the scraper repo forked that may not need that many sessions of FL data. Happy to ensure those sessions are actually rescraped for accurate sponsorship this week if you'd like!

nickrecchi:
Thanks for the explanation on how past sessions are rescraped. I would really appreciate if you ensured those sessions are rescraped this week. You can use the years from the session identifiers mentioned in the pull request to pass into the [task-definitions](https://github.com/openstates/task-definitions/blob/main/tasks/fl.yml).

PR #4146, opened 2022-10-20
NC: data validation fix on email attribute collected in process_page

chrisyamas:
Scraper was failing because process_page had not converted legislator's scraped email address from match object type to string type. Method `.text_content()` was added. Scraper now runs successfully without data validation error.


chrisyamas:
ValueError exception in `process_page()` is still needed as written to handle ValueError that otherwise results from SenList scrape (but not RepList).

NewAgeAirbender (approved):
Looks good!

PR #4147, opened 2022-10-24
GA: index error fix for collecting staff email

chrisyamas:
People scraper was failing because a legislator's details page did not have their staff member's name hyperlinked with an email address. Consequently, without `href` tags, the list `info` created within `for` loop in `process_page` for that legislator was of length = 1, resulting in `IndexError` when attempting to retrieve a staff name at `info[3]` and a staff email using `info[2]`.

Fix adds conditional statement to check the length of list `info` before retrieving staff name and email, with appropriate indexes given for each situation. Scraper now runs successfully.


NewAgeAirbender (approved):
Nice work, looks good!

PR #4148, opened 2022-10-24
TN: fix for XPath selector errors in people scraper

chrisyamas:
People scraper was failing when a legislator's details page did not content for District Address or Staff Contacts. This commit puts the XPath selectors for both content sections into `try...except` blocks to prevent either a `SelectorError` or `IndexError` from resulting. Scraper now runs successfully.


NewAgeAirbender (approved):
Looks good!

PR #4152, opened 2022-10-26
VT: improving bill action classification

sruthivedantham:
Improvements associated with [OS-783](https://civic-eagle.atlassian.net/browse/OS-783). 

These changes update the VT scraper to:

- perform bill action classification in a separate function (outside of the main bill scraping function)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 default_language_version:
-    python: python3.9
+    python: python3
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.4.0  # Use the ref you want to point at
--- a/scrapers/vt/bills.py
+++ b/scrapers/vt/bills.py
@@ -12,34 +12,35 @@ class VTBillScraper(Scraper, LXMLMixin):
 
     # Dictionary matching bill action phrases to classifications
     _actions = {
-        "signed by governor": ["executive-signature"],
-        "become law without signature of governor": ["became-law", "executive"],
-        "vetoed by the governor": ["executive-veto"],
         "read first": ["introduction"],
         "read 1st time": ["introduction"],
         "first read": ["introduction"],
-        "reported favorably": ["committee-passage-favorable"],
         "read second": ["reading-2"],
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-    python: python3.9
```
johnseekins, line 2: This actually does matter. Probably don't change this.

scrapers/vt/bills.py
```diff
@@ -9,6 +9,40 @@
...
+        "vetoed": ["executive-veto"],
+        "governor allowed to become law": ["became-law"],
+        "allowed to go into effect without the signature of the governor": [
+            "became-law"
+        ],
+        "proposal of amendment": ["amendment-introduction"],
+    }
+
+    def categorize_actions(self, action_string):
+        for action_key in list(self._actions):
```
johnseekins, line 40: `for action_key in self._actions.keys()`?

scrapers/vt/bills.py
```diff
@@ -198,24 +232,18 @@ def scrape(self, session=None):
...
+                )
+
+                # If both an action type and actor were assigned
+                if action_classification:
+                    action_type = action_classification[0]
+
+                    if len(action_classification) > 1:
+                        actor = action_classification[1]
+
+                # More action categorization (these were the tricky ones)
```
johnseekins, line 246: Since these are tricky, is it worth adding some examples in comments?

scrapers/vt/bills.py
sruthivedantham: good point! will change

scrapers/vt/bills.py
```diff
@@ -9,6 +9,41 @@
 
 
 class VTBillScraper(Scraper, LXMLMixin):
+
+    # Dictionary matching bill action phrases to classifications
```
johnseekins, line 13: Is there a location that our final classification list exists? Is it worth linking to that here?

sruthivedantham: yes there is! I will link

PR #4154, opened 2022-10-27
AK: improving bill action classification

sruthivedantham:
Improvements associated with [OS-782](https://civic-eagle.atlassian.net/browse/OS-782).

These changes update the AK scraper to:

### DIFF
--- a/scrapers/ak/bills.py
+++ b/scrapers/ak/bills.py
@@ -502,12 +502,11 @@ class AKBillScraper(Scraper):
         action = action.replace("PREFILE RELEASED", "Prefile released")
 
         atype = []
-        for action_key in self._actions.keys():
+        for classification_phrase, classifications in self._actions.items():
             # If we can detect a phrase that there is an OS action classification for
-            if action_key in action.lower():
+            if classification_phrase in action.lower():
                 # Our classification method for AK allows for more than one classification for certain actions
-                actions = self._actions[action_key]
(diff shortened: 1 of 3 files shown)


scrapers/ak/bills.py
```diff
@@ -479,40 +502,31 @@ def clean_action(self, action):
...
-        if "Prefile released" in action:
-            atype.append("filing")
-        if "Approved by the Governor" in action:
-            atype.append("executive-signature")
+        for action_key in self._actions.keys():
+            # If we can detect a phrase that there is an OS action classification for
+            if action_key in action.lower():
+                # Our classification method for AK allows for more than one classification for certain actions
+                actions = self._actions[action_key]
+                [atype.append(this_action) for this_action in actions]
```
johnseekins, line 510: `atype.extend(a for a in actions)`?

scrapers/ak/bills.py
```diff
@@ -479,40 +502,31 @@ def clean_action(self, action):
...
+
+                # Some cleaning that was done in the original code
+                if "TRANSMITTED TO GOVERNOR" in action:
+                    action = action.replace(
```
johnseekins, line 514: Consider `str.title()` for some of these changes?

scrapers/ak/bills.py
sruthivedantham: oh I didn't know you could do it this way!

PR #4155, opened 2022-10-28
MT: bill action classification updates

sruthivedantham:
Improvements associated with [OS-783](https://civic-eagle.atlassian.net/browse/OS-783).

These changes update the MT scraper to:

- reorganize and simplify the way actions are being classified (not using regex if we don't have to)

### DIFF
--- a/scrapers/mt/actions.py
+++ b/scrapers/mt/actions.py
@@ -219,7 +219,6 @@ _actions = {
 }
 
 
-# TODO: write function description comment, remove unncessary comments, CHANGE MT INIT FILE BACK
 # Takes in a string description of an action and returns the respective OS classification
 def categorize_actions(act_desc):
     atype = []
--- a/scrapers/mt/actions.py
+++ b/scrapers/mt/actions.py
@@ -2,358 +2,131 @@ import re
 
 
 # ----------------------------------------------------------------------------
-# Data for action categorization.
-
-_categories = {
+# Dictionary matching bill action phrases to classifications. Classifications can be found here:
+# https://github.com/openstates/openstates-core/blob/5b16776b1882da925e8e8d5c0a07160a7d649c69/openstates/data/common.py#L87
+# The "regex" included in some actions denotes that regex should be used to identify those phrases
+_actions = {
     # Bill is introduced or prefiled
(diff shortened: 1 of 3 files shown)


scrapers/mt/actions.py
```diff
@@ -2,358 +2,131 @@
...
-    "passage": {
-        "rgxs": [
-            "3rd Reading Passed",
-            "^Resolution Adopted",
-            "3rd Reading Concurred",
-            "3rd Reading Passed as Amended by Senate",
-            "3rd Reading Passed as Amended by House",
-        ]
-    },
+    "^introduced$": ["introduction", "regex"],
```
johnseekins, line 10: Is it worth making this format something like:
```
"^introduced$": {"type": "regex", "mappings": ["introduction"]},
"3rd reading passed": {"type": "compare", "mappings": ["passage", "reading-3"]},
...
```
This lets us avoid the whole `_actions[key][-1]` to check for `regex`.

So then processing looks something like:

```
def categorize_actions(act_desc):
    atype = []
    for action_key in _actions.keys():
        # If the action requires regex to be correctly classified
        if _actions[action_key]["type"] == "regex":
            if re.search(action_key, act_desc.lower()):
                # Add all action classifications in the mappings list
                atype.extend(a for a in _actions[action_key]["mappings"])
        else:
            if action_key.lower() in act_desc.lower():
                atype.extend(a for a in _actions[action_key]["mappings"])
```

scrapers/mt/actions.py
```diff
@@ -2,358 +2,237 @@
...
-        "Veto Override Failed in Legislature",
-        "Veto Override Motion Failed in House",
-        "Veto Override Motion Failed in Senate",
-        "Veto Override Vote Mail Poll Letter Being Prepared",
-        "Veto Override Vote Mail Poll in Progress",
-        "Vetoed by Governor",
-    ]
-)
 
+# TODO: write function description comment, remove unncessary comments, CHANGE MT INIT FILE BACK
```
NewAgeAirbender, line 222: Looks like you've done this so can be removed?

sruthivedantham: thanks for catching that!

PR #4156, opened 2022-10-29
AK: people scraper rewritten for HTML List Page

chrisyamas:
Alaska replaced the XML list page of members with an HTML list page. This PR contains a rewritten AK people scraper that successfully scrapes all member data from the new HTML list page.

### DIFF
--- a/scrapers_next/ak/people.py
+++ b/scrapers_next/ak/people.py
@@ -1,3 +1,4 @@
+import re
 import attr
 from spatula import HtmlPage, HtmlListPage, XPath, CSS, SelectorError
 from openstates.models import ScrapePerson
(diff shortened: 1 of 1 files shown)


scrapers_next/ak/people.py
```diff
@@ -1,69 +1,143 @@
...
+            "Fax",
+        ]
+
+        for pattern in detail_patterns:
+            match = re.search(rf"{pattern}:\s\S+", div_text)
+            if match:
+                detail = div_text[match.start() + len(pattern) + 2 : match.end()]
+                details[pattern] = detail
+
+        address_patterns = [r"Interim.+99\d\d\d", r"Session.+99801"]
```
NewAgeAirbender, line 45: May want to rethink this pattern, just in case of cases where both the session & interim addresses are ending with the 99801 zip code like [Senator Kiehl](https://www.akleg.gov/basis/Member/Detail/32?code=KIE)

scrapers_next/ak/people.py
```diff
@@ -1,69 +1,143 @@
...
-            given_name=item_dict["FirstName"],
-            family_name=item_dict["LastName"],
+        name_span = CSS(".formal_name").match(details_div)[0].text_content()
+        name_list = name_span.split(" ")
+        given_name = name_list[1]
+        family_name = " ".join(name_list[2:])
+
+        email = CSS("a").match(details_div)[0].text_content().strip()
+
+        div_text = details_div.text_content().replace("\r\n", " ")
```
NewAgeAirbender, line 27: May want to `split` up the text on new lines so that it's easier to get text for specific lines rather than saying exactly how many characters should be captured on match (thinking phone numbers). Another option is to create separate arrays for each contact block since they have two `<br>` between to remove some of the similar code/sanity checks that happen further down.

scrapers_next/ak/people.py
```diff
@@ -1,69 +1,143 @@
...
+        return p
 
-        if item_dict["Building"] == "CAPITOL":
-            person.capitol_office.address = (
-                "State Capitol Room {}; Juneau, AK, 99801".format(item_dict["Room"])
-            )
 
-        return person
+class LegList(HtmlListPage):
+    session_num = "32"
```
NewAgeAirbender, line 125: Nice following the practice of breaking this up even if it's set in the file instead of a passed in variable.

NewAgeAirbender (changes requested):
Overall, nice work! Have a few comments on how to break up the Contact blocks a little more, could also reference something like this [regex check](https://github.com/openstates/openstates-scrapers/blob/f97fe365665733a343e1addbf4f19ef25e5417bf/scrapers_next/id/committees.py#L74) for getting the contact details in a clean way. Happy to sync if you have questions or want clarification!

scrapers_next/ak/people.py
chrisyamas: The [newest commit has address pattern handling broken up](https://github.com/openstates/openstates-scrapers/pull/4156/commits/c6ab03807dad7b26ae8f8f3a2f5941177dea3fff#diff-fb9ba31d11ce10e77c56b79a04bbc75d0b174873ef40566f473109efb522f0f0R49-R63), which more effectively deals with cases like Senator Kiehl.

I also made the regex checks more elegant, and built a separate `format_address()` [helper function](https://github.com/openstates/openstates-scrapers/pull/4156/commits/c6ab03807dad7b26ae8f8f3a2f5941177dea3fff#diff-fb9ba31d11ce10e77c56b79a04bbc75d0b174873ef40566f473109efb522f0f0R7-R13) to ensure addresses have:
- collapsed whitespace
- a comma between city and state
- a comma between address lines (i.e. "PO Box 844 201 Katlian Street" becomes "PO Box 844, 201 Katlian Street")

scrapers_next/ak/people.py
chrisyamas: Re: our earlier conversation, I still chose to handle using text_content() for the whole div for now, in case AK's future web developer brings page more current (i.e. removes `<br>` tags).

scrapers_next/ak/people.py
chrisyamas: Thanks!

johnseekins:
Need to pull `main` into this branch to address changes from @showerst

chrisyamas:
@johnseekins which changes by @showerst were you [referencing](https://github.com/openstates/openstates-scrapers/pull/4156#issuecomment-1317521262)? I don't see any recent (or otherwise) changes to `scrapers_next/ak/people.py` from them.

chrisyamas:
@johnseekins did you mean [these changes to the AK events scraper](https://github.com/openstates/openstates-scrapers/commit/ad95e829f23249e3d46390b9cf9e471bcd151943)?

NewAgeAirbender (approved):

PR #4157, opened 2022-11-01
[should be deleted] ND: people scraper rewritten for new website

chrisyamas:
[Updated description on 11/24/22]: This PR should be deleted because the new rewrite of the ND people scraper was accidentally pushed from a different branch into the same pull request that has the rewrite of the ND events scraper. [That PR can be found here](https://github.com/openstates/openstates-scrapers/pull/4162).

[Original description]: Entire rewrite of ND people scraper built for state's new website. Includes pagination.

### DIFF
--- a/scrapers_next/nd/people.py
+++ b/scrapers_next/nd/people.py
@@ -1,118 +1,160 @@
-from spatula import URL, CSS, HtmlListPage, HtmlPage, XPath, SelectorError
-from openstates.models import ScrapePerson
 import re
+import attr
+from spatula import HtmlPage, HtmlListPage, XPath, CSS, SelectorError
+from openstates.models import ScrapePerson
+
+
+class PaginationError(Exception):
+    pass
+
+
+class PartyError(Exception):
+    pass
+
+
+@attr.s(auto_attribs=True)
+class PartialMember:
+    image: str
+    name: str
+    chamber: str
+    district: str
+    party: str
+    url: str
+    start_date: str = ""
 
 
 class LegDetail(HtmlPage):
+    example_source = "https://www.ndlegis.gov/biography/mary-adams"
+
     def process_page(self):
-        p = self.input
+        name_list = self.input.name.split()
+        given_name = name_list.pop(0)
+        family_name_list = [x for x in name_list if "." not in x]
+        family_name = " ".join(family_name_list)
 
-        img = CSS("div.field-person-photo img").match_one(self.root).get("src")
-        p.image = img
+        address_span = CSS(".address").match(self.root)[0].text_content()
+        address = ", ".join(address_span.split("\n")[:-1])
 
-        bio_info = CSS("div.pane-content ul li").match(self.root)
-        if len(bio_info) > 0:
-            p.extras["bio info"] = []
-            for info in bio_info:
-                p.extras["bio info"] += info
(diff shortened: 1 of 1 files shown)


scrapers_next/nd/people.py
```diff
@@ -1,118 +1,160 @@
...
+        start_date = ""
+        if status:
+            status_text = status[0].text_content().strip()
+            if "Expelled" in status_text:
+                self.skip()
+            elif "Active" in status_text:
+                start_date = status_text.split()[-1].strip()
+
+        district_div = item.find_class("district")[0].text_content().strip()
+        dist_match = re.search(r"District\s+\S+", district_div)
```
NewAgeAirbender, line 132: Should try to capture the district number in the search using something like [this](https://github.com/openstates/openstates-scrapers/blob/f97fe365665733a343e1addbf4f19ef25e5417bf/scrapers_next/tx/people.py#L86) so you don't need the next few lines

scrapers_next/nd/people.py
```diff
@@ -1,118 +1,160 @@
...
+        for phone_type in ("Cell", "Home"):
+            if collected_contacts.get(phone_type):
+                p.extras[phone_type.lower()] = collected_contacts[phone_type]
+
+        p.add_source(str(self.source), "member details page")
+
+        members_list_url = str(self.input.url)
+        p.add_source(members_list_url, "members list page")
+
+        url_match = re.search(r".+members", members_list_url)
```
NewAgeAirbender, line 78: Just one list page and the member detail page would be enough, but if you want to have it be the cleaner looking first page url, go for it!

NewAgeAirbender (changes requested):
Nice work, had a few areas to tweak, but I think the contact pattern check works well here.

scrapers_next/nd/people.py
chrisyamas: I implemented [this one-line solution](https://github.com/openstates/openstates-scrapers/pull/4162/commits/ed81cdcef99f666fd2d8464c036fe5263b687dee#diff-b4eb1b6b6189a33b171b390670cb2f7c25db0b25acd7e596c8c5afb685d778a6R136) in the most updated commit (which is on nd_events branch).

scrapers_next/nd/people.py
chrisyamas: I took your guidance and [deleted those lines](https://github.com/openstates/openstates-scrapers/pull/4162/commits/ed81cdcef99f666fd2d8464c036fe5263b687dee#diff-b4eb1b6b6189a33b171b390670cb2f7c25db0b25acd7e596c8c5afb685d778a6L78-L81) in the new commit.

chrisyamas:
> Nice work, had a few areas to tweak, but I think the contact pattern check works well here.

Thanks! Those areas got tweaked in the newest commit which is now a part of this [Nd events and people scraper rewrites](https://github.com/openstates/openstates-scrapers/pull/4162) pull request. I had accidentally pushed up recent additional tweaks to the ND people scraper while still on my nd_events I had been working on. However since that PR has the most updated version of both the new ND events and ND people scrapers, I think it would makes sense to merge it into openstates:main and then close/delete this branch and PR without merging as it's behind the other branch/PR at this point.

PR #4162, opened 2022-11-07
Nd events and people scraper rewrites

chrisyamas:
This pull request includes rewrites for both the ND events and ND people scrapers. Both rewrites were undertaken in response to ND legislature's launch of a new website that made our prior ND scrapers obsolete.
- This new people scraper handles pagination of member list pages.
- This new events scraper is scraping only from the Standing Committee Hearing schedule, as that was identified as the most relevant events page for our current purposes. The events scraper can be expanded upon later to scrape events from other pages on the new site: such an endeavor should have a process for preventing creation of duplicate event objects.
- **Note: the events currently listed on their Standing Committee Hearing schedule (as of the writing of this description on 11/24/22) are based on past events from the last legislative session in 2021. We presume this will be updated accordingly upon the start of new session in January 2023.**


scrapers/nd/events.py
```diff
@@ -1,114 +1,174 @@
...
+            date_time = item["date_time"]
+            local_date = dateutil.parser.parse(date_time)
+
+            item_date, item_time = str(local_date).split()
+
+            committee = item["committee"]
+            location = item["location"]
+            event_key = f"{item_date}-{committee}-{location}"
+
+            if not self.events.get(event_key):
```
johnseekins, line 39: `.get` should typically have a default value (e.g. `events.get(event_key, None)` to avoid a missing key error causing a stack trace.

scrapers/nd/events.py
```diff
@@ -1,114 +1,174 @@
...
+                self.events[event_key] = {}
```
johnseekins, line 40:
```
self.events[event_key] = {
    "event_start_time": item_time,
    "item_keys": set(),
}
```
Is probably cleaner here.

scrapers/nd/events.py
```diff
@@ -1,114 +1,174 @@
...
+                    descr_with_time = f"[{time}]: {item['description']}"
+                    item_descr = event_obj.add_agenda_item(descr_with_time)
+                    if item["bill_name"]:
+                        item_descr.add_bill(item["bill_name"])
+                    if item["sub_com"]:
+                        item_descr.extras["sub_committee"] = item["sub_com"]
+
+            event_obj.add_source(self.url)
+
+            event_collection.append(event_obj)
```
johnseekins, line 93: Perhaps do a `yield event_obj` here (and remove the `event_collection` object entirely) so the upper generator can actually yield results as they come in rather than all at once?

scrapers/nd/events.py
```diff
@@ -1,114 +1,174 @@
...
+        table_rows = self.root.xpath(".//tbody//tr")
+        agenda_items_list = []
+
+        for row_item in table_rows:
+            columns = row_item.xpath("./td")
+            columns_content = [x.text_content().strip() for x in columns]
+            if len(columns_content) == 7:
+                columns_content = columns_content[:-1]
+
+            bill, part_date, com, sub_com, loc, descr = columns_content
```
johnseekins, line 127: Given the number of values we're pulling from this column, is it worth adding an example column in a comment for later reference?

scrapers/nd/events.py
```diff
@@ -1,114 +1,174 @@
...
+                else:
+                    bill_name = ""
+
+            date_time_parts = part_date.split()
+            month = int(re.search(r"\d+", date_time_parts[0]).group())
+            if month == 1 and not start_year == end_year:
+                event_year = end_year
+            else:
+                event_year = start_year
+            date_time_parts[0] = date_time_parts[0] + f"/{event_year}"
```
johnseekins, line 151: `date_time_parts[0] += f"/{event_year}"` You can use `+=` to append to a string.

scrapers_next/nd/people.py
```diff
@@ -1,118 +1,160 @@
...
+    start_date: str = ""
 
 
 class LegDetail(HtmlPage):
+    example_source = "https://www.ndlegis.gov/biography/mary-adams"
+
     def process_page(self):
-        p = self.input
+        name_list = self.input.name.split()
+        given_name = name_list.pop(0)
```
johnseekins, line 31: Worth documenting why we're popping a random magic number.

scrapers_next/nd/people.py
```diff
@@ -1,118 +1,160 @@
...
-            town = CSS("span.locality").match_one(self.root).text_content().strip()
-            zip_code = (
-                CSS("span.postal-code").match_one(self.root).text_content().strip()
-            )
-            address = street + ", " + town + ", ND " + zip_code
-            p.district_office.address = address
-        except SelectorError:
-            pass
+        contact_patterns = ["Cell", "Fax", "Home", "Work", "Email"]
+        collected_contacts = {}
```
johnseekins, line 42:
```
collected_contacts = {
    "Cell": "", "Fax": "", "Home": "", "Work": "", "Email": ""
}
```

scrapers_next/nd/people.py
```diff
@@ -1,118 +1,160 @@
...
-                phone_number = phone.getnext().text_content().strip()
-                if phone_type == "Cellphone:":
-                    p.extras["Cell phone"] = phone_number
-                elif phone_type == "Home Telephone:":
-                    p.extras["Home phone"] = phone_number
-                elif phone_type == "Office Telephone:":
-                    p.district_office.voice = phone_number
-        except SelectorError:
-            pass
+        for pattern in contact_patterns:
```
johnseekins, line 44: With above change:
```
for pattern in collected_contacts.keys():
```

scrapers/nd/events.py
chrisyamas: Thanks, I'll make that common practice moving forward. [Here's the committed change to that line](https://github.com/openstates/openstates-scrapers/pull/4162/commits/037209511528934a8e11f7616c83939e636b489a#diff-fa16689e28573ece469e8dfea222521c243ef203a1c8c974d48d6b050b3ed308R39).

scrapers/nd/events.py
chrisyamas: Definitely cleaner. I [changed the construction](https://github.com/openstates/openstates-scrapers/pull/4162/commits/037209511528934a8e11f7616c83939e636b489a#diff-fa16689e28573ece469e8dfea222521c243ef203a1c8c974d48d6b050b3ed308R40-R43) accordingly.

scrapers/nd/events.py
chrisyamas: That makes sense, [and I changed it accordingly here](https://github.com/openstates/openstates-scrapers/pull/4162/commits/037209511528934a8e11f7616c83939e636b489a#diff-fa16689e28573ece469e8dfea222521c243ef203a1c8c974d48d6b050b3ed308L93-R93). However, I noticed the feedback in the terminal still doesn't reflect the creation of any new data files until the final moment of the scraper runtime. I'm assuming this is because it's the NDEventScraper(Scraper) object that finally initializes the json files when the scrape has completed? Or is it due to the way I'm yielding/returning objects in the current class construction anywhere in [the overall file](https://github.com/openstates/openstates-scrapers/pull/4162/commits/037209511528934a8e11f7616c83939e636b489a#diff-fa16689e28573ece469e8dfea222521c243ef203a1c8c974d48d6b050b3ed308)?

johnseekins: I believe that's a result of how scrapelib/spatula generates the final json file(s), yes. This is more a "if you're treating the function as a generator, probably actually make it one" type of suggestion. I don't think it _really_ affects performance/etc. in any meaningful way.

scrapers/nd/events.py
chrisyamas: For sure. [Is this example column](https://github.com/openstates/openstates-scrapers/pull/4162/commits/037209511528934a8e11f7616c83939e636b489a#diff-fa16689e28573ece469e8dfea222521c243ef203a1c8c974d48d6b050b3ed308R126-R128) what you were imagining, or at least does it seem satisfactory for later reference?

scrapers/nd/events.py
chrisyamas: Rookie move! [Fixed.](https://github.com/openstates/openstates-scrapers/pull/4162/commits/037209511528934a8e11f7616c83939e636b489a#diff-fa16689e28573ece469e8dfea222521c243ef203a1c8c974d48d6b050b3ed308R152)

scrapers_next/nd/people.py
chrisyamas: Thank you for pointing this out, because upon review I realized I never needed to pop() at all. I went ahead and [changed to an indexing approach](https://github.com/openstates/openstates-scrapers/pull/4162/commits/ed81cdcef99f666fd2d8464c036fe5263b687dee#diff-b4eb1b6b6189a33b171b390670cb2f7c25db0b25acd7e596c8c5afb685d778a6R35-R37) for accessing  a member's `given_name` and `family_name`.

But after running this I also realized my approach for isolating `family_name` wasn't sufficient. While it excluded middle initials like "B." and suffixes like "Jr.", in the cases where someone had a name with a suffix (i.e. "Howard C. Anderson, Jr." it accidentally captured the comma as part of the family_name so it was getting stored as "Anderson,"
- the newly committed line 37 eliminates any such commas.

scrapers_next/nd/people.py
chrisyamas: [Done.](https://github.com/openstates/openstates-scrapers/pull/4162/commits/ed81cdcef99f666fd2d8464c036fe5263b687dee#diff-b4eb1b6b6189a33b171b390670cb2f7c25db0b25acd7e596c8c5afb685d778a6R48-R54)

scrapers_next/nd/people.py
chrisyamas: [Also done.](https://github.com/openstates/openstates-scrapers/pull/4162/commits/ed81cdcef99f666fd2d8464c036fe5263b687dee#diff-b4eb1b6b6189a33b171b390670cb2f7c25db0b25acd7e596c8c5afb685d778a6R56)

scrapers/nd/events.py
chrisyamas: Okay, good to know.

bfossen-ce (approved):
Looks good to me. Good job with all the resolved changes from prior reviews. I ran people and events and it works well.

NewAgeAirbender:
People scraper looks good! Ran the events scraper and looks like they're getting consolidated a little too well and we're missing separate meetings in a day. It's a tricky puzzle but solid start!

PR #4168, opened 2022-11-10
MS: people scraper fixes

chrisyamas:
This PR was undertaken to fix an issue that was causing the MS people scraper to fail, receiving a 403 error in response to certain requests of member details page URLs.
- This was due to the fact that invalid URLs were being assembled for those requests whenever the scraper encountered a cell in the table of member names that did not contain a link. As of 11/24/22, this happens when vacant seats are encountered on the House members page (i.e. "Vacancy - District 23" instead of a hyperlinked Member name), but would presumably happen whenever hyperlinked text is not present in a cell.
- Solution to this issue in most recent file commit: a cell is skipped whenever there are absent XML page links.

In addition to fixing the above issue, two additional changes were added to improve upon the existing code:
- a `set()` was created to collect member detail XML page URL strings as each member is scraped, including a check to prevent duplicates
  - this is a longer-term solution than the existing code's handling which specifically named an individual ("Dean Kirby") in Senate leadership and instructed a skip of his duplicate listing. The new approach will prevent any duplicates on either house or senate people scrapes, should they occur in the future sessions with different leadership/members.
- cleaner, non-individual-name-specific code was implemented for skipping the "President of the Senate" (a.k.a. the Lieutenant Governor, an executive/non-legislative government official)

### DIFF
--- a/scrapers_next/ms/people.py
+++ b/scrapers_next/ms/people.py
@@ -157,26 +157,28 @@ class LegDetail(HtmlPage):
 class Legislators(HtmlPage):
     def process_page(self):
         members = self.root.getchildren()
-        count = 0
+        member_links = set()
         for member in members:
-            count += 1
             children = member.getchildren()
             if not children:
                 continue
             elif len(children) == 3:
                 lt_gov = "President of the Senate"
-                pres_pro_temp = "President Pro Tempore"
                 title = children[0].text_content().strip()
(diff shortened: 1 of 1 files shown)


bfossen-ce:
Looks good to me!

bfossen-ce (approved):

NewAgeAirbender (approved):

PR #4185, opened 2022-11-22
AR: Fix bill scraper for new ftp

showerst:
@NewAgeAirbender FYI you'll need to add a new set of secrets to the job definition 

AR_FTP_USER
AR_FTP_PASSWORD

I don't remember off the top of my head where this lives but if you can't find it let james and I know.


NewAgeAirbender:
Added them in, so should be good to go!

PR #4190, opened 2022-11-23
ND: bills scraper rewrite draft for new site, needs sponsors and more added

chrisyamas:
Draft of rewrite for ND bills scraper to work with the state's new legis site.


scrapers/nd/bills.py
```diff
@@ -1,229 +1,174 @@
...
+from spatula import HtmlListPage, HtmlPage, XPath, CSS
 
-class NDBillScraper(Scraper, LXMLMixin):
-    """
-    Scrapes available legislative information from the website of the North
-    Dakota legislature and stores it in the openstates  backend.
-    """
 
-    categorizer = NDCategorizer()
+def random_sleep(avg_secs):
```
johnseekins, line 12: Let's document this function. _Why_ do we need random sleeps? Also worth asking, should this be a general function (in the global `utils` or even openstates-core) instead?

chrisyamas: Yes, I'll add a comment. The sleeps were an attempt to put less strain on the server after having several runs that resulted in server errors indicating it might have been overloaded with requests. The randomization of the frequency and duration of the sleeps was to seem less-automated. The site's [robots.txt](https://ndlegis.gov/robots.txt) does not explicitly address these issues.

I could write a version of this function that could be added to global utils that could have default parameters for the magnitude of duration and frequency, so that future scraper writers could adjust them to suit specific needs of other sites.

However, if there is functionality within spatula or openstates-core already for something like this, I will defer to that. This was my ad hoc solution.

chrisyamas: Also I meant to comment out the print statement within that function so that it can be optional (and to delete the `print(bill_id)` on line 41). Both of which I can include on the next commit.

johnseekins: I see no _problem_ with you adding the function. Just making sure you're going to explain it. This is the sort of function that people come upon later and are very confused by.

Also, remember to _always_ document magic numbers (e.g. the 7, 14, 1, and 2 in this function).

showerst:
Rather than the random sleeps, you can also pass RPM=20 (or whatever) to the scrape command to knock the total requests per minute down. Will that resolve it without the custom function?

scrapers/nd/bills.py
```diff
@@ -1,229 +1,161 @@
...
+
+            (date,) = row.xpath("td[1]/b/text()")
+            date += f"/{self.input.legislative_session}"
+            date = dt.datetime.strptime(date, "%m/%d/%Y")
+
+            self.input.add_action(
+                action,
+                date.strftime("%Y-%m-%d"),
+                chamber=actor,
+                classification="introduction",
```
NewAgeAirbender, line 153: Make sure to use the action classifier here!

NewAgeAirbender (changes requested):
Just two slight tweaks but great work!

scrapers/nd/bills.py
```diff
@@ -1,229 +1,161 @@
...
+                if name:
+                    (name,) = name
+                else:
+                    (name,) = link.xpath("text()")
+
+                type_badge = link.xpath("span[@data-toggle='tooltip']")
+                if type_badge:
+                    (vers_type,) = type_badge[0].xpath("@title")
+                    if vers_type.strip() == "Marked up":
+                        name += "(Marked up)"
```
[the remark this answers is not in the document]
chrisyamas: Good catch! [Added it in](https://github.com/openstates/openstates-scrapers/pull/4190/commits/765499233afd0d3168d57142cb4b5a039c6f6d67#diff-1ada3395bc035a7192d515167751ad1b19db1e3ef3b2b7d4ae9052d41a75b15cR115).

scrapers/nd/bills.py
chrisyamas: Of course, I forgot to take out that placeholder. [Here it is added in](https://github.com/openstates/openstates-scrapers/pull/4190/commits/765499233afd0d3168d57142cb4b5a039c6f6d67#diff-1ada3395bc035a7192d515167751ad1b19db1e3ef3b2b7d4ae9052d41a75b15cR156).

NewAgeAirbender (approved):
Looks good!

PR #4201, opened 2022-11-29
Draft: Misc Event Updates

showerst:
Just batching together a bunch of minor event scraper improvements.


johnseekins:
@showerst Any reason not to merge this? Most of these jobs aren't actually _enabled_ in OpenStates, and having these changes would help us better fix events properly.

showerst:
@johnseekins I hadn't merged because I am still not thrilled with the match_coordinates implementation; it feels a bit un-pythonic to me. That said, no need to let perfect be the enemy of the good here, I'll merge it up.

FYI I see you're enabling a bunch of these for OS, many of the event scrapers still don't throw emptyscrape because there's no way to tell "no events" from "broken page or changed structure" so you'll see a lot of errors until January. That's normal.

PR #4203, opened 2022-11-30
update for legislative session dates

sruthivedantham:
session dates updated as per [OS-730](https://civic-eagle.atlassian.net/browse/OS-730)


NewAgeAirbender (approved):
LGTM!

PR #4204, opened 2022-11-30
[NH,CT,ID]: action classification updates

sroomf:
PR for action classification updates. These updates aim to increase action classification coverage as well as standardize the classification code so that it utilizes the BaseCategorizer class defined [here](https://github.com/openstates/openstates-scrapers/blob/3e5b5b1eec50f71a7b15b8be1c409a8be483694d/scrapers/utils/actions.py#L62).

### DIFF
--- a/scrapers/id/bills.py
+++ b/scrapers/id/bills.py
@@ -220,7 +220,6 @@ class IDBillScraper(Scraper):
             if len(row[2]):
                 action = "".join(row[2].itertext())
             action = action.replace("\xa0", " ").strip()
-            # atype = get_action(actor, action)
             attrs = self.categorizer.categorize(action)
             atype = attrs["classification"]
             if atype and "passage" in atype:
--- a/scrapers/id/bills.py
+++ b/scrapers/id/bills.py
@@ -4,6 +4,8 @@ import re
 import datetime
 from collections import defaultdict
 import lxml.html
+from .actions import Categorizer
+
 
 BILLS_URL = "https://legislature.idaho.gov/sessioninfo/%s/legislation/minidata/"
 BILL_URL = "https://legislature.idaho.gov/sessioninfo/%s/legislation/%s/"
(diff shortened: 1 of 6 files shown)


scrapers/nh/actions.py
```diff
@@ -0,0 +1,35 @@
...
+    Rule(r"^(?!.*Committee Report).*[Rr]eferred [Tt]o.*$", "referral-committee"),
+    Rule(r"Lay.*on Table.*$", "deferral"),
+    Rule(
+        r"^(?!.*Minority).*Committee Report: Inexpedient to Legislate",
+        "committee-failure",
+    ),
+)
+
+
+class Categorizer(BaseCategorizer):
```
johnseekins, line 29: I see this boilerplate in a few different scrapers. Is it possible to simplify this? Maybe make this part of the default object so people can just say `from utils.actions import Categorizer`? Or is that worth the effort?

johnseekins:
`scrapers/id/actions.py` was committed as an empty file. Was this intentional?

scrapers/nh/actions.py
sroomf: Just for my understanding, are you suggesting we move this chunk of code:
```
class Categorizer(BaseCategorizer):
    rules = rules

    def categorize(self, text):
        attrs = BaseCategorizer.categorize(self, text)
        return attrs
```
to utils.actions so  we can directly import `Categorizer`? I like that idea but the `rules` definition would still need to happen in each jurisdiction's `actions.py`.  I think we would still need to have
```
class Categorizer(BaseCategorizer):
     rules = rules
```
in every actions.py?

sroomf:
> `scrapers/id/actions.py` was committed as an empty file. Was this intentional?

scrapers/nh/actions.py
johnseekins: I missed that bit. Probably best to keep things this way and just document _how/why_ we import classifiers.

scrapers/id/bills.py
```diff
@@ -294,7 +220,9 @@ def _split(string):
             if len(row[2]):
                 action = "".join(row[2].itertext())
             action = action.replace("\xa0", " ").strip()
-            atype = get_action(actor, action)
+            # atype = get_action(actor, action)
```
NewAgeAirbender, line 223: Can delete this instead of comment out

NewAgeAirbender (approved):
Looks good besides that one comment

scrapers/id/bills.py
sroomf: good catch, thank you!

PR #4225, opened 2022-12-13
ND: votes rewritten for new site, still needs select form option handling

chrisyamas:
ND legislative site redesign required a rewrite of our votes scraper.

NOTE: Because all vote data from prior session has been removed from the Roll Call Vote page we are seeking to scrape, this rewrite is currently writing over the page source url with a url for an archived copy of the page. This will allow the new votes scraper to be tested until new vote data appears on the page in the new session.

### DIFF
--- a/scrapers/nd/votes.py
+++ b/scrapers/nd/votes.py
@@ -1,307 +1,74 @@
-import os
-import re
+import logging
 import datetime
-import requests.exceptions
-from utils import LXMLMixin
-from openstates.utils import convert_pdf
 from openstates.scrape import Scraper, VoteEvent as Vote
+from spatula import HtmlPage
 
 
-date_re = (
-    r".*(?P<date>(MONDAY|TUESDAY|WEDNESDAY|"
-    + r"THURSDAY|FRIDAY|SATURDAY|SUNDAY),\s\w+\s\d{1,2},\s\d{4}).*"
-)
-chamber_re = r".*JOURNAL OF THE ((HOUSE)|(SENATE)).*\d+.*DAY.*"
-page_re = r"Page\s\d+"
+# TODO: Identify method for scraping how each individual lawmaker voted
+#  (does not yet seem to be available on page)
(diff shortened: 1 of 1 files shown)


johnseekins:
Testing this PR I see the following errors:
1. The `session_slug` variable (line 26) isn't pulling from __init__.py, so hasn't been updated for 2023 _or_ 2021.
2. The `url` variable points to https://www.ndlegis.gov/assembly/68-2023/regular/journals/senate-journal.html, but that doesn't exist, while https://www.ndlegis.gov/assembly/68-2023/regular/journals/journal-index.html does.

The second problem means that the way we're currently processing pages doesn't seem to work at all. It's possible this scraper needs to be re-evaluated entirely.

chrisyamas:
@johnseekins re: your comment regarding those errors, are you sure that you tested the rewrite in this PR and not the file as it is currently in main branch that prompted this rewrite?

I ask because (1) `session_slug` does not exist as a variable in this rewrite (but it does in the version on main, [as seen here removed by this PR](https://github.com/openstates/openstates-scrapers/pull/4225/files#diff-2f6b7c8c09c31b342c31231c6f5ec6e2bca40c8a4784ec01fea0e970e453f608L24-L32)), and (2) `url` only exists as a variable in the rewrite [when pointing to links on each vote item](https://github.com/openstates/openstates-scrapers/pull/4225/files#diff-2f6b7c8c09c31b342c31231c6f5ec6e2bca40c8a4784ec01fea0e970e453f608R58-R59), but it was [pointing to the path you linked in the prior version on main](https://github.com/openstates/openstates-scrapers/pull/4225/files#diff-2f6b7c8c09c31b342c31231c6f5ec6e2bca40c8a4784ec01fea0e970e453f608R58-R59).

johnseekins:
Rescinding all my comments. This works well. We should merge it.

johnseekins (approved):

PR #4226, opened 2022-12-13
[UT, WI, VA, FL, TX, RI]: Action classification updates

sroomf:
PR for action classification updates. These updates aim to increase action classification coverage as well as standardize the classification code so that it utilizes the BaseCategorizer class defined [here](https://github.com/openstates/openstates-scrapers/blob/3e5b5b1eec50f71a7b15b8be1c409a8be483694d/scrapers/utils/actions.py#L62).


scrapers/va/csv_bills.py
```diff
@@ -9,38 +9,10 @@
 import time
 
 from .common import SESSION_SITE_IDS, COMBINED_SESSIONS
+from .actions import Categorizer
 
 tz = pytz.timezone("America/New_York")
 SKIP = "~~~SKIP~~~"
```
johnseekins, line 15: Dup of `SKIP` above in actions.py?

sroomf: yes! and actually unnecessary in both places - will delete, thank you

NewAgeAirbender (approved):

PR #4230, opened 2022-12-14
SC people: fixes for missing address, alt address format

chrisyamas:
SC people scraper was failing with:

1. IndexError when working on capitol address
  - Problem line: `line2 = cap_addr_path.getchildren()[0].tail`
  - Cause: member's page missing a capitol address
  - Solution: wrapped section in conditional to handle and store capitol address only when XPath selector encounters one
2. AttributeError when working on home address
  - Problem line: `city, h_zip = self.zip2_re.search(home_line2).groups()`
  - Cause: member who had only a zip code on the second line of their home address
  - Solution: split section into conditionals to handle each case

Additionally discovered:
1. Unnecessary/unwanted space stored before comma in a couple addresses
  - Solution: regex substitution

### DIFF
--- a/scrapers_next/sc/people.py
+++ b/scrapers_next/sc/people.py
@@ -17,14 +17,16 @@ class LegDetail(HtmlPage):
             .match_one(self.root)
             .getnext()
         )
-        cap_addr = cap_addr_path.text or ""
-        cap_addr += " "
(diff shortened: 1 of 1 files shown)


NewAgeAirbender (approved):
LGTM

PR #4233, opened 2022-12-14
IN people: missing phone handling add, BlueSen social fix

chrisyamas:
Issue that sparked this investigation was fixed by an edit to a member's page on the legislative site.

However, two unrelated `AttributeError: 'NoneType' object has no attribute 'groups'` issues were discovered to be causing failures in running DemocraticSenate.

1. First on `phone1, phone2 = re.search(r"Phone:\s(\d{3}-\d{3}-\d{4})\s\|\s(.+)", phones).groups()`
  - Cause: One or more dem senate members have less than two phone numbers
  - Solution: Restructured how the scraper was ingesting the "Phone" data field and all other information in that section of the page (at this point that also includes "Legislative Assistant", "Email", and "Media Contact"), including addition of a custom exemption to be raised (and named accordingly) when `NewDetailFieldEncountered` because the legislative site has added/modified a data field.
2. Second on `re.search(r"https://(www\.)?facebook\.com/(.+)", fb).groups()[1].rstrip("/")`
  - Cause: Approach of accessing each social media href by indexing as in `fb = CSS("div .fusion-social-links a").match(self.root)[1].get("href")` assumed same order of social media. This failed when one member had an instagram icon/link between icons for twitter and facebook.
  - Solution: Iterate through list of hrefs encountered by CSS selector, using dictionary keys for each potential social media platform, and storing any handles grabbed as values in the dictionary.

### DIFF
--- a/scrapers_next/in/people.py
+++ b/scrapers_next/in/people.py
@@ -46,37 +46,30 @@ class BlueSenDetail(HtmlPage):
             .strip()
         )
 
-        social_dict = {
-            "facebook": [],
-            "instagram": [],
-            "twitter": [],
-            "youtube": [],
-        }
+        p.capitol_office.address = addr
 
-        for soc in social_dict.keys():
-            soc_pattern = re.compile(rf"(.+)({soc}\.com/)(.+)")
-            social_dict[soc].append(soc_pattern)
+        socials = ["facebook", "instagram", "twitter", "youtube"]
+        handles = {x: None for x in socials}
+        patterns = {x: re.compile(rf"(.+)({x}\.com/)(.+)") for x in socials}
 
         social_links = CSS("div .fusion-social-links a").match(self.root)
         for link in social_links:
             href = link.get("href").lower()
-            for soc in social_dict.keys():
+            for soc in socials:
                 if soc in href:
-                    pattern = social_dict[soc][0]
+                    pattern = patterns[soc]
                     raw_handle = pattern.search(href).groups()[-1]
                     handle = re.sub("/", "", raw_handle)
-                    social_dict[soc].append(handle)
+                    handles[soc] = handle
 
-        p.capitol_office.address = addr
-
-        if len(social_dict["facebook"]) > 1:
-            p.ids.facebook = social_dict["facebook"][1]
-        if len(social_dict["instagram"]) > 1:
-            p.ids.instagram = social_dict["instagram"][1]
-        if len(social_dict["twitter"]) > 1:
-            p.ids.twitter = social_dict["twitter"][1]
-        if len(social_dict["youtube"]) > 1:
-            p.ids.youtube = social_dict["youtube"][1]
+        if handles["facebook"]:
+            p.ids.facebook = handles["facebook"]
+        if handles["instagram"]:
+            p.ids.instagram = handles["instagram"]
+        if handles["twitter"]:
+            p.ids.twitter = handles["twitter"]
+        if handles["youtube"]:
+            p.ids.youtube = handles["youtube"]
 
         return p
 
--- a/scrapers_next/in/people.py
+++ b/scrapers_next/in/people.py
@@ -1,4 +1,4 @@
-from spatula import HtmlListPage, XPath, CSS, URL, HtmlPage  # , SelectorError
+from spatula import HtmlListPage, XPath, CSS, URL, HtmlPage
 from openstates.models import ScrapePerson
 import re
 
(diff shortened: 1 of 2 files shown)


scrapers_next/in/people.py
```diff
@@ -20,9 +20,19 @@ def process_page(self):
         phones = (
             CSS("div .fusion-text.fusion-text-2 p").match(self.root)[1].text_content()
         )
-        phone1, phone2 = re.search(
-            r"Phone:\s(\d{3}-\d{3}-\d{4})\s\|\s(.+)", phones
-        ).groups()
+
+        try:
```
johnseekins, line 24: Maybe something akin to:
```
phone_numbers = re.search(
                r"Phone:\s(\d{3}-\d{3}-\d{4})\s\|\s(.+)", phones
            ).groups()
```
And then we can look at the size of the list returned? Seems cleaner than nested `try` blocks. e.g.

```
phone1 = None
phone2 = None
if len(phone_numbers) == 1:
  phone1 = phone_numbers[0]
if len(phone_numbers) > 1:
  phone1 = phone_numbers[0]
  phone2 = phone_numbers[1]
```

johnseekins: Could even skip the local variable declaration and add the actual values if they are found...

chrisyamas: Your feedback led me to rethink how the entire section of details was being grabbed, and [I rewrote that in my most recent commit.](https://github.com/openstates/openstates-scrapers/pull/4233/commits/ed72a5d681efad647a8ed541062edfc62cbf383a#diff-9eae9c00f2832ba4d89c6294f928772d6d554299758b67c70db01777d9283412R19-R40)

This new solution is able to store however many phones get added to that field, while also handling the Legislative Assistant, Media Contact, and Email fields all from the same CSS Selector result. This, with the addition of the custom exemption `class NewDetailFieldEncountered(BaseException)` will help us to catch if any modifications or additions happen with these fields.
- Additional win: in actually looking for fully at the whole section, I realized the previous code totally forgot to store the member's email (was only storing the email for their media contact) so this solution grabs that data too.

scrapers_next/in/people.py
```diff
@@ -12,48 +16,60 @@ def process_page(self):
...
+            "instagram": None,
+            "twitter": None,
+            "youtube": None,
+        }
+        social_links = CSS("div .fusion-social-links a").match(self.root)
+        for link in social_links:
+            href = link.get("href").lower()
+            for soc in social_dict.keys():
+                if soc in href:
+                    raw_handle = re.search(rf"(.+)({soc}\.com/)(.+)", href).groups()[-1]
```
johnseekins, line 59: Where possible, we should pre-compile regexes. It's good practice and is _slightly_ faster. Particularly in loops.

chrisyamas: Great catch! My newest commit now precompiles the regex patterns for both [phone](https://github.com/openstates/openstates-scrapers/pull/4233/commits/8160a417548f9091286faa87c41610a294b0c520#diff-9eae9c00f2832ba4d89c6294f928772d6d554299758b67c70db01777d9283412R20-R28) and [social media](https://github.com/openstates/openstates-scrapers/pull/4233/commits/8160a417548f9091286faa87c41610a294b0c520#diff-9eae9c00f2832ba4d89c6294f928772d6d554299758b67c70db01777d9283412R20-R28).

scrapers_next/in/people.py
```diff
@@ -46,30 +47,36 @@ def process_page(self):
...
 
-        if social_dict["facebook"]:
-            p.ids.facebook = social_dict["facebook"]
-        if social_dict["instagram"]:
-            p.ids.instagram = social_dict["instagram"]
-        if social_dict["twitter"]:
-            p.ids.twitter = social_dict["twitter"]
-        if social_dict["youtube"]:
-            p.ids.youtube = social_dict["youtube"]
+        if len(social_dict["facebook"]) > 1:
```
johnseekins, line 72: This is a bit confusing. Will there always be at least one value in each key? Shouldn't they be empty unless you find something?

chrisyamas: Totally. I was handling it all with one dictionary with `"social media": []` where the first element of the list was always the precompiled regex pattern for that social media platform, and the second element would user handle for that social media platform (if encountered).

But I went ahead in my newest commit and separated things out on lines 51-53 into:
- A list of the social media we're looking for
- A dictionary to store any handles we find
- A dictionary to store the regex patterns

With this change,[ the code is much simpler / more pythonic now](https://github.com/openstates/openstates-scrapers/pull/4233/commits/5f77c3a7c225632f85267946108f8646614db4a8#diff-9eae9c00f2832ba4d89c6294f928772d6d554299758b67c70db01777d9283412R51-R60).

NewAgeAirbender (approved):
Nice work, like your approach to the social handles issue!

PR #4238, opened 2022-12-14
NV people: new process for parsing/storing extra details

chrisyamas:
NV people scraper was failing because indexing approach to identifying each data field was reliant on consistency across all members' detail rows (but the number/types of these fields differed throughout the table).
- Error: `IndexError: list index out of range` on line 58: `leg_bldg_phone = extra_info_detail[3].text_content().strip()`
- Solution: Iterate through details fields, using conditionals to store details in person object according to the field name / label on the site.
  - Also added custom exception to be raised (`NewDetailFieldEncountered(BaseException)`). If the legislative site adds any other detail fields, or changes the names of any of the current labels, this exception will be raised so scraper can be updated accordingly.

### DIFF
--- a/scrapers_next/nv/people.py
+++ b/scrapers_next/nv/people.py
@@ -3,8 +3,13 @@ from openstates.models import ScrapePerson
 import re
 
 
+class NewDetailFieldEncountered(BaseException):
+    pass
+
+
 class Legislators(HtmlListPage):
     selector = CSS("tbody tr.thisRow.listRow")
+    field_names = set()
 
     def process_item(self, item):
         name_title = XPath(".//td/span/a/text()").match(item)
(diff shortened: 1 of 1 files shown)


scrapers_next/nv/people.py
```diff
@@ -3,8 +3,13 @@
 import re
 
 
+class NewDetailFieldEncountered(BaseException):
+    pass
+
+
 class Legislators(HtmlListPage):
     selector = CSS("tbody tr.thisRow.listRow")
+    field_names = set()
```
chrisyamas, line 12: This was accidentally left in. I will push up a commit to remove this set creation.

NewAgeAirbender: Just gotta do that & looks good

NewAgeAirbender (approved):

PR #4240, opened 2022-12-14
IA people: new process for parsing/storing extra details

chrisyamas:
IA people scraper was failing because the XPath selector was encountering zero items when a member detail page had no content in the "Other Information" section.
- Error: `spatula.selectors.SelectorError: XPath(//div[@class='legisIndent divideVert']//tr) on <html lang='en'> @ line 2 got 0 results, expected at least 1`
- Solution: Changed XPath selector to target the parent `<table>` which exists in section even when devoid of display content. If table contains more than just a `<style>` tag, code now iterates through the rows with details fields, and uses conditionals to store details in the person object according to the field name / label on the site.
  - Also added custom exception to be raised (NewDetailFieldEncountered(BaseException)). If the legislative site adds any other detail fields, or changes the names of any of the current labels, this exception will be raised so scraper can be updated accordingly.
  - This was also recently added to the NV people scraper. We may want to make this a global exception within the codebase, because it can be useful to identify how a webpage has changed when a scraper fails.

### DIFF
--- a/scrapers_next/ia/people.py
+++ b/scrapers_next/ia/people.py
@@ -69,7 +69,7 @@ class LegDetail(HtmlPage):
                 if len(row) > 1:
                     raw_field_name = row[0].text_content().strip().lower()
                     field_name = re.sub(":", "", raw_field_name)
-                    field_text = row[0].text_content().strip()
+                    field_text = row[1].text_content().strip()
 
                     extra_fields = {
                         "home email",
--- a/scrapers_next/ia/people.py
+++ b/scrapers_next/ia/people.py
@@ -1,6 +1,14 @@
 import attr
+import re
 from spatula import HtmlPage, HtmlListPage, CSS, XPath, SelectorError
 from openstates.models import ScrapePerson
+
+
+def get_field(field):
+    if field.endswith(":"):
+        return field[:-1]
+    else:
+        return field
 
 
 @attr.s(auto_attribs=True)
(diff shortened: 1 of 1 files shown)


scrapers_next/ia/people.py
```diff
@@ -53,21 +59,40 @@ def process_page(self):
...
-            p.extras[fields[i].lower()] = extra[i].strip()
+        )[0]
+
+        if len(other_info.getchildren()) > 1:
+            info_rows = XPath("//tr").match(other_info)
+            for row in info_rows:
+                if len(row) > 1:
+                    raw_field_name = row[0].text_content().strip().lower()
+                    field_name = re.sub(":", "", raw_field_name)
+                    field_text = row[0].text_content().strip()
```
NewAgeAirbender, line 72: There's one issue with this line that's evident looking at the scraped data

NewAgeAirbender (changes requested):
There's one slight change that needs to be made, but otherwise looks good!

scrapers_next/ia/people.py
chrisyamas: Indexing [fixed](https://github.com/openstates/openstates-scrapers/pull/4240/commits/f392aad64f0976b08a3e729b487e813f4ec31cf1#diff-5a784d04c8f85a0f44a3f26120c302cbb53f79a668d824d66fb8c6f45dc4e443R72) on that line.

NewAgeAirbender (approved):

PR #4258, opened 2022-12-28
NH: exception handling for bad redirects, adding comments, cleaning up variable names

sroomf:
I forgot to check what branch I was working on before I wrote this :/ . Sorry for the misleading branch name! These are the changes for the **NH committee scraper** . Fixes related to [OS-875](https://civic-eagle.atlassian.net/browse/OS-875) .

### DIFF
--- a/scrapers_next/nh/committees.py
+++ b/scrapers_next/nh/committees.py
@@ -9,37 +9,42 @@ class HouseCommitteeDetail(HtmlPage):
     example_input = "Public Works and Highways"
 
     def process_page(self):
-        try:
-            com = self.input
-            roles = XPath("//*[@id='form1']/div/div/div/div/div[1]/text()").match(
-                self.root
-            )
-            chair_mem = (
-                CSS("#form1 div div div div div a")
-                .match(self.root)[0]
-                .text_content()
-                .strip()
-            )
-            chair_role = roles[0].replace(":", "").strip()
-            # Some committees do not have a member assigned to Chair, so we do not add them
-            if "n/a" not in chair_role:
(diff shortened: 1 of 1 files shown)


scrapers_next/nh/committees.py
```diff
@@ -9,24 +9,37 @@ class HouseCommitteeDetail(HtmlPage):
...
-        com.add_member(Chair_mem, Chair_role)
-        VChair_mem = (
-            CSS("#form1 div div div div div a")
-            .match(self.root)[1]
-            .text_content()
-            .strip()
-        )
-        VChair_role = Rolez[1].replace(":", "").strip()
-        com.add_member(VChair_mem, VChair_role)
+        try:
```
johnseekins, line 12: Generally speaking, it's considered bad practice to have big chunks of code in a single try block. It can unintentionally suppress some errors and make handling processing more complex.

I see why you're doing it here, and this comment shouldn't block merging, but it's worth seeing if there isn't a simpler test for the redirect.

sroomf: Thanks for the input, John! Yeah, that makes sense. I've changed it so that only what's necessary is in the try block. Do you think this is the right fix or should I be looking for solutions outside of using a try/except?

johnseekins: `try` is a totally valid (and encouraged) tool in python. General practice just discourages wrapping large amounts of data in a single try (although wrapping a _function call_ in a `try` block is okay...)

NewAgeAirbender (approved):

PR #4259, opened 2022-12-28
Updated ND People Scraper for 2023 Session

bfossen-ce:
The scraper was set to the 2021-2022 session and the 2023-2024 members are now available


chrisyamas (approved):
Great, thanks for updating it!

PR #4260, opened 2022-12-28
simplify ut event loop and bump openstates dep

johnseekins:


showerst:
@johnseekins once a few events with bills in the new session load and we can confirm it's good, we can probably safely kill that commented out block about bills; seems like that same data is now consistently in both endpoints.

PR #4269, opened 2022-12-29
Fix for HJR that includes HJR Z

bfossen-ce:
The previous solution never correctly evaluated to Z or ZZ

### DIFF
--- a/scrapers/mi/bills.py
+++ b/scrapers/mi/bills.py
@@ -18,11 +18,10 @@ def jres_id(n):
 
     Example: n = 27, we go to the else clause and 27 mod 26 equals 1 so AA is returned
     """
-    # Offset by 1 to account for zero
-    n = n + 1
+
     # This if clause accounts for Joint Resolution IDs from HJR A to HJR Z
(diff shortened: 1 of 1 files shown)


scrapers/mi/bills.py
```diff
@@ -13,12 +13,13 @@
...
+    # Offset by 1 to account for zero
+    n = n + 1
     if n <= 26:
         return chr(ord('@') + n)
-    elif n > 26:
-        return chr(ord('@') + n % 26) + chr(ord('@') + n % 26)
     else:
-        return n
+        offset = n % 26 if n % 26 == 0 else 26
+        return chr(ord('@') + offset) + chr(ord('@') + offset)
```
johnseekins, line 22: Probably worth documenting the logic here a bit. May be difficult for someone to track through this later.

bfossen-ce: Good point, I will add in some explanation here and an example

johnseekins (approved):

NewAgeAirbender (approved):

PR #4271, opened 2022-12-29
In event fix

johnseekins:


showerst:
@johnseekins i noticed you killed the cloudflare headers here, did you test that change on any servers? Last session it worked fine from my home IP but bombed out when run on AWS/Hetzner, because those triggered the cloudflare challenge.

johnseekins:
We're actually looking at just completely re-doing this with the API. I'm gonna close this attempt.

PR #4276, opened 2022-12-30
CA people: AssemblyList fixes for img + office handling

chrisyamas:
Scraper was failing when running `AssemblyList` class. The following are the errors and solutions:

1. Data validation error for `image` attribute of person object
   - Issue: When img src wasn't present on page, a `None` type was being stored when only a `str` type is valid.
   - Solution: Added `or ""` to store an empty string when `phone_url` is `None` type.
2. Capitol and district office handling was experiencing selector errors and was not storing all offices listed
   - Solution: Capitol and district office address and phone handling entirely rewritten to utilize regex pattern matching.


bfossen-ce (approved):
LGTM!

PR #4295, opened 2023-01-03
OH: Prep 2023

NewAgeAirbender:
No bills or resolutions yet, still checking
https://search-prod.lis.state.oh.us/solarapi/v1/general_assembly_135/bills


showerst:
I had to do some other changes to get the scraper going so I created a separate PR, metadata is the same.

PR #4314, opened 2023-01-04
IL committee scraper

sroomf:
Committee scraper for IL - associated with [OS-463](https://civic-eagle.atlassian.net/browse/OS-463). 

Specific concern: I wasn't sure if it was necessary to have two separate HouseDetail and SenateDetail classes if they both did the same thing. I currently have both HouseList and SenateList pointing to a singular ChamberDetail class to grab membership information. Let me know if you'd like me to change this (or anything else)!

### DIFF
--- a/scrapers_next/il/committees.py
+++ b/scrapers_next/il/committees.py
@@ -5,20 +5,39 @@ from openstates.models import ScrapeCommittee
 class SenDetail(HtmlPage):
     def process_page(self):
         com = self.input
-        try:
-            members = CSS(".notranslate").match(self.root)
+        # the member page for joint commissions is slightly different
+        if list(com)[1][1] != "upper":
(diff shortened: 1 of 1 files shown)


scrapers_next/il/committees.py
```diff
@@ -0,0 +1,264 @@
...
+                com.add_source(self.source.url, note="homepage")
+                com.add_source(detail_link)
+
+                return JointCommissionDetail(com, source=detail_link)
+
+        else:
+            self.skip()
+
+
+"""
```
NewAgeAirbender, line 110: Can remove this chunk

NewAgeAirbender (approved):
One change but looks pretty good! The committee names are inconsistent, but I'll blame the source on that one.

PR #4315, opened 2023-01-04
WA bills: update to action_text to include text of action links

chrisyamas:
Scraper updated to include text of links for each action in the `action_text` stored as value of `description` attribute in bill object's `add_action()` method.

**Example using prior session bill:**
- Before this commit, [HB 1841 (from the 2021-22 session)](https://app.leg.wa.gov/billsummary?BillNumber=1841&Initiative=false&Year=2021) had it's second `action_text` stored as:
   - "First reading, referred to Finance."
- With this commit, that `action_text` will include the text of the link associated with it and be stored as:
   - "First reading, referred to Finance. (View Original Bill)"

*Note*: the scope of this commit was limited to grabbing the text. A future commit will get the `href` for the links, so that they can be also stored during the execution of bill object's `add_action()` method.

### DIFF
--- a/scrapers/wa/bills.py
+++ b/scrapers/wa/bills.py
@@ -447,16 +447,14 @@ class WABillScraper(Scraper, LXMLMixin):
 
             rows = header.xpath("following-sibling::div[1]/div")
             for row in rows:
+                # skip later lines that are just links to files
                 if row.xpath("div[1]")[0].text_content().strip() != "":
                     action_day = row.xpath("div[1]")[0].text_content().strip()
-                # skip later lines that are just links to files
-                action_text = (
-                    row.xpath("div[2]")[0]
-                    .text_content()
-                    .strip()
-                    .split("\r\n")[0]
-                    .strip()
-                )
+
+                raw_action = row.xpath("div[2]")[0].text_content()
+                action_text = " ".join(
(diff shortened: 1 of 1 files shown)


bfossen-ce (approved):
LGTM

scrapers/wa/bills.py
```diff
@@ -447,16 +447,14 @@ def scrape_actions(self, bill, chamber, bill_url, prefile_year, second_year):
 
             rows = header.xpath("following-sibling::div[1]/div")
             for row in rows:
+                # skip later lines that are just links to files
                 if row.xpath("div[1]")[0].text_content().strip() != "":
```
johnseekins, line 451: `if not`?

chrisyamas: Good catch, I didn't go looking for syntax improvements of the original code during my initial commit. I'll give it another look and push a new commit with changes.

chrisyamas: [New commit](https://github.com/openstates/openstates-scrapers/pull/4315/commits/c20c240503b1354d3845ffb5640f9bb9fdd31181) with change to this and other logic, syntax, and regex improvements (including following your prior guidance about pre-compiling all regular expressions).

NewAgeAirbender (approved):

PR #4319, opened 2023-01-05
OH people: fixes for changes on member lists, senate details pages

chrisyamas:
Scraper was failing on multiple CSS selectors when running for either House or Senate. This was due to several class name changes of divs on the list pages of members.

Given the identical nature of the list pages, this commit consolidated the two different object types being used to handle each chamber's list page into one. The `House` and `Senate` class each now extend the `LegList` object.

Within `LegList`, this solution fixes the class names targeted by the CSS selectors.
- Additional improvements:
   - tweak to conditional skipping of vacant seats
   - addition of `list_page_url` as attribute of `LegPartial` object passed to `LegDetail`. Prior code was twice storing the member detail page in the ScrapePerson object, while failing to store the member list page url. This allowed that to be rectified in `LegDetail` changes (including adding the optional note within `add_source()` method indicating which source the url points to).

With `LegDetail`, scraper was still failing for senate because the info bar containing capitol address and phone was moved to each member's Contact page.
- This commit gets that data from each senator's contact page, and uses regex patterns to fix spacing issues in each senator's address.

### DIFF
--- a/scrapers_next/oh/people.py
+++ b/scrapers_next/oh/people.py
@@ -2,8 +2,14 @@ import re
 import attr
 from spatula import HtmlListPage, HtmlPage, CSS, URL, SelectorError
 from openstates.models import ScrapePerson
+import requests
+import lxml.html
 
 background_image_re = re.compile(r"background-image:url\((.*?)\)")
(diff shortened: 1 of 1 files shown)


NewAgeAirbender (approved):

bfossen-ce (approved):
LGTM

PR #4322, opened 2023-01-06
Updated Hardcoded Value for WI People Scraper

bfossen-ce:
Updated Hardcoded Value for WI People Scraper for 2023 Session


chrisyamas (approved):
The sources work when testing in the browser, and thanks for changing in the commented example code as well. Looks good!

PR #4323, opened 2023-01-06
Updated VT url for 2023-2024 Session

bfossen-ce:
Updated VT url for 2023-2024 Session


chrisyamas (approved):
Just ran it with the change, and it's all good!

PR #4327, opened 2023-01-06
NE people: new url format for detail pages, new list url storage

chrisyamas:
Scraper was failing with HTTPError due to requests of member district page urls that are inconsistently maintained by host site. This commit replaced inconsistent url format with consistent one.

Error example:
- `scrapelib.HTTPError: 410 while retrieving http://news.legislature.ne.gov/dist04/`

Solution:
- Retrieves member landing pages (i.e. http://nebraskalegislature.gov/senators/landing-pages/index.php?District=04) instead

Additional:
- the url for the member list page was not being stored in each person object
   - this commit stores list page in person object's `"sources"` collection, with note added indicating which source is the list page of members and which is member detail page

### DIFF
--- a/scrapers_next/ne/people.py
+++ b/scrapers_next/ne/people.py
@@ -8,6 +8,7 @@ class LegPartial:
     name = attr.ib()
     district = attr.ib()
     url = attr.ib()
+    member_list_url = attr.ib()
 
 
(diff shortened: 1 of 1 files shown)


bfossen-ce (approved):
LGTM!

PR #4332, opened 2023-01-06
MT Events: handle duplicate documents better

johnseekins:


showerst:
@johnseekins you should be able to do 
```
event.add_document(... , on_duplicate="ignore") 
```

and that will let you avoid keeping a set and checking for it.

PR #4334, opened 2023-01-07
VA: committee scraper

sroomf:
Committee scraper for Virginia. Associated with [OS-874](https://civic-eagle.atlassian.net/browse/OS-874). Put in some comments to explain some weird things, but let me know if I was too excessive.


scrapers_next/va/committees.py
```diff
@@ -0,0 +1,149 @@
...
+from openstates.models import ScrapeCommittee
+import time
+
+# VA lists the full names of committee members on individual separate pages
+# MemberDetail grabs a member's full name from their specific page
+
+
+class MemberDetail(HtmlPage):
+    def process_page(self):
+        com = self.input[0]
```
johnseekins, line 11: Might be worth handling when `input` _isn't_ a >= 2 item list.

scrapers_next/va/committees.py
```diff
@@ -0,0 +1,149 @@
...
+        role = self.input[1]
+        try:
+            mem_name = CSS("#mainC > h3").match(self.root)
+        except SelectorError:
+            return com
+
+        if "Delegate" in mem_name[0].text:
+            cleaned_name = mem_name[0].text.split("Delegate")[1].strip()
+        else:
```
johnseekins, line 20: There's always a possibility that `Delegate` or `Senator` _won't_ be in the string. Perhaps `elif "Senator" in mem_name[0].text:`?

scrapers_next/va/committees.py
```diff
@@ -0,0 +1,149 @@
...
+            cleaned_name = mem_name[0].text.split("Senator")[1].strip()
+
+        if cleaned_name is not None:
```
johnseekins, line 23: Should this be explicitly calling out `None` or be the more generic `if not cleaned_name` (which covers empty strings as well as `None`)?

scrapers_next/va/committees.py
```diff
@@ -0,0 +1,149 @@
...
+    def process_item(self, item):
+        chamber_text = item.getparent().getparent().getparent().getchildren()[0].text
+        comm_name = item.text
+
+        # this is hacky but I couldn't figure out another way to ignore a lack of subcommittees
+        if "SENATE" in chamber_text:
+            if comm_name == "Transportation":
+                raise SkipItem("no subcommittees")
+            if comm_name == "Rules":
+                time.sleep(4)
```
johnseekins, line 144: Un-commented `sleep` statements are usually a bad sign. Why do we need to sleep for 4 seconds here?

scrapers_next/va/committees.py
```diff
@@ -0,0 +1,149 @@
...
+                raise SkipItem("no subcommittees")
+
+        detail_link = item.get("href")
+        time.sleep(4)
```
johnseekins, line 148: Again, we should either clarify _why_ we're sleeping or remove the `sleep` statement.

scrapers_next/va/committees.py
sroomf: good point! will change

scrapers_next/va/committees.py
sroomf: ok! sleep statements helped the scraper run more smoothly (was getting a lot of SelectorErrors or source errors which I believe comes from the VA site being a little slow). I can add that in as a comment, unless you think I should be doing something else!

johnseekins: I think you want `SCRAPELIB_RPM`:
```
scrapers/ca/__init__.py:settings = dict(SCRAPELIB_RPM=30)
scrapers/ct/__init__.py:settings = {"SCRAPELIB_RPM": 20}
scrapers/ia/__init__.py:settings = {"SCRAPELIB_RPM": 30}
scrapers/ks/__init__.py:settings = dict(SCRAPELIB_TIMEOUT=300, SCRAPELIB_RPM=12)
scrapers/nd/__init__.py:    "SCRAPELIB_RPM": 10,
scrapers/pa/__init__.py:settings = {"SCRAPELIB_RPM": 30}
scrapers/va/__init__.py:settings = {"SCRAPELIB_RPM": 40}
scrapers/vt/__init__.py:settings = dict(SCRAPELIB_RPM=20)
```
It's built-in rate-limiting that slows down requests to remote sites. Although there may be a similar function in spatula?

johnseekins: The point is that having random sleeps in a code base makes debugging inconsistent and can unintentional suppress (or create) problems that are difficult to find.

scrapers_next/va/committees.py
sroomf: Ok! I'm going to throw a ValueError if there isn't a 2 item list. This part of the code should only get called with a 2 item list (line 68) so I feel like if the two items don't make it into this portion that's...not good.

scrapers_next/va/committees.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class FindSubCommittees(HtmlListPage):
+
+    selector = CSS("#mainC > ul:nth-child(12) > li a")
+
+    def process_item(self, item):
+        try:
+            comm_name = item.text
+
```
johnseekins, line 77: Probably don't need this empty line...

scrapers_next/va/committees.py
```diff
@@ -0,0 +1,149 @@
...
+        com = ScrapeCommittee(
+            name=comm_name,
+            classification="committee",
+            chamber=chamber,
+        )
+
+        detail_link = item.get("href")
+        com.add_source(self.source.url, note="homepage")
+        com.add_source(detail_link)
+        time.sleep(3)
```
johnseekins, line 126: As noted elsewhere in this code base, `sleep`s should be at least justified...

scrapers_next/va/committees.py
sroomf: Gotcha. spatula has a way to set requests per minute from the command line. i'll take out the sleep statements and use that instead

NewAgeAirbender:
I've tried running the scraper to test it out, and have just been getting a bunch of files containing `null`. May want to revisit this, happy to chat about it

NewAgeAirbender (changes requested):
I've tried running the scraper to test it out, and have just been getting a bunch of files containing `null`. May want to revisit this, happy to chat about it

sroomf:
> 

@NewAgeAirbender added in a comment at the top of the file talking about how this source isn't super great and that we should try to find a more stable source next time around. The version of the scraper in this PR right now is the version that was run to get the data we currently have. Based on our convo last night, it sounds like we should merge this. I'm going to create a draft PR with the start of the scraper that uses the other site so we can refer to that when we eventually revisit/re-write this scraper.

NewAgeAirbender (approved):

PR #4336, opened 2023-01-07
refactor: use official API for Indiana

elseagle:
A refactor on the INDIANA events scraper to use the official API because we were getting blocked  by cloudflare

### DIFF
--- a/scrapers/in/events.py
+++ b/scrapers/in/events.py
@@ -86,9 +86,8 @@ class INEventScraper(Scraper):
             event.add_source(url)
             event.add_participant(chamber, type="committee", note="host")
             event.add_media_link("Video of Hearing", video_url, media_type="text/html")
-
+            agenda = event.add_agenda_item("Bills under consideration")
             for bill in extra_details["agenda"]:
-                agenda = event.add_agenda_item("Bills under consideration")
 
                 if bill.get("bill"):
                     bill_id = bill.get("bill").get("billName")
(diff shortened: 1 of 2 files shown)


scrapers/in/events.py
```diff
@@ -1,90 +1,97 @@
...
 from openstates.scrape import Scraper, Event
+from .utils import add_space
+
+log = logging.getLogger(__name__)
 
 
 class INEventScraper(Scraper):
     _tz = pytz.timezone("America/Indianapolis")
     # avoid cloudflare blocks for no UA
     cf_headers = {
```
johnseekins, line 17: I don't think we need these headers now that we aren't using the direct site any more.

scrapers/in/events.py
```diff
@@ -1,90 +1,97 @@
...
+            link = meeting["link"]
+            _id = link.split("/")[-1]
 
-            time_and_loc = meta.xpath("span/text()")[0].strip()
-            time_and_loc = time_and_loc.split("\n")
-            time = time_and_loc[0]
-            loc = time_and_loc[1]
+            extra_details = self.in_request(f"{self.base_url}{link}").json()
 
-            if loc == "":
```
johnseekins, line 46: You may want to hang on to this if statement (after you define the location) as occasionally jurisdictions will create meetings without locations.

scrapers/in/events.py
elseagle: We need it, without it we get a 403 forbidden url error

scrapers/in/events.py
elseagle: Understood, this is fixed. Added one more possible option to get the location

johnseekins (approved):

PR #4337, opened 2023-01-08
Add devcontainer

jamesturk:
This creates a devcontainer that installs the necessary dependencies locally, this allows JSON scrapes to be run from the container.  (This does not attempt to set up the database/etc. but it'd be theoretically possible to add a more complex devcontainer if desired.)


johnseekins:
Not to be overly critical, but how does this differ from (e.g.) `docker compose run scrape al bills --scrape`?

jessemortenson (approved):
Looks good to me!

PR #4338, opened 2023-01-09
Fix DE bill sponsorship links

johnseekins:
Delaware has added new sites for individual caucuses and is using those links to indicate sponsorship on bills. This change makes it possible to handle those new links.

### DIFF
--- a/scrapers/de/bills.py
+++ b/scrapers/de/bills.py
@@ -14,6 +14,12 @@ class DEBillScraper(Scraper, LXMLMixin):
     legislators = {}
     legislators_by_short = {}
     legislators_by_district = {}
+    potential_sponsor_urls = {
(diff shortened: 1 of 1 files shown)


scrapers/de/bills.py
```diff
@@ -276,8 +300,13 @@ def scrape_vote(self, bill, vote_id, session):
 
             yield vote
 
-    def add_sponsor_by_legislator_id(self, bill, legislator_id, sponsor_type):
-        sponsor = self.legislators[str(legislator_id)]
+    def add_sponsor_by_legislator_id(
+        self, bill, legislator_id, sponsor_type, sponsor_key="PersonId"
```
johnseekins, line 304: Adding `sponsor_key` with a default value preserves the old functionality pattern.

NewAgeAirbender (approved):

PR #4341, opened 2023-01-09
NH people: HouseDetail class creation, cleaning up old code

chrisyamas:
NH people scraper lacked any scraper of House member detail pages. This PR resolves that issue.

Primary contributions of this PR include:
1. Creation of `HouseDetail` class
2. Creation of helper function `scrape_house_vals()`
   - returns a dictionary of values needed to construct the url for each House member detail page
   - is called in `Legislators` class to provide the source attribute for `HouseDetail` object initializations
3. Removal of unnecessary, redundant, or incorrect scraping and text processing
   - especially seen in `SenateDetail` class, as contact details that class was getting had already been collected from the CSV list page in `Legislators` class
   - also identified the type of addresses being stored for each chamber's members, and added comment with explanation
4. Editing of existing logical operators and code syntax to be more Pythonic
   - i.e. `if county != "":` changed to `if county:`
5. Addition of `TODO` comment suggesting member detail pages be checked in the future for addition of useful/desirable data
   - currently, the member image `src` is the only relevant data on detail pages that isn't in CSV list page for most House and all Senate members

### DIFF
--- a/scrapers_next/nh/people.py
+++ b/scrapers_next/nh/people.py
@@ -1,38 +1,80 @@
 import io
 import csv
-from spatula import URL, CsvListPage, HtmlPage, CSS, XPath, SelectorError
+from spatula import URL, CsvListPage, HtmlPage, CSS
 from openstates.models import ScrapePerson
 import re
+import requests
+import lxml.html
+
+# Regex compile for isolating variety of address formats on house detail pages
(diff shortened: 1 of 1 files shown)


bfossen-ce (approved):
LGTM!

PR #4346, opened 2023-01-10
CO people: reduce min_items for member list selector

chrisyamas:
Scraper was failing because CSS selector for `LegList` had its `min_items` attribute set one number too high.

This commit reduces `min_items` from 101 to 100 (as there are 100 total members of CO legislature, with 35 Senators and 65 Representatives).


bfossen-ce (approved):
LGTM

PR #4356, opened 2023-01-11
chore: update openstates

elseagle:
update openstates to 6.13.9


johnseekins:
I'm a little concerned this branch was out of date with upstream when it was done. `openstates` should have gone from `6.13.8` -> `6.13.9`. Instead, it says `6.13.4` -> `6.13.9`. I'd like to see this branch pulled even before we merge.

PR #4365, opened 2023-01-12
GA people: add extras['city'] None handle, API/HTML source+link w notes

chrisyamas:
Scraper was failing with `AttributeError` when attempting to call `strip()` string method on member's `city` value when value was `None`.

Solution:
- one-line conditional to save empty string instead, when city value is `None`

Additional modification:
- `add_source()` now called only for API paths that are used to scrape list of members and member detail data
- `add_link()` now used to save the HTML page paths for both chamber-specific member list page and member detail page


bfossen-ce (approved):
LGTM

PR #4368, opened 2023-01-13
NV people: add regex handling for member name parsing

chrisyamas:
Scraper was failing with `IndexError` during member name parsing, triggered by vacant seats.

Additionally, in analyzing the data, it was discovered that name suffixes (i.e. "III", "Jr.", "Esq.", etc.) were not being stored in the correct position within a name string.

This PR rewrote the code block for name parsing to (1) use regular expressions (instead of prior approach which relied on splitting and concatenating the name parts by index) and (2) have a conditional regex check -- and separate handling for parsing and reconstructing -- names with suffixes.

### DIFF
--- a/scrapers_next/nv/people.py
+++ b/scrapers_next/nv/people.py
@@ -11,12 +11,21 @@ class Legislators(HtmlListPage):
     selector = CSS("tbody tr.thisRow.listRow")
     field_names = set()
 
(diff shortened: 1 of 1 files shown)


bfossen-ce (approved):
Good catch on the name with suffixes

PR #4373, opened 2023-01-14
[WIP] VA: committee scraper source 2

sroomf:
WIP for another source for VA committee data. Potentially the source we want to use if we're rewriting the scraper.

### DIFF
--- a/scrapers_next/va/committees_csv.py
+++ b/scrapers_next/va/committees_csv.py
@@ -1,0 +1,37 @@
+from paramiko.client import SSHClient, AutoAddPolicy
+import paramiko
+import time
+from spatula import CsvListPage
+
+
+class ChamberList(CsvListPage):
+    client = SSHClient()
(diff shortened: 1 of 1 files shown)


scrapers_next/va/committees_csv.py
```diff
@@ -0,0 +1,37 @@
...
+    client = SSHClient()
+    client.set_missing_host_key_policy(AutoAddPolicy)
+    connected = False
+    attempts = 0
+    while not connected:
+        try:
+            client.connect(
+                "sftp.dlas.virginia.gov",
+                username="rjohnson",
+                password="E8Tmg%9Dn!e6dp",
```
johnseekins, line 17: I realize this password is likely already exposed elsewhere, but can we pull the password from an environment variable instead?

johnseekins: Something like `os.environ["VA_SFTP_PASSWORD"]` or something would be appropriate.

PR #4377, opened 2023-01-14
adding OR committee scraper

rezarzky:
Hi,

I added the OR committee scraper using the Spatula.

### DIFF
--- a/scrapers_next/or/committee.py
+++ b/scrapers_next/or/committee.py
@@ -1,0 +1,61 @@
+from spatula import XPath, URL, HtmlListPage, HtmlPage, CSS
+from openstates.models import ScrapeCommittee
+import re
+
+class CommitteeDetail(HtmlListPage):
+    source = URL("https://olis.oregonlegislature.gov/liz/2023R1/Committees/JCT/CommitteeMembers")
+    selector = CSS('tbody tr')
+
+    def process_item(self, item):
+        com = self.input
+        position, name = item.getchildren()
+        name = re.sub(r"Senator |Representative ", "", name.text_content().strip())
+        position = position.text_content().strip()
+        com.add_member(name, position)
+
+        return com
+
+
+class CommitteeList(HtmlListPage):
+    source = "https://olis.oregonlegislature.gov/liz/Committees/list/"
+
+    def process_item(self, item):
+        link = item.get("href").split('/')
+        link = 'https://olis.oregonlegislature.gov/liz/' + link[4] +\
+                "/Committees/" + link[6] + "/CommitteeMembers"
+
+        try:
+            parent = self.parent
+        except:
+            parent = None
+
+        com = ScrapeCommittee(
+                name=item.text_content(),
+                chamber=self.chamber,
+                classification=self.classification,
+                parent=parent,
+            )
+        com.add_source(link)
+
+        return CommitteeDetail(com, source=link,)
+
+
+class HouseCommitteeList(CommitteeList):
+    selector = CSS("#HouseCommittees_search li a")
+    chamber = "lower"
+    classification = "committee"
+
+class SenateCommitteeList(CommitteeList):
+    selector = CSS("#SenateCommittees_search li a")
+    chamber = "upper"
+
+class JointCommitteeList(CommitteeList):
+    selector = CSS("#SenateCommittees_search > li > a")
+    chamber = "legislature"
+
+class WayMeansCommitteeList(CommitteeList):
+    selector = CSS("#JointCommittees_search > li > ul > li > a")
+    chamber = "legislature"
+    parent = "Ways and Means"
+    classification = "subcommittee"
+
--- a/scrapers_next/or/committee.py
+++ b/scrapers_next/or/committee.py
@@ -1,18 +1,21 @@
-from spatula import XPath, URL, HtmlListPage, HtmlPage, CSS
+from spatula import URL, HtmlListPage, HtmlPage, CSS
 from openstates.models import ScrapeCommittee
 import re
 
-class CommitteeDetail(HtmlListPage):
-    source = URL("https://olis.oregonlegislature.gov/liz/2023R1/Committees/JCT/CommitteeMembers")
-    selector = CSS('tbody tr')
 
-    def process_item(self, item):
+class CommitteeDetail(HtmlPage):
+    example_source = URL(
+        "https://olis.oregonlegislature.gov/liz/2023R1/Committees/JCT/CommitteeMembers"
+    )
+
+    def process_page(self):
         com = self.input
-        position, name = item.getchildren()
-        name = re.sub(r"Senator |Representative ", "", name.text_content().strip())
-        position = position.text_content().strip()
-        com.add_member(name, position)
-
+        members = CSS("tbody tr").match(self.root)
+        for member in members:
+            position, name = member.getchildren()
+            name = re.sub(r"Senator |Representative ", "", name.text_content().strip())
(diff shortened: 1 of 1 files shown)


scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
...
+class CommitteeList(HtmlListPage):
+    source = "https://olis.oregonlegislature.gov/liz/Committees/list/"
+
+    def process_item(self, item):
+        link = item.get("href").split('/')
+        link = 'https://olis.oregonlegislature.gov/liz/' + link[4] +\
+                "/Committees/" + link[6] + "/CommitteeMembers"
+
+        try:
+            parent = self.parent
```
johnseekins, line 28: Do we need this assignment? It seems like all we're doing is assigning a class-level variable to a local variable, then adding it as an attribute to the `ScrapeCommittee` object. Could we skip these four lines and just say:

```
com = ScrapeCommittee(
    ...
    parent=self.parent,
    ...
```

johnseekins:
Really clean implementation. Nice work.

scrapers_next/or/committee.py
rezarzky: Previously, I tried to directly use parent=self.parent. However, I got an error because most of the committee doesn't have parent's variable

scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
+from spatula import XPath, URL, HtmlListPage, HtmlPage, CSS
+from openstates.models import ScrapeCommittee
+import re
+
```
chrisyamas, line 4: PEP 8 pythonic styling encourages two lines of space between a class and anything else. Would be good to add a space here and also more down near the bottom of the file to have two spaces between those other classes.

scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
...
+    selector = CSS('tbody tr')
+
+    def process_item(self, item):
+        com = self.input
+        position, name = item.getchildren()
+        name = re.sub(r"Senator |Representative ", "", name.text_content().strip())
+        position = position.text_content().strip()
+        com.add_member(name, position)
+
+        return com
```
chrisyamas, line 16: While it seems intuitive to use `HtmlListPage` here, this is currently returning the committee object (and consequently writing an additional JSON file) for every item processed on the page using the `process_item()` method. That means we end up with as many committee objects as there are total members across all committees.

Instead, you can simply reformat what you've already done into a `HtmlPage` class like this:
```
class CommitteeDetail(HtmlPage):
    example_source = URL("https://olis.oregonlegislature.gov/liz/2023R1/Committees/JCT/CommitteeMembers")

    def process_page(self):
        com = self.input
        members = CSS('tbody tr').match(self.root)
        for member in members:
            position, name = member.getchildren()
            name = re.sub(r"Senator |Representative ", "", name.text_content().strip())
            position = position.text_content().strip()
            com.add_member(name, position)
        return com
```

That should serve this purpose better. While the list page scraper is very helpful -- and a huge reason we love spatula -- details pages (i.e. your final destination for scraping each committee) are typically best left to a normal page type class.

scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
...
+    source = "https://olis.oregonlegislature.gov/liz/Committees/list/"
+
+    def process_item(self, item):
+        link = item.get("href").split('/')
+        link = 'https://olis.oregonlegislature.gov/liz/' + link[4] +\
+                "/Committees/" + link[6] + "/CommitteeMembers"
+
+        try:
+            parent = self.parent
+        except:
```
chrisyamas, line 29: It's good practice to specify the type of exception you're allowing for in your `except` clause.

In this case, when you don't have `parent` specified, you get an `AttributeError`. So you can make this line:

`except AttributeError:`

scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
...
+            parent = None
+
+        com = ScrapeCommittee(
+                name=item.text_content(),
+                chamber=self.chamber,
+                classification=self.classification,
```
chrisyamas, line 35: Rather than specifying classification as an attribute of the `CommitteeList` class, you can just rely on the fact that interdependency of `parent` and `classification` here. If there is not truthy value for `parent`, then it must be a committee and not a subcommittee.

Therefore you can rewrite this line as:
`classification="subcommittee" if parent else "committee",`

And then just get remove wherever you specified `classification` for classes at the bottom of the page.

scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
...
+            )
+        com.add_source(link)
+
+        return CommitteeDetail(com, source=link,)
+
+
+class HouseCommitteeList(CommitteeList):
+    selector = CSS("#HouseCommittees_search li a")
+    chamber = "lower"
+    classification = "committee"
```
chrisyamas, line 46: You can remove this.

scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
...
+
+class JointCommitteeList(CommitteeList):
+    selector = CSS("#SenateCommittees_search > li > a")
+    chamber = "legislature"
+
+class WayMeansCommitteeList(CommitteeList):
+    selector = CSS("#JointCommittees_search > li > ul > li > a")
+    chamber = "legislature"
+    parent = "Ways and Means"
+    classification = "subcommittee"
```
chrisyamas, line 60: You can remove this.

scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
...
+
+class SenateCommitteeList(CommitteeList):
+    selector = CSS("#SenateCommittees_search li a")
+    chamber = "upper"
+
+class JointCommitteeList(CommitteeList):
+    selector = CSS("#SenateCommittees_search > li > a")
+    chamber = "legislature"
+
+class WayMeansCommitteeList(CommitteeList):
```
chrisyamas, line 56: I'd just change the name of this class to `WaysMeansCommitteeList` (i.e. just add an "s" after the "Way") so that it is more accurate to the name of the parent committee it is named after.

scrapers_next/or/committee.py
```diff
@@ -0,0 +1,61 @@
...
+        except:
+            parent = None
+
+        com = ScrapeCommittee(
+                name=item.text_content(),
+                chamber=self.chamber,
+                classification=self.classification,
+                parent=parent,
+            )
+        com.add_source(link)
```
chrisyamas, line 38: Two things here:
1. We try to add a `note` attribute as much as possible here, so we know later in the data where the url is supposed to point to. In this case, you can make it `com.add_source(link, note="Committee Details Page")`.
2. It would be good to add an additional source for the actual list page that you are scraping from in this class. In a spatula page scraper like this you can access the url for the source page as `self.source.url`.
   - so adding that source here would look like `com.add_source(self.source.url, note="Committees List Page")`.

chrisyamas (changes requested):
Overall, this was great! The only functional sticking point is the issue with the `HtmlListPage` being used for the committee details/membership page.

Mostly everything else is comprising of simple or stylistic changes.

Nice work! Once you make those changes then I think we'll be good to go!

scrapers_next/or/committee.py
rezarzky: Thanks for the inputs, will change this one

chrisyamas (approved):

PR #4387, opened 2023-01-18
fix: NY committee scraper chair title

elseagle:


scrapers_next/ny/committees.py
```diff
@@ -50,21 +51,26 @@ def process_page(self):
...
 
         # in case there are co-chairs
-        num_chairs = len(chairs)
-
-        for chair in chairs:
+        for index, chair in enumerate(chairs):
             chair_name = CSS(".comm-chair-name").match_one(chair).text_content().strip()
+
+            # index for the current chair in the list of chairs
+            actual_index = index + 1
```
johnseekins, line 63: Since this is going to happen on every step, could we just say `{index + 1}` below and skip the extra variable?
And also maybe explain why we're iterating past our current index?

elseagle: Yeah, used that when I was testing, it broke somehow. I'll retry it again since everything is working good now. Will add the comment too

johnseekins (approved):

NewAgeAirbender (approved):

PR #4397, opened 2023-01-21
CA committees: rewrite of scraper after site changes

chrisyamas:
Scraper was failing for both Assembly and Senate, due to a variety of formatting changes and variations across pages on the site.

This PR provides a rewritten/reformatted committee scraper that can handle the substantial amount of HTML inconsistencies across CA's committee pages, while also being simpler and more class-based.

### DIFF
--- a/scrapers_next/ca/committees.py
+++ b/scrapers_next/ca/committees.py
@@ -1,260 +1,118 @@
-from spatula import HtmlListPage, URL, XPath, CSS, HtmlPage, SelectorError
+from spatula import HtmlListPage, URL, XPath, CSS, HtmlPage
 from openstates.models import ScrapeCommittee
 import re
+import logging
 
 
-class ChooseType(HtmlPage):
+member_name_role_re = re.compile(r"(.+)\((.+)\)")
+chair_name_re = re.compile(r"(Senator\s|Assembly\sMember\s*)(.+[a-z])")
+chair_title_re = re.compile(r"(.*Chair)")
+vacancy_re = re.compile(r"vacant|vacancy|vacancies")
+members_check_re = re.compile(r"(Senator|Assemblymember)\s+([A-Z].+)")
+
+
+class CommMembership(HtmlPage):
+    example_source = "https://aaar.assembly.ca.gov/membersstaff"
+
     def process_page(self):
+        committee = self.input
 
-        # this link was being considered type one, but should be type four
-        if self.source.url == "https://ajed.assembly.ca.gov":
-            return process_page_type4(self.root, self.input, self.source)
+        sen_page_check = self.root.xpath(".//div[@class='banner senate-red col-xs-12']")
 
-        if data := process_page_type1(self.root, self.input):
-            return data
-        elif data := process_page_type2(self.root, self.input):
-            return data
-        elif data := process_page_type3(self.root, self.input):
-            return data
-        return process_page_type4(self.root, self.input, self.source)
+        # Handles senate page type format
+        if sen_page_check:
+            members = []
+            a_tags_text_list = self.root.xpath(
+                ".//div[@class='field-item even']//a//text()"
+            )
+            for text in a_tags_text_list:
+                member_match = members_check_re.search(text)
+                if member_match:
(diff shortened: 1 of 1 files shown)


scrapers_next/ca/committees.py
```diff
@@ -287,43 +146,43 @@ def process_item(self, item):
...
-        com.add_link(comm_url, note="homepage")
-        return ChooseType(com, source=URL(comm_url))
+
+        com.add_source(self.source.url, note="Committee List Page")
+        com.add_source(comm_url, note="Committee Detail Page")
+
+        return CommDetails(com, source=URL(comm_url))
 
 
 class AssemblyCommitteeList(HtmlListPage):
```
bfossen-ce, line 156: You arent scraping any of the committees here. It looks like only subcommittees are being scraped for the Assembly. I see there are som committees with the same name like Public Safety in the Senate and Assembly but this one isnt listed as a joint committee. I am only getting the 7 subcomittees when I run `poetry run spatula scrape scrapers_next.ca.committees.AssemblyCommitteeList` Can you update the code or address this issue why the Assembly committees are not showing up if that is the desired outcome?

chrisyamas: Ah yes I see why that looks concerning. It's actually due to [this part of the code](https://github.com/openstates/openstates-scrapers/pull/4397/files#diff-d2282de25dd52de0c420d1d9e3ba2378a39f614b5b6ef2c0590624dbba51f483R164-R167) where all Joint Committees are skipped when running `AssemblyCommitteeList` so as to avoid duplicates (since they are ingested during `SenateCommitteeList()`). That's how it was written before, and I didn't see a reason to change it.

When I did the rewrite I also double-checked that they were definitely all the same committees (i.e. there aren't any joint committees listed on Assembly's page but not on Senate's page, and vice versa), and that is the case. So we're good there.

chrisyamas: This was bizarre to me, because the CSS selector was grabbing from the top of the list when I ran it last on Friday. Either I've lost my marbles, or the CA site adjusted styling: either of which are very possible.

Anyway, [here is the new selector path](https://github.com/openstates/openstates-scrapers/pull/4397/files/65c52f5361ab0e670b6e6b907154064597359749..5ddda75f5f180d8f90ba20c0841a7117fcb2a8d8#diff-d2282de25dd52de0c420d1d9e3ba2378a39f614b5b6ef2c0590624dbba51f483R157) which gets the whole list. I tested it again and it looks good. Lmk what you get!

bfossen-ce: My bet is CA updated their website, seems like a lot of updates are going on now. I just ran this and the results look good to me. You are correctly scraping the committees as I would expect.

bfossen-ce (approved):

scrapers_next/ca/committees.py
```diff
@@ -287,43 +145,43 @@ def process_item(self, item):
...
-        ):
-            self.skip()
-        elif comm_name.startswith("Subcommittee"):
-            parent_comm = item.getparent().getparent().getchildren()[0].text_content()
-            com = ScrapeCommittee(
-                name=comm_name,
-                classification="subcommittee",
-                chamber="lower",
-                parent=parent_comm,
+        if comm_name.startswith("Subcommittee"):
```
NewAgeAirbender, line 168: Some Assembly Subcommittees are getting scraped with the Parent committee of `Subcommittee No. 1 on Health and Human Services`

NewAgeAirbender (changes requested):
There's a slight change on the Parent selector for Assembly Subcommittees, but the rest looks good!

scrapers_next/ca/committees.py
chrisyamas: I have a feeling this was because of the selector change I made in my prior commit. However[ I just fixed the parent selection](https://github.com/openstates/openstates-scrapers/pull/4397/files#diff-d2282de25dd52de0c420d1d9e3ba2378a39f614b5b6ef2c0590624dbba51f483R170-R176) so we should be good now.

NewAgeAirbender (approved):

PR #4403, opened 2023-01-23
CO comm: selector change to handle task force w/o members

chrisyamas:
Committee scraper was failing with `SelectorError` on `CSS(.member)`.
- This was happening for just one committee that lacked membership: the "Task Force for the Consideration of Facial Recognition Services"

This PR solves this by changing to a standard `xpath` selector, and logging a `logging.warning` when a committee does not have a `<div>` with `class='member'`.


bfossen-ce (approved):
LGTM

PR #4404, opened 2023-01-23
AK comm: adjusted num_items down on CSS selector

chrisyamas:
Committee scraper was failing with error `spatula.selectors.SelectorError: CSS(div.area-frame ul.list li) on <html> @ line 3 got 110 results, expected 112`

This PR simply adjusts `num_items` attribute of the CSS selector down to match the number of items actually appearing on the page.


bfossen-ce (approved):
LGTM

PR #4415, opened 2023-01-24
2023 01 24 fixes

johnseekins:
Fix USA events (bad validation of `nomination` event type)
Fix IA events (bad grammar)
Fix MD events (empty committee names)
Fix MO scrape (writing empty notes into bill versions)
Fix NM scrape (doubling up on https://github.com/openstates/openstates-scrapers/pull/4414, this makes custom mappings more consistent)

### DIFF
--- a/scrapers/md/events.py
+++ b/scrapers/md/events.py
@@ -114,7 +114,8 @@ class MDEventScraper(Scraper, LXMLMixin):
             )
 
             com_name = re.sub(r"[\s\-]*Work Session", "", com_row)
-            event.add_participant(name=com_name, type="committee", note="host")
+            if com_name:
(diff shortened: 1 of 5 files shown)


scrapers/md/events.py
```diff
@@ -114,6 +114,8 @@ def scrape(self, session=None, start=None, end=None):
             )
 
             com_name = re.sub(r"[\s\-]*Work Session", "", com_row)
+            if not com_name:
+                com_name = "See Agenda"
             event.add_participant(name=com_name, type="committee", note="host")
```
showerst, line 119: We should make it so it only adds a participant if com_name is set.

PR #4420, opened 2023-01-25
re-write KS event scraper to mostly pull bills

johnseekins:

This rewrite switches the source of KS events significantly, but does actually discover bills being worked on in committees again.

### DIFF
--- a/scrapers/ks/events.py
+++ b/scrapers/ks/events.py
@@ -1,29 +1,24 @@
 import datetime
-import lxml
-import json
-import re
 import dateutil.parser
+import lxml
+import pytz
+import re
 
 from openstates.scrape import Scraper, Event
 from openstates.exceptions import EmptyScrape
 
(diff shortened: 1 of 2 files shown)


showerst:
This looks great! Have you been able to keep it consistently scraping? In the past we've favored the API with KS because their rate limiting is obscene (i've triggered IP blocks from normal browsing in chrome). Would be great if they've lightened that up a bit.

johnseekins:
Shouldn't the RPM already be set by `scrapers/ks/__init__.py`? It's set to 12 in there...

showerst (approved):

showerst:
> Shouldn't the RPM already be set by `scrapers/ks/__init__.py`? It's set to 12 in there...

Huh, you're right. I wonder if that's working -- i just re-ran it with --rpm=10 and it ran fine.

PR #4426, opened 2023-01-26
Added MA committees scraper  

EhtishamSabir:
Please review the code I had to add verify false flag & I suppressed warnings for ssl error.

### DIFF
--- a/scrapers_next/ma/committees.py
+++ b/scrapers_next/ma/committees.py
@@ -7,7 +7,9 @@ requests.packages.urllib3.disable_warnings()
 
 class SubcommitteeFound(BaseException):
     def __init__(self, com_name):
-        super().__init__(f"Scraper has no process for ingesting subcommittee classification: {com_name}")
+        super().__init__(
+            f"Scraper has no process for ingesting subcommittee classification: {com_name}"
+        )
 
 
 class CommitteeDetail(HtmlPage):
@@ -56,7 +58,7 @@ class CommitteeList(HtmlListPage):
 
     def process_page(self):
         for committees in XPath('//*[contains(@class, "committeeList")]').match(
-                self.root
+            self.root
         ):
             for each_committee in committees:
                 name = str(each_committee.text_content())
--- a/scrapers_next/ma/committees.py
+++ b/scrapers_next/ma/committees.py
@@ -1,8 +1,13 @@
-from spatula import URL, HtmlListPage, HtmlPage, CSS, XPath, selectors
+from spatula import URL, HtmlListPage, HtmlPage, CSS, XPath, SelectorError, SkipItem
 from openstates.models import ScrapeCommittee
 import requests
 
 requests.packages.urllib3.disable_warnings()
+
+
+class SubcommitteeFound(BaseException):
+    def __init__(self, com_name):
+        super().__init__(f"Scraper has no process for ingesting subcommittee classification: {com_name}")
 
 
 class CommitteeDetail(HtmlPage):
@@ -16,12 +21,12 @@ class CommitteeDetail(HtmlPage):
             member_blocks = XPath('//*[contains(@class, "committeeMemberList")]').match(
                 self.root
             )[0]
-        except selectors.SelectorError:
-            return com
+        except SelectorError:
+            raise SkipItem(f"No membership data found for: {com.name}")
         for data in member_blocks:
             # break if no members is in the list
             if "There are no members" in data.text:
-                break
+                raise SkipItem(f"No membership data found for: {com.name}")
             # continue to extract members if available
             name = None
             role = "Member"
@@ -51,39 +56,42 @@ class CommitteeList(HtmlListPage):
 
     def process_page(self):
         for committees in XPath('//*[contains(@class, "committeeList")]').match(
-            self.root
+                self.root
         ):
             for each_committee in committees:
                 name = str(each_committee.text_content())
-                if any(name.startswith(s) for s in ["House", "Senate", "Joint"]):
-                    cm_url = CSS("a").match_one(each_committee).get("href")
-                    if "house committee" in name.lower():
-                        chamber = "lower"
-                    elif "senate committee" in name.lower():
-                        chamber = "upper"
-                    else:
-                        chamber = "legislature"
-                    classification = "committee"
-                    parent = None
 
-                    com = ScrapeCommittee(
-                        name=name,
-                        chamber=chamber,
-                        parent=parent,
-                        classification=classification,
-                    )
+                if "subcommittee" in name.lower():
+                    raise SubcommitteeFound(name)
 
-                    com.add_source(
+                cm_url = CSS("a").match_one(each_committee).get("href")
+                if "house committee" in name.lower():
+                    chamber = "lower"
+                elif "senate committee" in name.lower():
+                    chamber = "upper"
+                else:
+                    chamber = "legislature"
+                classification = "committee"
+                parent = None
+
+                com = ScrapeCommittee(
+                    name=name,
+                    chamber=chamber,
+                    parent=parent,
+                    classification=classification,
+                )
+
+                com.add_source(
+                    cm_url,
+                    note="Committee List page of malegislature gov site",
+                )
+
+                yield CommitteeDetail(
+                    com,
+                    source=URL(
                         cm_url,
-                        note="Committee List page of malegislature gov site",
-                    )
-
-                    yield CommitteeDetail(
-                        com,
-                        source=URL(
-                            cm_url,
-                            timeout=30,
-                            headers=self.source.headers,
-                            verify=False,
-                        ),
-                    )
+                        timeout=30,
+                        headers=self.source.headers,
+                        verify=False,
+                    ),
+                )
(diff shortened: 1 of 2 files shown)


scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,100 @@
...
+            self.source.url,
+            note="Committee Details Page",
+        )
+        return com
+
+
+class CommitteeList(HtmlListPage):
+    source = URL(
+        "https://malegislature.gov/Committees",
+        headers={
```
johnseekins, line 43: Do we need these headers? I think at least a few of them may cause strange behavior at the edge.

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,100 @@
...
+            "Pragma": "no-cache",
+            "Cache-Control": "no-cache",
+        },
+        verify=False,
+    )
+
+    def process_page(self):
+        for comm in XPath('//*[contains(@class, "committeeList")]').match(self.root):
+            for each_committee in comm:
+                if comm is not None:
```
johnseekins, line 63: If it were me, I'd do:

```
for each_committee in comm:
    if not comm:
        continue
    name = str(each_committee.text_content())
    ...
```
One less layer of indent and a slightly less specific evaluation makes the code slightly easier to read. But both ways do work...

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,100 @@
...
+                # continue to extract members if available
+                for members_info in XPath(
+                    '//*[contains(@class, "committeeMemberList")]'
+                ).match_one(members_area):
+                    for data in members_info:
+                        name = None
+                        role = "Member"
+                        for info in data.text_content().splitlines():
+                            if info.strip():
+                                if name is None:
```
johnseekins, line 26: Should this be `if not name:`? Catches more `False`-y values that way (like empty strings).

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,100 @@
...
+                    break
+                # continue to extract members if available
+                for members_info in XPath(
+                    '//*[contains(@class, "committeeMemberList")]'
+                ).match_one(members_area):
+                    for data in members_info:
+                        name = None
+                        role = "Member"
+                        for info in data.text_content().splitlines():
+                            if info.strip():
```
johnseekins, line 25: As I mentioned elsewhere:
```
if not info:
    continue
```
Might be nicer than indenting the actual comparison another level.

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,100 @@
...
+
+class CommitteeDetail(HtmlPage):
+    def process_page(self):
+        com = self.input
+        try:
+            for members_area in XPath(
+                '//*[contains(@class, "committeeMemberList")]'
+            ).match_one(self.root)[0]:
+                if "There are no members" in members_area.text:
+                    break
```
johnseekins, line 16: `break` or `continue` here? `break` obviously means any lines _after_ the one you found will be missed.

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,100 @@
...
+from openstates.models import ScrapeCommittee
+import requests
+
+requests.packages.urllib3.disable_warnings()
+
+
+class CommitteeDetail(HtmlPage):
+    def process_page(self):
+        com = self.input
+        try:
```
johnseekins, line 11: If possible, let's keep `for` loops out of `try` blocks. It isn't critical, and if this is the right solution, that's fine. But having a huge `try` block usually makes it harder to debug errors later.

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,100 @@
...
+        },
+        verify=False,
+    )
+
+    def process_page(self):
+        for comm in XPath('//*[contains(@class, "committeeList")]').match(self.root):
+            for each_committee in comm:
+                if comm is not None:
+                    name = str(each_committee.text_content())
+                    if (
```
johnseekins, line 65: How about `if any(name.startswith(s) for s in ["House", "Senate", "Joint"]):`? Makes it easier to expand/contract later if need be.

scrapers_next/ma/committees.py
EhtishamSabir: we don't need them but I just removed few...

scrapers_next/ma/committees.py
EhtishamSabir: less layer now.

scrapers_next/ma/committees.py
EhtishamSabir: done

scrapers_next/ma/committees.py
EhtishamSabir: yes break... no need to look into other divs if there is text mentioning that there are no members

scrapers_next/ma/committees.py
EhtishamSabir: actually I'm skipping temporary committees

scrapers_next/ma/committees.py
EhtishamSabir: actually info some times has spaces only... so I'm checking if info.strip()

scrapers_next/ma/committees.py
EhtishamSabir: done :) thanks

johnseekins (approved):

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,89 @@
...
+        com = self.input
+        com.add_source(
+            self.source.url,
+            note="Committee Details Page",
+        )
+        try:
+            member_blocks = XPath('//*[contains(@class, "committeeMemberList")]').match(
+                self.root
+            )[0]
+        except selectors.SelectorError:
```
chrisyamas, line 19: Change the import from `selectors` to `SelectorError`, because you are already directly importing other selectors like `CSS` and `XPath`.

Then this line will change to simply `except SelectorError:`

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,89 @@
...
+            return com
```
chrisyamas, line 20: Rewrite this line as `raise SkipItem(f"No membership data found for: {com.name}")`

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,89 @@
...
+        for data in member_blocks:
+            # break if no members is in the list
+            if "There are no members" in data.text:
+                break
```
chrisyamas, line 24: also rewrite this line as `raise SkipItem(f"No membership data found for: {com.name}")`

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,89 @@
+from spatula import URL, HtmlListPage, HtmlPage, CSS, XPath, selectors
```
chrisyamas, line 1: In addition to changing import of `selectors` to `SelectError` to import it directly, you will also need to import `SkipItem` from spatula for the changes I recommend below.

The reason we want to raise `SkipItem` for any committees without membership is because there needs to be at least one member in each scraped committee JSON file for the merge to our [people data repository](https://github.com/openstates/people/tree/main/data). We will run the scrapers again later in the session to get committee membership that was added later.

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,89 @@
...
+        verify=False,
+    )
+
+    def process_page(self):
+        for committees in XPath('//*[contains(@class, "committeeList")]').match(
+            self.root
+        ):
+            for each_committee in committees:
+                name = str(each_committee.text_content())
+                if any(name.startswith(s) for s in ["House", "Senate", "Joint"]):
```
chrisyamas, line 58: You actually don't need such a narrow filter (or any filter for committees) here because we want the Special, Temporary Senate, and Temporary House Committees on the page also.

scrapers_next/ma/committees.py
```diff
@@ -0,0 +1,89 @@
...
+                    cm_url = CSS("a").match_one(each_committee).get("href")
+                    if "house committee" in name.lower():
```
chrisyamas, line 60: While Massachusetts does not currently list subcommittees, we'd like the scraper to be ready to alert us if a subcommittee appears in the future (ideally we would write handling for it now, but we don't know how it would appear). Above line 60 (i.e. where you start checking for "house committee" etc.) let's add a check for the word "subcommittee" in `name.lower()`, and raise a custom exception in the event that it is.

Can you create that custom exception at the top of the file (just below your `disable_warnings()` call) that reads:
```
class SubcommitteeFound(BaseException):
    def __init__(self, com_name):
        super().__init__(f"Scraper has no process for ingesting subcommittee classification: {com_name}")
```

chrisyamas (changes requested):
Data looks good now, but don't be alarmed that all the committees will be skipped once you've added the requested changes. We will be waiting for Massachusetts to post their committee assignments before we do our initial data ingestion.

However, lint check looks good at this point, so as long as it still is after you've made the requested changes, then we should be ready to merge.

scrapers_next/ma/committees.py
chrisyamas: Sorry for giving you mixed signals with my request to remove all filters here! But it will benefit us to have the data for temporary and special committees also, so I appreciate you making the [requested change](https://github.com/openstates/openstates-scrapers/pull/4426/files#r1089225916).

scrapers_next/ma/committees.py
EhtishamSabir: added SubcommitteeFound

chrisyamas (approved):
Looks like there actually was membership data for the Temporary House and Temporary Senate committees, and that data looks good! It's ready for merge now.

PR #4427, opened 2023-01-27
MA: Bills: Better handling of changing bill ids

showerst:
@NewAgeAirbender @johnseekins @jessemortenson  -- Flagging y'all while this is still a WIP to get your feedback.

MA has a system where they file a bill with one number "docket number", "HD123", then once it hits committee they list it under a normal bill number, unrelated to the docket number. "H345". 

Traditionally Openstates has just duped them, so the docket version stops updating after intro and we have a ton of dupes under the different bill numbers.

I've updated the scraper to do two things:

1. Now when we pull an intro-d bill, we add the docket version as a companion, using the "replaces" relationship. This is in the database but I don't think it's ever been used.
2. I updated the scraper to use the API for the bill list, instead of a big, slow paginated search. I think someone on your team evaluated the API last year and determined that it wasn't reliable for prime time, but in my tests this one endpoint seems ok, and saves us a TON of requests while also providing the docket/into mapping we need.

I have no clue if openstates-core does any replacement or matching if you import a bill with a replaces relationship, so that's an exercise left to the reader.

If you wanna test this without it taking forever:

```python
poetry run os-update ma bills session=193rd bill_no=H49 --scrape
```

Will give you one that's been introduced.

Curious on your thoughts on this approach, it would be nice to get this settled before they get too many bill numbers swapped out.

### DIFF
--- a/scrapers/ma/bills.py
+++ b/scrapers/ma/bills.py
@@ -71,11 +71,9 @@ class MABillScraper(Scraper):
                 single_bill_chamber = "upper"
 
             for bill_meta in self.bill_list[single_bill_chamber]:
-                if (
-                    bill_no == bill_meta["DocketNumber"]
-                    or bill_no == bill_meta["BillNumber"]
-                ):
+                if bill_no in [bill_meta["DocketNumber"], bill_meta["BillNumber"]]:
                     yield from self.scrape_bill(session, bill_meta, single_bill_chamber)
+                    self.info("Finished bill scrape, exiting.")
                     return
 
         if not chamber:
--- a/scrapers/ma/bills.py
+++ b/scrapers/ma/bills.py
@@ -18,7 +18,7 @@ class MABillScraper(Scraper):
     chamber_filters = {}
     house_pdf_cache = {}
 
-    bill_list = {"lower": [], "upper": []}
+    bill_list = []
 
     chamber_map = {"lower": "House", "upper": "Senate"}
     chamber_map_reverse = {
@@ -53,27 +53,18 @@ class MABillScraper(Scraper):
             refiner_list[label] = refiner
         return refiner_list
 
-    # sort can be set to "latest" to sort by date modified descending
-    # os-update ma bills --scrape sort=latest
     # bill_no can be set to a specific bill (no spaces) to scrape only one
     # os-update ma bills --scrape bill_no=H2
-    # page_limit can be set to stop scraping after a certain number of pages (for each chamber)
-    def scrape(
-        self, chamber=None, session=None, bill_no=None, sort=None, page_limit=None
-    ):
+    def scrape(self, chamber=None, session=None, bill_no=None):
         self.scrape_bill_list(session)
 
+        # optionally scrape a single bill then exit
         if bill_no:
-            single_bill_chamber = False
-            if "H" in bill_no:
-                single_bill_chamber = "lower"
-            else:
-                single_bill_chamber = "upper"
-
-            for bill_meta in self.bill_list[single_bill_chamber]:
+            single_bill_chamber = "lower" if "H" in bill_no else "upper"
+            for bill_meta in self.bill_list:
                 if bill_no in [bill_meta["DocketNumber"], bill_meta["BillNumber"]]:
                     yield from self.scrape_bill(session, bill_meta, single_bill_chamber)
-                    self.info("Finished bill scrape, exiting.")
+                    self.info("Finished individual bill scrape, exiting.")
                     return
 
         if not chamber:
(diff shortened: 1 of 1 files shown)


scrapers/ma/bills.py
```diff
@@ -68,91 +70,55 @@ def scrape(
             else:
                 single_bill_chamber = "upper"
 
-            yield from self.scrape_bill(session, bill_no, single_bill_chamber)
-            return
+            for bill_meta in self.bill_list[single_bill_chamber]:
+                if (
```
johnseekins, line 74: `if bill_no in [bill_meta["DocketNumber"], bill_meta["BillNumber"]:`  instead?

scrapers/ma/bills.py
```diff
@@ -68,91 +70,55 @@ def scrape(
...
+                    bill_no == bill_meta["DocketNumber"]
+                    or bill_no == bill_meta["BillNumber"]
+                ):
+                    yield from self.scrape_bill(session, bill_meta, single_bill_chamber)
+                    return
```
johnseekins, line 79: I know the pattern is valid, but this feels weird. Maybe `break` with a comment instead? Or at least explain the `return` here.

scrapers/ma/bills.py
```diff
@@ -68,91 +70,55 @@ def scrape(
...
+            if "H" in str(row["BillNumber"]) or "H" in str(row["DocketNumber"]):
+                chamber = "lower"
+            elif "S" in str(row["BillNumber"]) or "S" in str(row["DocketNumber"]):
+                chamber = "upper"
+            else:
+                self.error(
+                    f"Unknown bill type - bill {row['BillNumber']} - docket {row['DocketNumber']}"
                 )
+
+            self.bill_list[chamber].append(
```
johnseekins, line 105: This object feels a little weird. Should we just have a dict instead of a list of dicts?

scrapers/ma/bills.py
```diff
@@ -68,91 +70,55 @@ def scrape(
...
-                '//ul[contains(@class,"pagination-sm")]/li[last()]/a/@onclick'
-            )[0]
-            maxPage = re.sub(r"[^\d]", "", maxPage).strip()
-        else:
-            maxPage = 1
+    def scrape_bill(self, session, bill_meta, chamber):
+        # https://malegislature.gov/Bills/189/SD2739
 
-        return int(maxPage)
+        if bill_meta["BillNumber"]:
```
johnseekins, line 117: `bill_id = bill_meta.get("BillNumber", bill_meta["DocketNumber"])`?

scrapers/ma/bills.py
```diff
@@ -216,6 +182,13 @@ def scrape_bill(self, session, bill_id, chamber):
         if bill_summary:
             bill.add_abstract(bill_summary, "summary")
 
+        if bill_meta["BillNumber"] and bill_meta["DocketNumber"]:
```
johnseekins, line 185: In other places in the code, we rely on `bill_meta["DocketNumber"]` having a value. Why do we check for it here?

johnseekins:
I think it's a good idea.

scrapers/ma/bills.py
showerst: I think since I was adapting the old code (which is chamber based) I had it in my head that there was some reason to split it, but there's really not; good idea i'll refactor that out.

scrapers/ma/bills.py
showerst: This is in a nested block, so break won't work -- i don't see a great way to refactor out the nest. I'll drop in a self.info comment through so it's clear what's happening.

NewAgeAirbender:
I think this is a great idea! Definitely helps keep track of the full life of a bill. I was also thinking of reevaluating the API approach while the new session is still smaller, so appreciate incorporating that change as well.

scrapers/ma/bills.py
showerst: Mirroring the states's json, these keys are always defined, and are either `String` or `None`. 

If they're both not `None`, then we want to add the relation because the bill has both numbers.

I like the way this simplifies the logic of "Bill Number if its not None, otherwise fall back docket number" into one `or` statement, but if you feel like that's leaning a bit too much on type trickery i'm happy to restructure. The alternative is probably to only define the key when it's set, and then reference it by `.get("BillNumber", ` with a default of the docket number like your previous suggestion.

johnseekins: Really I'm just calling out differing behavior within the scraper. If this is the right solution, we should follow it throughout the scraper. If falling back to `DocketNumber` is the correct solution, we should do that.

PR #4428, opened 2023-01-27
MI people: senate fix, senate both and house gop images add

chrisyamas:
This PR was undertaken to address URL and page modifications that caused Senate scraper to break for MI people.

It addresses those issues. It also:
- adds ingestion of member image paths for all Senate members, and for GOP members of House
   - creates one new object class, one new helper function, and two custom exceptions to this end 
- adds documentation for prior-created helper function
- includes documentation of work done to eventually also get Dem members of House
- various other small modifications

### DIFF
--- a/jurisdiction_configs.json
+++ b/jurisdiction_configs.json
@@ -149,8 +149,8 @@
   {
     "name": "mi",
     "people": [
-      "scrapers_next.mi.people.Senators",
-      "scrapers_next.mi.people.Representatives"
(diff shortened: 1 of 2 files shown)


bfossen-ce (commented):
Since you made changes to the run class names could you reflect that here in the https://github.com/openstates/openstates-scrapers/blob/main/jurisdiction_configs.json We use that to do our automated runs. Besides that it LGTM

PR #4430, opened 2023-01-27
VT Committee scraper

EhtishamSabir:
Scraper for VT using JSON page.

### DIFF
--- a/scrapers_next/vt/committees.py
+++ b/scrapers_next/vt/committees.py
@@ -75,12 +75,17 @@ class CommitteeList(JsonListPage):
             if not com.members:
                 logging.warning(f"No membership data found for: {name}")
                 continue
-            if each_committee.get("WebSiteURL"):
-                com.add_link(each_committee["WebSiteURL"], note="Committee web page")
-            if each_committee.get("StreamingURL"):
-                com.add_link(
-                    each_committee["WebSiteURL"], note="Committee streaming page"
-                )
+            url_dict = {
+                "WebSiteURL": "web",
+                "StreamingURL": "streaming",
+                "ExpensesURL": "expenses",
+            }
+            for url_type, note_term in url_dict.items():
+                if each_committee.get(url_type):
+                    if len(each_committee[url_type]):
+                        com.add_link(
+                            each_committee[url_type], note=f"Committee {note_term} page"
+                        )
 
             com.add_source(
                 self.source.url,
--- a/scrapers_next/vt/committees.py
+++ b/scrapers_next/vt/committees.py
@@ -43,7 +43,9 @@ class CommitteeList(JsonListPage):
             cleaner = Cleaner(allow_tags=[""], remove_unknown_tags=False)
             members_list = cleaner.clean_html(doc).text_content()
 
-            name = name.lower().replace("house", "").replace("senate", "").strip().title()
+            name = (
+                name.lower().replace("house", "").replace("senate", "").strip().title()
+            )
             com = ScrapeCommittee(
                 name=name,
                 chamber=chamber,
@@ -59,7 +61,7 @@ class CommitteeList(JsonListPage):
                     member = member.lower().replace("rep.", "").replace("sen.", "")
                     splits = member.split(",")
                     member_name = splits[0]
-                    role = ' '.join(splits[1:len(splits)]).strip()
+                    role = " ".join(splits[1 : len(splits)]).strip()
                 except IndexError:
                     member_name = member
                     role = "Member"
@@ -76,7 +78,9 @@ class CommitteeList(JsonListPage):
             if each_committee.get("WebSiteURL"):
                 com.add_link(each_committee["WebSiteURL"], note="Committee web page")
             if each_committee.get("StreamingURL"):
-                com.add_link(each_committee["WebSiteURL"], note="Committee streaming page")
+                com.add_link(
+                    each_committee["WebSiteURL"], note="Committee streaming page"
+                )
 
             com.add_source(
                 self.source.url,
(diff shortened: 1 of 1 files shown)


johnseekins (approved):
Looks great.

scrapers_next/vt/committees.py
```diff
@@ -0,0 +1,76 @@
...
+from lxml.html.clean import Cleaner
+from lxml.etree import ParserError
+from spatula import JsonListPage, URL, SkipItem
+from openstates.models import ScrapeCommittee
+
+
+class CommitteeList(JsonListPage):
+    source = URL(
+        "https://legislature.vermont.gov/committee/loadList/2024/",
+        headers={
```
johnseekins, line 11: Do we need this header? If we don't have to lie about our user agent, I'd rather we didn't.

EhtishamSabir: okk

scrapers_next/vt/committees.py
```diff
@@ -0,0 +1,73 @@
...
+                doc = fromstring(members_html)
+            except ParserError:
+                SkipItem(f"No membership data found for: {name}")
+                continue
+
+            cleaner = Cleaner(allow_tags=[""], remove_unknown_tags=False)
+            members_list = cleaner.clean_html(doc).text_content()
+
+            com = ScrapeCommittee(
+                name=name,
```
chrisyamas, line 47: Before storing name, we also want to remove "House" and "Senate".

Examples:
- "House Ethics Panel" becomes "Ethics Panel"
- "Senate Rules Committee" becomes "Rules Committee"

scrapers_next/vt/committees.py
```diff
@@ -0,0 +1,73 @@
...
+                classification = "subcommittee"
+                parent = name.lower().split()[-1].replace(")", "")
+
+            # extracting members
+            members_html = each_committee["Members"].strip()
+            members_html = members_html.replace("</br>", "\n")
+            try:
+                doc = fromstring(members_html)
+            except ParserError:
+                SkipItem(f"No membership data found for: {name}")
```
chrisyamas, line 40: Given that you're running this scraper in a `process_page()` method rather than `process_item()`, the `SkipItem()` class isn't appropriate here.
- Change it to `logging.warning(f"No membership data found for: {name}")` instead (you'll need to `import logging` and the top and also remove `SkipItem` from your imports)

scrapers_next/vt/committees.py
```diff
@@ -0,0 +1,73 @@
...
+            com = ScrapeCommittee(
+                name=name,
+                chamber=chamber,
+                parent=parent,
+                classification=classification,
+            )
+
+            for member in members_list.splitlines():
+                if not member.strip():
+                    SkipItem(f"No membership data found for: {name}")
```
chrisyamas, line 55: As in above comment, change this one also to `logging.warning(f"No membership data found for: {name}")`

scrapers_next/vt/committees.py
```diff
@@ -0,0 +1,73 @@
...
+                except ValueError:
+                    name = member
+                    role = "Member"
+                com.add_member(name.strip().title(), role=role.strip().title())
+
+            # additional check if no member found
+            if not com.members:
+                SkipItem(f"No membership data found for: {name}")
+                continue
+
```
chrisyamas, line 68: In looking at the source JSON just now, I noticed that there are two potentially useful fields involving URLs for each committee: `"WebSiteURL"` and `"StreamingURL"`.

Could you add a `com.add_link()` method call for each case to store the respective values if they do exist?
- Example for the website url:
```
if each_committee.get("WebSiteURL"):
    com.add_link(each_committee["WebSiteURL"], note="Committee web page")
```

And same for the `"StreamingURL"` key, with the note reading "Committee streaming page".

chrisyamas (changes requested):
Another great scraper, and just a few changes before we're all the way there.

Thanks again for working with us to make this scraper as effective and useful as reasonably possible before merging!

scrapers_next/vt/committees.py
```diff
@@ -0,0 +1,89 @@
...
+            # additional check if no member found
+            if not com.members:
+                logging.warning(f"No membership data found for: {name}")
+                continue
+            if each_committee.get("WebSiteURL"):
+                com.add_link(each_committee["WebSiteURL"], note="Committee web page")
+            if each_committee.get("StreamingURL"):
+                com.add_link(
+                    each_committee["WebSiteURL"], note="Committee streaming page"
+                )
```
chrisyamas, line 83: This is my fault, but in looking at the data now I see that this actually stores any non-null values for these urls (including empty strings). Could you push up a commit with this solution:
```
url_dict = {
                "WebSiteURL": "web",
                "StreamingURL": "streaming",
                "ExpensesURL": "expenses"
            }
            for url_type, note_term in url_dict.items():
                if each_committee.get(url_type):
                    if len(each_committee[url_type]):
                        com.add_link(
                            each_committee[url_type],
                            note=f"Committee {note_term} page"
                        )
```
Once you've done that, it'll be ready for merge!

chrisyamas: (I also decided to just consolidate it into a dictionary, and then include the expense url key in case there are ever values for that added in the future)

chrisyamas (changes requested):
If you can just push up that one change I wrote there, then I will review, approve and merge!

scrapers_next/vt/committees.py
EhtishamSabir: yes sure

chrisyamas (approved):
Data looks good, we're there. Merging now.

PR #4431, opened 2023-01-28
OH committee scraper

a-cat-named-snowball:


scrapers_next/oh/committees.py
```diff
@@ -0,0 +1,385 @@
...
+    def process_page(self):
+        com = ScrapeCommittee(
+            name=self.input.get("name"),
+            chamber="legislature",
+            classification="committee",
+        )
+
+        com.add_source(self.source.url, note="Committee member listing page")
+
+        for (name, role) in self.get_members():
```
johnseekins, line 30: Given the signature of `get_members` below...won't this _always_ be an empty list?

a-cat-named-snowball: No, JointCommittee is a base class and the get_members method is overridden by the classes that extend it. I've added a comment and changed the default implementation to raise an error if it's ever called.

chrisyamas (approved):
An incredibly thorough job here, and great documentation within the code. Thank you for your attention to detail, once again! Data looks good, and this is ready to merge.

johnseekins (approved):

PR #4443, opened 2023-02-01
UT comms: helper function moved into committees file, utils.py deleted

chrisyamas:
The `utils.py` file was deleted because the helper function `get_membership_dict()` contained within it was not being used. For simplicity, the helper function was moved into the `committees.py` file, and the `CommitteeList()` class therein was reformatted to utilize this function.


NewAgeAirbender (approved):

NewAgeAirbender:
Actually, could you update the `add_link` to have the note="homepage" since you're updating the file anyway?

PR #4446, opened 2023-02-02
HI: Bills: Detect correct session for companion bills

showerst:
We were always assuming companions were from the prior session, which was wrong in some cases. For example, 2023 SB 181 has a 2023 companion, which was getting marked as 2022.

### DIFF
--- a/scrapers/hi/bills.py
+++ b/scrapers/hi/bills.py
@@ -301,12 +301,16 @@ class HIBillScraper(Scraper):
         if url:
             b.add_source(url)
 
-        prior_session = "{} Regular Session".format(str(int(session[:4]) - 1))
         companion = meta["Companion"].strip()
         if companion:
+            companion_url = bill_page.xpath(
+                "//span[@id='ctl00_MainContent_ListView1_ctrl0_companionLabel']/a/@href"
+            )[0]
+            companion_year = companion_url[-4:]
+            companion_session = self.session_from_scraped_name(companion_year)
             b.add_related_bill(
                 identifier=companion.replace("\xa0", " "),
-                legislative_session=prior_session,
+                legislative_session=companion_session,
                 relation_type="companion",
             )
         if bill_page.xpath(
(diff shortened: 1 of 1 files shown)


scrapers/hi/bills.py
```diff
@@ -315,6 +319,7 @@ def scrape_bill(self, session, chamber, bill_type, url):
             prior = bill_page.xpath(
                 "//table[@id='ContentPlaceHolderCol1_GridViewStatus']/tr/td/font/text()"
             )[-1]
+            prior_session = "{} Regular Session".format(str(int(session[:4]) - 1))
```
johnseekins, line 322: Just a nit, but we have f-strings and format-strings peppered through this scraper. Should we settle on one?

scrapers/hi/bills.py
```diff
@@ -301,12 +301,16 @@ def scrape_bill(self, session, chamber, bill_type, url):
         if url:
             b.add_source(url)
 
-        prior_session = "{} Regular Session".format(str(int(session[:4]) - 1))
         companion = meta["Companion"].strip()
         if companion:
+            companion_url = bill_page.xpath(
+                "//span[@id='ctl00_MainContent_ListView1_ctrl0_companionLabel']/a/@href"
+            )[0]
+            companion_year = companion_url[-4:]
```
johnseekins, line 309: Magic number?

scrapers/hi/bills.py
```diff
@@ -413,3 +418,12 @@ def scrape(self, chamber=None, session=None):
                 bill_types.append("gm")
             for typ in bill_types:
                 yield from self.scrape_type(chamber, session, typ)
+
+    # backtrack from the URL session for companion bills
+    def session_from_scraped_name(self, scraped_name):
+        details = next(
```
johnseekins, line 424: I (mostly) understand what you're doing here, but is it worth adding a little clarification here about how we're getting a single object from a generator?

scrapers/hi/bills.py
showerst: oops, that's a cut & paste from the previous code. -- I always do f-strings now, will try to watch for them in cut and pastes as well

scrapers/hi/bills.py
showerst: Sure, this is also a cut & paste from older code hence the weird syntax. TBH I think some rainy we should overhaul how we access other sessions and look stuff up, we have this goofy pattern littered everywhere.

Maybe in the future a utility class with some kind of session_by_scraped_name and session_by_id helpers would be nice.

I'll drop in a better comment here anyhow.

PR #4451, opened 2023-02-03
MT committee scraper

a-cat-named-snowball:
Created a scraper for MT committees.


johnseekins (approved):

scrapers_next/mt/committees.py
```diff
@@ -0,0 +1,185 @@
...
+        if len(com.members) == 0:
+            raise SkipItem(f"No membership data found for: {com.name}")
+
+        return com
+
+
+# This page contains information on the Administrative committees only
+# This site is a little buggy or perhaps there's a rate limiter involved
+# Sometimes requests to this site do not return the expected html
+class AdministrativeCommitteeList(HtmlListPage):
```
chrisyamas, line 124: We want to get Interim Committees too (let's reformat the code to be able to do that even though the actual committees haven't been fully updated with new session membership yet). Good news is, I tested it out, and your existing code for handing Administrative Committees will also work for Interim.

You'll just need to:
- Rename this class `OtherCommList` and remove the `source` attribute defined within it.
- Define the class instances of `Administrative(OtherCommList)` and `Interim(OtherCommList)` as child classes with their respective source attributes defined within each child class (for interim it'll be https://leg.mt.gov/committees/interim/)
- Then, in the details class, add an additional selector that grabs the alternative name of the committee using xpath `".//div[@class='container white']/h2"`, which, in cases like the [2021-2022 Children, Families, Health, and Human Services Interim Committee](https://leg.mt.gov/committees/interim/cfhhs/) gets the more complete name than the string `"2021-2022 CFHHS"` which is what the current selector grabs
   - Then, compare the length of the two strings, and if the one you're currently grabbing from the breadcrumb items is shorter, concatenate it to the long one (after any date range is removed, perhaps using regex).
   - So, for this example, the final saved committee name should be `"Children, Families, Health, and Human Services Interim (CFHHS)"
- **Note: just as you should allow the word "Legislative" to stay in the name, also keep "Interim" to help distinguish from any standing committees that may have the same name as interim committees (important for us internally when we're matching / referencing committee names in bills).**

- *Also, add a conditional check in the list page class, that checks if "interim" is in `self.source.url`. If it is, use `logging.warning` to log a warning to the console something like: "Interim Committee data may not be updated yet for this current session: a manual check should be conducted". (of course, you'll need to import `logging`)*

Let me know if you have any questions while executing on this change

scrapers_next/mt/committees.py
```diff
@@ -0,0 +1,185 @@
...
+    source = "https://leg.mt.gov/committees/admincom/"
+    selector = XPath("//*[@id='cont']/section/div/div[1]/div/div/div[1]/ul[1]/li/a[1]")
+
+    def process_item(self, link):
+        return CommitteeDetailsPage(source=URL(link.get("href"), timeout=30))
+
+
+class CommitteeDetailsPage(HtmlPage):
```
chrisyamas, line 132: Rename to `OtherCommDetail`

scrapers_next/mt/committees.py
```diff
@@ -0,0 +1,185 @@
...
+            com.add_member(name=member, role=role)
+
+        # Skip if no members found
+        if len(com.members) == 0:
+            raise SkipItem(f"No membership data found for: {com.name}")
+
+        return com
+
+
+# This page contains information on the Administrative committees only
```
chrisyamas, line 121: Adjust comment accordingly for the addition of Interim committees

chrisyamas (changes requested):
Overall, another well-performing scraper with thorough attention to detail. I left comments requesting some minor modifications, and one more substantial change to include ingestion and proper processing of Interim Committees.

Thanks for another great contribution! Once these changes have been made successfully, we should be ready to merge.

a-cat-named-snowball:
Those changes have been made. Thank you for the detailed notes for the interim committees.

scrapers_next/mt/committees.py
```diff
@@ -0,0 +1,228 @@
...
+            committee_title = committee_title[len(house_prefix) :]
+        if committee_title.startswith(senate_prefix):
+            chamber = "upper"
+            committee_title = committee_title[len(senate_prefix) :]
+
+        # 2) A joint prefix may exist
+        joint_prefix = "Joint "
+        if committee_title.startswith(joint_prefix):
+            chamber = "legislature"
+            committee_title = committee_title[len(joint_prefix) :]
```
chrisyamas, line 62: Remove this (and adjust the code and comment numbering accordingly). Apologies, but it's recently come to my attention that we want to keep any occurrences of either "Joint" or "Legislative" substring at the beginning of a committee name.

chrisyamas (changes requested):
Last bit of minor changes, but data looks good and I'll merge it as soon as those are pushed up!

scrapers_next/mt/committees.py
```diff
@@ -0,0 +1,228 @@
...
+        # Build the committee object with calculated metadata
+        com = ScrapeCommittee(
+            name=committee_title,
+            chamber=chamber,
+            classification=classification,
+            parent=parent,
+        )
+        com.add_source(member_info_href, note="Committee membership page")
+        com.add_link(member_info_href, note="homepage")
+
```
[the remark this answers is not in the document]
a-cat-named-snowball: Fixed with https://github.com/openstates/openstates-scrapers/pull/4451/commits/26a72f5fb6b3679c018bce5e78ce8100b01e911c

scrapers_next/mt/committees.py
a-cat-named-snowball: Fixed with https://github.com/openstates/openstates-scrapers/pull/4451/commits/9c05a8d545d8990e651dc3449aed11c3085e17be and https://github.com/openstates/openstates-scrapers/pull/4451/commits/5864afd567afd470841ee195a6f3f5d6d21743bb

chrisyamas (approved):
Data looks good, and I appreciate the quick turnaround on those changes! Approving and merging now.

PR #4454, opened 2023-02-03
MN bills: fix gets bill versions for resolutions

chrisyamas:
This PR fixes a gap in data ingestion, wherein versions were only being retrieved for bills and not for resolutions (normal or concurrent).

Cause
- The formatting on resolution detail pages do not have the "Version List" table we were targeting for this data.

Solution
- This PR added a backup target in the case when the versions table isn't found.
- A request is made for the html page containing the "Current Bill Text", on which is found:
   - the html text of the that version
   - a link to the pdf page for that version
   - a list of links to the html pages of any other versions that may exist
      - in the case that those version links do exist, additional requests are made of each link's href in order to retrieve the pdf page urls for each version

MN resolutions and concurrent resolutions are now getting all versions.

### DIFF
--- a/scrapers/mn/bills.py
+++ b/scrapers/mn/bills.py
@@ -61,6 +61,14 @@ SITE_IDS = {
     "2023-2024": "0932023",
 }
 
+version_re = re.compile(r".+,\s+(.+):.+Session.+\)\s+Posted on\s+(.+)")
+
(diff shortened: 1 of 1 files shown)


bfossen-ce (approved):
LGTM

PR #4457, opened 2023-02-04
ID committees scraper

EhtishamSabir:
For some reason even XPath is present it's still not found at these 2 links Spent almost a day but maybe there is something else if you can help with these that would appreciated.

https://legislature.idaho.gov/sessioninfo/2023/joint/cec/#hcode-tab-style2members
https://legislature.idaho.gov/sessioninfo/2023/joint/eora/#hcode-tab-style2members


chrisyamas (changes requested):
@EhtishamSabir it took some time, but I figured out what the issue is here.

The issue is happening with three of the joint committee pages, including the two you identified in your comment:
- https://legislature.idaho.gov/sessioninfo/2023/joint/cec/#hcode-tab-style2members
- https://legislature.idaho.gov/sessioninfo/2023/joint/eora/#hcode-tab-style2members
and also this one:
- https://legislature.idaho.gov/sessioninfo/2023/joint/mill/

It's an issue with the html on those pages, but it can be fixed by converting html root content to text and then to parseable xml again. I tried that with the below code implementation to get the list of members (and their roles, if listed) at the beginning of `process_page()` in `CommitteeDetail()`, and it seems to cover all of the different committee types.
```
        try:
            members = XPath(
                ".//div[@class='vc-column-innner-wrapper']//strong[a[contains(@href, '/legislators/membership/')]]"
            ).match(self.root)
        except SelectorError:
            self.root = lxml.html.fromstring(self.root.text_content())
            members = XPath(
                ".//div[@class='vc-column-innner-wrapper']//strong[a[contains(@href, '/legislators/membership/')]]"
            ).match(self.root)
```

Let me know if it works out for you and if you have any follow-up questions!

EhtishamSabir:
sure I'll try this I'll update soon

chrisyamas (approved):
Nice job! Data looks good, and the only issue was one committee where the selector still wasn't getting the member names and roles, but I just went ahead and added the solution in for the interest of time.

I'm approving this, and once I've resolved the conflicts with main and lint checks are confirmed good, then I will merge.

PR #4458, opened 2023-02-05
WY committees Scraper

EhtishamSabir:
I don't know to identify committee chamber as name don't include anything & it appears all are legislature.


scrapers_next/wy/committees.py
```diff
@@ -0,0 +1,74 @@
...
+                    if each_member["chairman"].strip() == "":
+                        role = "Member"
+                    else:
+                        role = each_member["chairman"]
+                    com.add_member(name, role=role)
+                if not com.members:
+                    logging.warning(f"No membership data found for: {name}")
+                    continue
+
+                com.add_link(committee_url, note="homepage api")
```
chrisyamas, line 69: We want the HTML for the link always, because it is the link we will provide to users to go look at the committee homepage. Similar to how you created the `committee_url` above:

`comm_html = f"https://www.wyoleg.gov/Committees/{self.year}/{{each_comm['ownerID']}"`

Change the `add_link()` to:
- use `comm_html` as defined above
- have `note="homepage"`

chrisyamas (changes requested):
Overall, another great scraper! However, some changes are needed. All of these changes are important, but the biggest issue is the current duplicate scraping of all committees that the `for` loop is going.

scrapers_next/wy/committees.py
EhtishamSabir: added

chrisyamas (approved):
Thanks for making those changes! I just added one little fix to make sure we were still removing the substring "Subcommittee" for the one subcommittee that is present currently. Approving and merging this now.

PR #4461, opened 2023-02-05
NM committee Scraper

EhtishamSabir:
New spatula-based scraper written for New Mexico.

### DIFF
--- a/scrapers_next/nm/committees.py
+++ b/scrapers_next/nm/committees.py
@@ -4,6 +4,11 @@ from lxml import html
 from openstates.models import ScrapeCommittee
 
 
+class UnknownSubCommFound(BaseException):
+    def __init__(self, com_name):
+        super().__init__(f"unknown for subcommittee found: {com_name}")
+
+
 class CommitteeDetail(HtmlPage):
     def process_page(self):
         com = self.input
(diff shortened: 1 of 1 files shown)


johnseekins (approved):

scrapers_next/nm/committees.py
```diff
@@ -0,0 +1,109 @@
...
+
+        for chamber, elems in all_committees.items():
+            for item in elems:
+                name, com_url = item.text, item.get("href")
+                if not com_url:
+                    continue
+
+                if "subcommittee" in name.lower():
+                    classification = "subcommittee"
+                    parent = name.lower().replace("subcommittee", "").title()
```
chrisyamas, line 89: Make the same change here that you're making on the [Wyoming scraper](https://github.com/openstates/openstates-scrapers/pull/4458#discussion_r1106015163). There are only two committees so far with subcommittee in their name, and neither of them have clear parent, so just use that same kind of code block as what I suggested there to make them both `classification="committee"` and then raise a custom exception if any others are encountered in the future.

chrisyamas (changes requested):
One minor change, data looks good other than that!

scrapers_next/nm/committees.py
EhtishamSabir: done...

chrisyamas (approved):
I added some final reformatting and fixes to ensure the name cleaning function was working properly, but it's good to go now. Approving and merging!

PR #4465, opened 2023-02-06
FL: minor fix to House scraping

sroomf:
Since (non-empty) strings are always resolved as "true" in Python, there was an `if` statement that was always being entered, causing a bug. Made a minor change to the `if` statement so that this does not happen. Pertains to [OS-1009](https://civic-eagle.atlassian.net/browse/OS-1009).

### DIFF
--- a/scrapers_next/fl/people.py
+++ b/scrapers_next/fl/people.py
@@ -168,7 +168,8 @@ class Representatives(HtmlListPage):
         name = item.xpath("./a/div[@class='team-txt']/h5/text()")[0].strip()
 
         # skips empty chairs due to pending regular/special election
-        if "pending, special" in name.lower() or "pending, election" in name.lower():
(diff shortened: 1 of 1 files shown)


scrapers_next/fl/people.py
```diff
@@ -168,7 +168,7 @@ def process_item(self, item):
         name = item.xpath("./a/div[@class='team-txt']/h5/text()")[0].strip()
 
         # skips empty chairs due to pending regular/special election
-        if "pending, special" or "pending, election" in name.lower():
+        if "pending, special" in name.lower() or "pending, election" in name.lower():
```
johnseekins, line 171: Better formatting would be `if any(s in name.lower() for s in ["pending, special", "pending, election"])`.

NewAgeAirbender (approved):

PR #4471, opened 2023-02-07
PA committee scraper

a-cat-named-snowball:
Created scraper for PA committees.


scrapers_next/pa/committees.py
```diff
@@ -1,105 +1,603 @@
...
+                "Representative": "Member",
+                "Senator": "Member",
+                "Emeritus Member": "Emeritus Member",
+                "Governor's Appointee": "Governor's Appointee",
+                "Supreme Court Appointee": "Supreme Court Appointee",
+                "Executive Director": "Executive Director",
+            }
+            for k, v in roles.items():
+                if k in title:
+                    role = v
```
chrisyamas, line 212: This was very thorough, and thank you for that, but we want a bit more role information than what it is currently getting. I have a proposed solution.

**Issue:**
- Currently, [from the page being scraped](http://cpc.state.pa.us/about/capitol-preservation-committee-members.cfm), Andrea Bakewell is only being stored with role as "Executive Director", but we don't know of what. 
- Similar for Frank E. Dittenhafer, II: we are getting "Governor's Appointee" but not "FAIA, LEED AP".

**Solution for handling members:**
- if there is the placeholder image (indicating vacancy), then skip adding that member
- if the `p` for that members includes a `strong`, then get the text in the `strong` as the role, the text after the `strong` but before the comma or next `br` as the name, and then any any following text gets concatenated to the role with a comma and space (so John R. Bowie would have role as "Vice Chairman, Governor's Appointee") unless the substring that would be concatenated would be either "Senator" or "Representative" (as in the case of Treasurer Patty Kim)
- for the remainder of the members, all text after the comma / `br` should be added as role (to fix issues like in the cases of Blackwell and Dittenhafer) unless it is "Senator" or "Representative" (you already did a great job of handling Dittenhafer's case when it came to the comma in his name, so just make sure this still handles the ", II" as the end of his name and not part of the role

scrapers_next/pa/committees.py
```diff
@@ -1,105 +1,603 @@
...
+
+    return name, role
+
+
+class StateGovernmentComm(HtmlPage):
+    source = "http://jsg.legis.state.pa.us/about-us.cfm"
+
+    def process_page(self):
+        com = ScrapeCommittee(
+            name="State Government",
```
chrisyamas, line 379: Let's keep the full name of "State Government Commission" since it uses the word Commission rather than Committee.

scrapers_next/pa/committees.py
```diff
@@ -1,105 +1,603 @@
-from spatula import CSS, HtmlPage, HtmlListPage, SelectorError
+from spatula import HtmlPage, SelectorError, XPath, URL, SkipItem
 from openstates.models import ScrapeCommittee
```
chrisyamas, line 4: Let's add this docstring here to serve as documentation for those running the scrapers of which classes should be run to get all committees:
```
"""
The classes that should be run to ensure every committee is scraped include:
- House
- Senate
- JointCommitteeList
- LegReapportionmentComm
- LocalGovernmentComm
- LegBudgetFinanceComm
- LegAuditAdvisoryComm
- StateGovernmentComm
- CommissionOnSentencing
- CapitolPreservationComm
"""
```

chrisyamas (changes requested):
This was incredible work, going above expectations to ensure we were getting every committee despite the nuances of the Pennsylvania legislative site(s).

There are some minor changes requested, and one medium change specifically regarding the `CapitolPreservationComm`. However, the data looks good overall, so once those changes have been committed and the data looks good upon re-review, then this will be ready for merge! Thank you for all your hard work!

chrisyamas (approved):
Thanks for making those changes! Approving and merging now.

PR #4472, opened 2023-02-07
Created DE committee scraper

a-cat-named-snowball:
Created DE committee scraper.


chrisyamas (approved):
Flawless execution. Data looks great. I just made an edit that changed the note string on `add_link()` to be lowercase ("homepage"), but that was all that needed to be changed.

I am now approving and merging!

PR #4479, opened 2023-02-14
WY: action classification update

sroomf:
Action classification updates for WY. Relates to [OS-1076](https://civic-eagle.atlassian.net/browse/OS-1076)


NewAgeAirbender (changes requested):
Looks good, just make sure to remove the `def categorize_action(action)` stuff in `bills.py` since we don't need it anymore!

PR #4487, opened 2023-02-18
SC committee scraper

a-cat-named-snowball:
Created SC committee scraper. Subcommittee data is only available as a docx download, I wasn't able to scrape it with Spatula.


chrisyamas (approved):
Another phenomenal scraper. I especially appreciated the care you took in getting the roles from their abbreviated forms in the joint committees PDF.

Approving and merging!

PR #4498, opened 2023-02-22
AL people scraper

a-cat-named-snowball:
Created people scraper for AL.

Should I combine SenList and RepList into one class or should I keep them separate?

### DIFF
--- a/scrapers_next/al/people.py
+++ b/scrapers_next/al/people.py
@@ -27,12 +27,16 @@ class MemberList(JsonPage):
                 self.logger.warning(f"{name} has no listed district, skipping.")
                 continue
 
+            # Remove extra text from district
+            district = member.get("District").split(" ")[-1]
+            print(district)
+
             p = ScrapePerson(
                 name=member.get("FullName"),
                 state="al",
                 chamber=self.chamber,
                 party=member.get("Affiliation"),
-                district=member.get("District"),
+                district=district,
                 email=member.get("Email"),
                 image=member.get("NewImgUrl"),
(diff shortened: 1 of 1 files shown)


chrisyamas:
> Created people scraper for AL.
> 
> Should I combine SenList and RepList into one class or should I keep them separate?

@a-cat-named-snowball you're good to keep them separate.

scrapers_next/al/people.py
```diff
@@ -1,151 +1,113 @@
...
+                email=member.get("Email"),
+                image=member.get("NewImgUrl"),
+                given_name=member.get("FirstName"),
+                family_name=member.get("LastName"),
+            )
+
+            p.add_source(
+                self.source.url, note="graphql endpoint for member information"
+            )
+
```
chrisyamas, line 49: Let's also add a link for HTML so users can see the HTML pages. Since there are not individual URL paths for each member, it's fine to provide the path for the page with the tabs with leaders and members.

So if their chamber is lower, link is:
"https://alison.legislature.state.al.us/house-leaders-members"

And if upper:
"https://alison.legislature.state.al.us/senatee-leaders-members"

Similar to with committee scraper, let's just make the note="homepage"

chrisyamas (changes requested):
Overall great job, especially it being your first time with one of the people scrapers. Nice work taking advantage of the GraphQL query.

Just a few minor changes, and then it will be ready to merge!

a-cat-named-snowball:
Thanks, those changes have been made.

chrisyamas (approved):

PR #4520, opened 2023-03-03
bugfix for UT events.py bills

a-cat-named-snowball:
Changes made:
- Modified regex to only match these prefixes ['SJR', 'HCR', 'HB', 'HR', 'SCR', 'SB', 'HJR', 'SR'] and not others
- Adjusted bill id format to "HB 123" rather than "HB0123"


chrisyamas (approved):
Data looks good! Merging.

PR #4523, opened 2023-03-04
Bugfix for ND events

a-cat-named-snowball:
Only one bill was being added per event. This was caused by using the event time as the item_key for each bill. Bills were stored in a set so duplicate keys caused previous bills to be deleted.

I fixed this by adding the bill name to the key.

I also updated bill description to use the agenda time instead of the meeting start time since I was changing that area of the code. If this isn't correct I can change it back.

### DIFF
--- a/scrapers/nd/events.py
+++ b/scrapers/nd/events.py
@@ -2,7 +2,6 @@ import re
 import pytz
 import logging
 import dateutil.parser
-import datetime
 import requests
 import lxml.html
 from spatula import HtmlPage
@@ -32,6 +31,7 @@ class EventConsolidator(object):
             local_date = dateutil.parser.parse(date_time)
 
             item_date, item_time = str(local_date).split()
+            agenda_time = item["agenda_time"]
             bill_name = item["bill_name"]
             item_id = f"{item_time}+{bill_name}"
 
@@ -54,6 +54,7 @@ class EventConsolidator(object):
             agenda_item_details = {
                 "description": item["description"],
                 "bill_name": item["bill_name"],
+                "agenda_time": agenda_time,
                 "sub_com": item["sub_com"],
             }
             self.events[event_key][item_id] = []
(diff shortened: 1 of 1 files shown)


scrapers/nd/events.py
```diff
@@ -47,15 +49,16 @@ def consolidate(self):
...
 
             agenda_item_details = {
                 "description": item["description"],
                 "bill_name": item["bill_name"],
+                "agenda_time": agenda_time,
                 "sub_com": item["sub_com"],
             }
-            self.events[event_key][item_time] = []
-            self.events[event_key][item_time].append(agenda_item_details)
+            self.events[event_key][item_id] = []
```
johnseekins, line 60: I know this is the code from before, but it seems like:
```
self.events[event_key][item_id] = [agenda_item_details]
```
would do the same thing with less code.

Also, if this particular list only ever gets one item (which seems to be the case when you create an empty list before assigning one item to it...should it even be a list?

An alternative that's less likely to overwrite existing items:
```
if item_id in self.events[event_key]:
    self.events[event_key][item_id].append(agenda_item_details)
else:
    self.events[event_key][item_id] = [agenda_item_details]
```

scrapers/nd/events.py
```diff
@@ -85,10 +88,9 @@ def create_events(self):
             for item_key in self.events[event]["item_keys"]:
                 agenda = self.events[event][item_key]
                 for item in agenda:
-                    time = datetime.datetime.strptime(item_key, "%H:%M:%S").strftime(
-                        "%I:%M %p"
-                    )
-                    descr_with_time = f"[{time}]: {item['description']}"
+                    print(item)
```
chrisyamas, line 91: Remove print statement

chrisyamas (changes requested):
Data looks good, thanks for your work to get us all those missing bill associations!

I agree with @johnseekins on implementing the alternative approach to handling adding the `agenda_item_details`. Once you've done that and removed the print statement, this will be ready to merge.

scrapers/nd/events.py
a-cat-named-snowball: I looked further into it and you're correct, a list isn't required. Simplified with https://github.com/openstates/openstates-scrapers/pull/4523/commits/ba50645163f2eedaf6ab4d5e21828e7ec43d6b44

scrapers/nd/events.py
a-cat-named-snowball: Fixed with https://github.com/openstates/openstates-scrapers/pull/4523/commits/ba50645163f2eedaf6ab4d5e21828e7ec43d6b44

chrisyamas (approved):
Looks great! Merging now.

PR #4524, opened 2023-03-05
Bugfix for RI events missing first bill

a-cat-named-snowball:
The first bill was being skipped because the header was being interpreted as part of the bill text. There was a check to skip any bills that included "SCHEDULED FOR", which is always part of the header.

This fix checks the html tags so it doesn't bundle the header text into the bill name.

While I was working on this, I noticed that there were also some bill names being scraped that I think are incorrect, like "50", "18", "S0115". Should I try to fix these also?

### DIFF
--- a/scrapers/ri/events.py
+++ b/scrapers/ri/events.py
@@ -141,11 +141,13 @@ class RIEventScraper(Scraper, LXMLMixin):
         bill_links = page.xpath('//a[contains(@href,"/BillText/")]/@href')
         linked_bills = set()
 
-        bill_id_re = re.compile(r"\/(\w+\d+)\.pdf", flags=re.IGNORECASE)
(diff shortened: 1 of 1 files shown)


johnseekins:
> While I was working on this, I noticed that there were also some bill names being scraped that I think are incorrect, like "50", "18", "S0115". Should I try to fix these also?

We always like bug fixes. Please do!

johnseekins (approved):

a-cat-named-snowball:
All bill ids should now be consistently formatted. This change also fixes a bug where numbers inside a set of parentheses, e.g. (100), were incorrectly recognized as a bill id.

chrisyamas (approved):
Great work, and thank you for contributing with additional fixes than the initial problem domain. Also appreciate your informative comments and the regexes compiled before the for loops.

Data looks good, and this PR is ready for merge.

PR #4525, opened 2023-03-05
Update AR and CA people scrapers

sacerdoted:
Collect California Assemblymembers from HTML elements instead of the table (which is no longer used)


scrapers_next/ca/people.py
```diff
@@ -17,11 +18,17 @@ def process_item(self, item):
         name = CSS("a").match(item)[2].text_content()
```
NewAgeAirbender, line 18: `name` could also use an update on the selector because it's currently pulling all names as "Committee Membership" since it's pulling the second link that exists in that row

scrapers_next/ca/people.py
```diff
@@ -56,17 +63,19 @@ def process_item(self, item):
...
             if "District Office" in h3_text:
-                dist_offs = header.getnext().text_content().strip()
                 dist_list = []
-                if len(dist_offs):
-                    dist_list = [x for x in dist_offs.split(" edit") if len(x)]
+                i = 6
+                while i < len(header):
+                    dist_list.append(header[i].text_content().strip())
+                    i = i + 1
+
```
NewAgeAirbender, line 78: Interesting approach! This adds a few too many offices though since the index starts at a paragraph too high to match. I've gotten this to work if I set it to 
`i = 8
while i < (len(header)-1)`

scrapers_next/ca/people.py
```diff
@@ -5,7 +5,8 @@
 
 class AssemblyList(HtmlListPage):
     source = "https://www.assembly.ca.gov/assemblymembers"
-    selector = CSS("table tbody tr", num_items=80)
+    # selector = CSS("table tbody tr", num_items=80)
+    selector = CSS("div.rt__row > div", num_items=80)
```
NewAgeAirbender, line 9: Can just remove the commented out `selector`. We appreciate leaving it but if we need to see the history of a selector we can just check the history of the file on github. Also the selector shouldn't include the `> div` since we set other selectors outside of that `div` later

NewAgeAirbender (changes requested):
Really solid start! Thank you for taking this on, I recommend a few small tweaks but then it'll work as it should.

sacerdoted:
Changes made per suggestions.

scrapers_next/ar/people.py
```diff
@@ -131,3 +132,137 @@ def process_item(self, item):
...
+            "Biography": "",
+        }
+
+        for row in rows:
+            try: 
+                text = row.text_content()
+                text = text.split(": ")
+                if (len(text) < 2):
+                    continue
+                name = text[0]
```
NewAgeAirbender, line 184: This use of `name` is overwriting the actual name of the Senator since it is past when `p` is created

scrapers_next/ar/people.py
```diff
@@ -131,3 +132,137 @@ def process_item(self, item):
...
+                or key == "Email"
+                or key == "District"
+                or key == "Biography"
+            ):
+                continue
+            elif table[key] != "":
+                # remove the colon at the end
+                p.extras[key.lower()] = table[key]
+        try:
+            address = CSS(".col-md-12 p b").match_one(self.root).text_content()
```
NewAgeAirbender, line 239: Wouldn't you be able to use `address = table["District Address"]` here instead of getting it off the page again?

scrapers_next/ar/people.py
```diff
@@ -131,3 +132,137 @@ def process_item(self, item):
...
+        "https://senate.arkansas.gov/senators/senators-sorted-by-congressional-district-and-seniority/"
+    )
+    selector = XPath(
+        "//div[contains(@class,'col-sm-6')]"
+        "//a/@href"
+    )
+    chamber = "upper"
+
+    def process_item(self, item):
+        print("item = " + item)
```
NewAgeAirbender, line 262: Please delete print statements before PRs

scrapers_next/ar/people.py
```diff
@@ -131,3 +132,137 @@ def process_item(self, item):
...
+
+        p = PartialMember(name="", chamber="upper", url=self.source.url)
```
NewAgeAirbender, line 264: This would actually be an easy page to get the member name off of without having to do a replace on 147, since we don't want to be passing in empty strings if we don't have to

scrapers_next/ar/people.py
```diff
@@ -131,3 +132,137 @@ def process_item(self, item):
...
+        
+
+        table = {
+            "Party": "",
+            "Phone": "",
+            "Email": "",
+            "District": "",
+            "Seniority": "",
+            "Occupation": "",
+            "Party": "",
```
NewAgeAirbender, line 169: I don't think `Party` needs to be in here twice

NewAgeAirbender (changes requested):
CA looks good! AR could use a little bit of clean up, but nice work! Make sure you're running the scraper & spot checking the scraped json to catch some of the more obvious issues first

sacerdoted:
Yeah, AR needs work; didn't intend to push that upstream before getting the AR house handled too; like the state senate, they've set up their own website, which is better about providing phone numbers than the directory that's been used in the past.

sacerdoted:
Ok, cleaned up your requested items, and now includes separate scrapers for AR senate and house.  There will still be a handful of AR legislators who seem to use personal or sole-proprietorship-busniess phones not in the official directory for state business, and I'm not picking up their phone numbers, though they seem to be widely published.

NewAgeAirbender (approved):
Looks good to me

scrapers_next/ar/people.py
```diff
@@ -131,3 +131,235 @@ def process_item(self, item):
...
+            try:
+                text = row.text_content()
+                text = text.split(": ")
+                if len(text) < 2:
+                    continue
+                item_name = text[0]
+                value = text[1]
+                table[item_name] = value
+                if item_name == "Legislative Service":
+                    if value[0] != "H":
```
johnseekins, line 179: `if value.startswith("H"):`?

scrapers_next/ar/people.py
```diff
@@ -131,3 +131,235 @@ def process_item(self, item):
...
+                if c.text_content().find(k) >= 0:
+                    found = True
+            if not found:
+                if not first_committee:
+                    committees_string += ", "
+                committees_string += c.text_content().lower()
+                first_committee = False
+        p.extras["Committees"] = committees_string
+
+        if table["District Address"] != "":
```
johnseekins, line 213: `if not table["District Address"]:` catches the more general "empty" case. Should we use that?

scrapers_next/ar/people.py
```diff
@@ -131,3 +131,235 @@ def process_item(self, item):
...
+            address = table["District Address"]
+            full_address = address[:-5] + "AR " + address[-5:]
+            p.district_office.address = full_address
+        if table["Phone"] != "":
+            p.district_office.voice = table["Phone"]
+        if table["Biography"] != "":
+            p.add_link(table["Biography"], "Biography")
+
+        for key in table:
+            if (
```
johnseekins, line 223: `if key in ["Phone", "Email", "District", "Biography"]:`?

scrapers_next/ar/people.py
```diff
@@ -131,3 +131,235 @@ def process_item(self, item):
...
+    example_source = "https://www.arkansashouse.org/district/1"
+
+    def process_page(self):
+        heading = CSS(".py-sm-6").match(self.root)[0]
+        name = self.input.name
+        line2 = heading[2].text_content().split(" ")
+        party = line2[0]
+        districtName = line2[2]
+        districtNumber = ""
+        for c in districtName:
```
johnseekins, line 266: `districtNumber = "".join(c for c in districtName if c.isdigit())`?

scrapers_next/ar/people.py
```diff
@@ -131,3 +131,235 @@ def process_item(self, item):
...
+        officeContainer = XPath("//p[contains(@class,'mb-0')]").match(self.root)[1]
+        table["Phone"] = officeContainer[2].text_content()
+        addressArray = XPath("//p[contains(@class,'mb-0')]/text()").match(self.root)
+
+        address = ""
+        started = False
+        for c in addressArray[3]:
+            if c.isalnum():
+                started = True
+            if started:
```
johnseekins, line 306: Any reason `address += c` can't run in the `if c.isalnum():` block?

scrapers_next/ar/people.py
```diff
@@ -131,3 +131,235 @@ def process_item(self, item):
...
+            party=table["Party"],
+            image=image,
+            district=district,
+            email=table["Email"],
+        )
+        p.add_source(self.source.url)
+        p.add_source(self.input.url)
+
+        for key in table:
+            if (
```
johnseekins, line 331: `if key in ["Phone", "Name", "Email", "District", "Biography", "District Address", "Phone"]:`?

scrapers_next/ny/people.py
```diff
@@ -94,43 +94,16 @@ def parse_address_lines(text):
...
-        )
-    except AttributeError:
-        address_string = ""
-    if "Phone:" in phone:
-        phone = phone.split(": ")[1]
-    else:
+    text = address_block.getchildren()[0].text_content()
+    details = parse_address_lines(text)
+    phone = details["phone"]
+    if phone == None:
```
johnseekins, line 100: `If not phone:`

scrapers_next/ny/people.py
```diff
@@ -94,43 +94,16 @@ def parse_address_lines(text):
...
         phone = ""
+    address = details["address"]
+    if address == None:
```
johnseekins, line 103: `if not address:`

PR #4527, opened 2023-03-06
Fixed bill scraping for TN events.py

a-cat-named-snowball:
Fixed TN events.py so bill ids are scraped.


chrisyamas (changes requested):
Looks good overall. In testing this PR, however, I realized that House Floor Session and Senate Floor Session are also getting their agendas scraped via the process that goes to the list of Consent and Regular Calendars for each chamber. We actually want to skip floor session agenda pages entirely, and only get committee meetings.

So all this needs to be ready for merge is to have some process for skipping Floor Sessions.

chrisyamas (approved):
Looks great, thanks for your diligent work on this one!

PR #4532, opened 2023-03-07
Bugfix for TX events

a-cat-named-snowball:
- Prevented duplicate events from being added.
- Update regex so correct bill ids are extracted.

### DIFF
--- a/scrapers/tx/events.py
+++ b/scrapers/tx/events.py
@@ -57,7 +57,7 @@ class TXEventScraper(Scraper, LXMLMixin):
             chair = metainfo["CHAIR"]
 
         plaintext = re.sub(r"\s+", " ", plaintext).strip()
-        regexp = r"(S|J|H)(B|M|R) (\d+)"
+        regexp = r"(SJR|HCR|HB|HR|SCR|SB|HJR|SR) (\d+)"
         bills = re.findall(regexp, plaintext)
 
         event = Event(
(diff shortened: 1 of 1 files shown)


scrapers/tx/events.py
```diff
@@ -9,23 +9,44 @@
...
 
 class TXEventScraper(Scraper, LXMLMixin):
     _tz = pytz.timezone("US/Central")
 
+    events_seen = set()
+
+    # Checks if an event is a duplicate.
+    # Events are considered duplicate if they have the same
+    # name, date, start time, and end time
+    def is_duplicate(self, event):
```
chrisyamas, line 23: This is a good symptom-treating approach for preventing the duplicates. Were able you able to diagnose what was causing the duplicates in the first place, or has that remained a mystery so far?

scrapers/tx/events.py
```diff
@@ -9,23 +9,44 @@
 
 import pytz
 
+bill_re = re.compile(r"(SJR|HCR|HB|HR|SCR|SB|HJR|SR) (\d+)")
```
chrisyamas, line 12: Appreciate the updated pattern and pre-compiled regex!

chrisyamas (commented):
Before we merge, I just wanted to see if you noticed what is the ultimate cause of the duplicates?

Otherwise looks good. Just wanted to get your take on this before we proceed.

PR #4534, opened 2023-03-08
bugfix for CT events.py

a-cat-named-snowball:
Added bill id scraping from pdfs.

### DIFF
--- a/scrapers/ct/events.py
+++ b/scrapers/ct/events.py
@@ -8,15 +8,37 @@ import pytz
 from .utils import open_csv
 
 
-class CTEventScraper(Scraper):
+from spatula import PdfPage, URL
+import re
 
+bill_re = re.compile(r"(SJ|HJ|HB|HR|SB|SR)\s{0,10}0*(\d+)")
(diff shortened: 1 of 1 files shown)


chrisyamas (changes requested):
Currently the scraper is going back and getting events from past years. Could you modify the scraper to only scrape events that occur on or after the session `start_date`? It should be accessible from the `__init__.py` in that directory, the dictionary with the active session being on lines 102-109 of that file:
```
        {
            "_scraped_name": "2023",
            "identifier": "2023",
            "name": "2023 Regular Session",
            "start_date": "2023-01-04",
            "end_date": "2023-05-04",
            "active": True,
        },
```
If you have trouble accessing the `start_date` that way, then an acceptable solution for now would be to include session year of `2023` somewhere clearly marked near the top of the scraper, and only have the scraper get events that occur during or after that year (or the specific date of Jan 1, 2023, etc.)

chrisyamas (approved):
Nice solution, ready for merge!

PR #4537, opened 2023-03-08
Bugfix OK events.py

a-cat-named-snowball:
Updated to scrape linked pdf for bill ids.


chrisyamas (approved):
Data looks good! And nice catch making the change on line 121.

PR #4539, opened 2023-03-08
Fixed xpath selectors for bills in NH events.py

a-cat-named-snowball:
Fixed xpath selectors and formatted bill ids


chrisyamas (approved):
Looks good! Merging.

PR #4540, opened 2023-03-09
Updated NV events.py

a-cat-named-snowball:
Switched over to scraping this page: https://www.leg.state.nv.us/App/NELIS/REL/82nd2023/Meetings

### DIFF
--- a/scrapers/nv/events.py
+++ b/scrapers/nv/events.py
@@ -93,7 +93,7 @@ class CurrentMeetings(HtmlPage):
                 # indicate that this section is cancelled or floor meeting
                 if len(cur_row) == 3:
                     meetings.append(cur_row)
-                # Reset cur_row so leftover elements don't get mixed in with set of divs
+                # Reset cur_row so leftover elements don't get mixed in with next set of divs
                 cur_row = []
 
         for i in meetings:
--- a/scrapers/nv/events.py
+++ b/scrapers/nv/events.py
@@ -1,82 +1,122 @@
 import pytz
 import datetime
-import lxml
+
+# import lxml
 
 from utils import LXMLMixin
-from utils.events import match_coordinates
+
+# from utils.events import match_coordinates
 from openstates.scrape import Scraper, Event
+
+from spatula import HtmlPage, PdfPage, URL, XPath
+import json
+import re
+
+
+bills_re = re.compile(
+    r"(SJR|AR|AJR|IP|SCR|SB|ACR|SR|AB)\s{0,5}0*(\d+)", flags=re.IGNORECASE
+)
+
+
+class Agenda(PdfPage):
+    def process_page(self):
+        # Simplify text to make regex simpler
+        self.text = self.text.replace(". ", "").replace(".", "")
+
+        # Find all bill ids
+        bills = bills_re.findall(self.text)
+
+        # Format bills before yielding
+        formatted_bills = set()
+        for alpha, num in bills:
+            formatted_bills.add(f"{alpha.upper()} {num}")
+
+        yield from formatted_bills
+
+
+class Meetings(HtmlPage):
+    source = "https://www.leg.state.nv.us/App/NELIS/REL/82nd2023/Meetings"
+
+    def process_page(self):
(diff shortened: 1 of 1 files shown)


scrapers/nv/events.py
```diff
@@ -1,82 +1,167 @@
...
+            date = datetime.datetime.strptime(date, "%B %d, %Y %I:%M %p")
+
+            # Third element contains: a list of locations
+            locations = XPath(".//li/text()").match(i[2])
+
+            # Build event, scrape bill ids from agenda, and yield it
+            event = Event(
+                start_date=self._TZ.localize(date),
+                name=title,
+                location_name=locations[0],
```
chrisyamas, line 150: Can you add a replace statement to remove the no-break-space that is currently showing up in the location name between "Carson" and "City" on any events? `"\u00a0"` needs to be replaced with single space `" "`.

chrisyamas (changes requested):
Just the one change needed on no-break-space. Otherwise, looks great! Nice job, especially considering the amount of modification that was needed here.

scrapers/nv/events.py
```diff
@@ -1,82 +1,167 @@
...
+
+class AgendaStartTime(PdfPage):
+    def process_page(self):
+        # Simplify text to make regex simpler
+        self.text = self.text.replace(". ", "").replace(".", "")
+
+        start = start_time_re.search(self.text)
+        if not start:
+            # Default to 12am if no start time is found
+            yield "12:00 am"
```
johnseekins, line 27: A single item generator is no different from a regular `return`. Probably just return values here.

scrapers/nv/events.py
```diff
@@ -1,82 +1,167 @@
...
+        # Simplify text to make regex simpler
+        self.text = self.text.replace(". ", "").replace(".", "")
+
+        # Find all bill ids
+        bills = bills_re.findall(self.text)
+
+        # Format bills before yielding
+        formatted_bills = set()
+        for alpha, num in bills:
+            formatted_bills.add(f"{alpha.upper()} {num}")
```
johnseekins, line 44: Clever way to dodge duplicates or just a more complicated way to say:

```
        for alpha, num in bills:
            yield f"{alpha.upper()} {num}")
```
?

scrapers/nv/events.py
```diff
@@ -1,82 +1,167 @@
...
+        # divs and places them into the meetings variable
+        for i in self.root.getchildren():
+            if i.tag == "div":
+                cur_row.append(i)
+            elif i.tag == "hr":
+                # Only add if all three elements are present - missing elements
+                # indicate that this section is cancelled or floor meeting
+                if len(cur_row) == 3:
+                    meetings.append(cur_row)
+                cur_row = []
```
johnseekins, line 97: Are you _sure_ you want to empty this higher scoped variable here?

scrapers/nv/events.py
johnseekins: Oh. Except this needs `yield` to fulfill the object's contract, right? Never mind.

scrapers/nv/events.py
a-cat-named-snowball: My intent was to prevent duplicates

johnseekins: Cool. Just making sure. Thanks!

scrapers/nv/events.py
a-cat-named-snowball: Yes - If there aren't exactly 3 elements in cur_row when a <hr> is found, then cur_row should be emptied without adding it to the meetings list.

johnseekins: Seems reasonable. Can you add a quick note explaining why you're emptying the variable? I could see someone coming back to this later and being confused.

scrapers/nv/events.py
a-cat-named-snowball: Thanks, fixed with https://github.com/openstates/openstates-scrapers/pull/4540/commits/3fd4f2a413896dffb5f64f83bc51d257ceaac8f0

chrisyamas (approved):

PR #4541, opened 2023-03-10
Changed senate scraping for MS events.py

a-cat-named-snowball:
Changed the site used to scrape senate events and added regex to capture bill ids in house scraper

### DIFF
--- a/scrapers/ms/events.py
+++ b/scrapers/ms/events.py
@@ -37,7 +37,7 @@ class SenateAgendaPdf(PdfPage):
         event = None
 
         # Strip all lines and remove empty lines
-        lines = [line.strip() for line in self.text.splitlines() if line.strip() != ""]
+        lines = [line.strip() for line in self.text.splitlines() if line.strip()]
 
         i = 0
         event = None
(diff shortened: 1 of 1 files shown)


scrapers/ms/events.py
```diff
@@ -4,27 +4,119 @@
...
+        yield from SenateAgendaPdf(source=URL(pdf_link)).do_scrape()
+
+
+# Parses events from a pdf
+class SenateAgendaPdf(PdfPage):
+    def process_page(self):
+        event = None
 
+        # Strip all lines and remove empty lines
+        lines = [line.strip() for line in self.text.splitlines() if line.strip() != ""]
```
chrisyamas, line 40: Very minor stylistic change, but this can be modified to the more Pythonic:
```
lines = [line.strip() for line in self.text.splitlines() if line.strip()]
```
- given that empty strings are false-y.

chrisyamas (changes requested):
Only that minor change request and then it'll be ready for merge. Nice job building this into a dual source-type scraper!

scrapers/ms/events.py
a-cat-named-snowball: That's good to know - thank you. Changed with https://github.com/openstates/openstates-scrapers/pull/4541/commits/1e7ce3f30257842675348a025625519129e93c19

chrisyamas (approved):

PR #4543, opened 2023-03-14
2023 03 14 fixes

johnseekins:
Fix ID events with empty description
Handle empty events in KS
Handle duplicate events in MI
Handle poorly formatted times in MS events
Handle strange empty arrays in MS bills
Fix URL and empty events for WI events


bfossen-ce (approved):
LGTM

PR #4547, opened 2023-03-15
Added bill scraping to VT events.py

a-cat-named-snowball:
Added bill scraping functionality to VT events.py


chrisyamas (approved):
Great solution for the duplicate bills and multiple endpoints issue with use of the helper function. I rearranged its placement because we typically like to keep our helper functions at the top of the file. Data looks good though, and I'm merging this.

PR #4556, opened 2023-03-17
Bugfixes for SC events.py

a-cat-named-snowball:
- Modified to scrape all chambers if no chamber is provided
- Fixed some time parsing issues
- Added bill id scraping from agenda pdfs

### DIFF
--- a/scrapers/sc/events.py
+++ b/scrapers/sc/events.py
@@ -108,9 +108,9 @@ class SCEventScraper(Scraper):
         if chamber:
             yield self.scrape(chamber, session)
         else:
-            yield self.scrape_single_chamber("legislature")
-            yield self.scrape_single_chamber("upper")
-            yield self.scrape_single_chamber("lower")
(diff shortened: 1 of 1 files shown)


scrapers/sc/events.py
```diff
@@ -83,6 +105,14 @@ def get_stream_id(self, onclick: str) -> str:
         return re.findall(r"live_stream\((\d+)|$", onclick)[0]
 
     def scrape(self, chamber=None, session=None):
+        if chamber:
+            yield self.scrape(chamber, session)
+        else:
+            yield self.scrape_single_chamber("legislature", session)
```
johnseekins, line 111: Probably want `yield from` here so we aren't sending generators of generators back.

chrisyamas (approved):
Glad that your trusty `Agenda()` spatula class made an appearance again. Also, `scrape_single_chamber()` was a good solution to the issue. Data checks out, and I'm merging it.

PR #4557, opened 2023-03-17
AL Bills: scrape past first page

showerst:


scrapers/al/bills.py
```diff
@@ -60,70 +25,132 @@ def scrape(self, session):
...
 
-        # WARNING: 2023 session id is currently hardcoded
-        json_data = {
-            "query": '{allInstrumentOverviews(instrumentType:"B", instrumentNbr:"", body:"", sessionYear:"2023", sessionType:"2023 Regular Session", assignedCommittee:"", status:"", currentStatus:"", subject:"", instrumentSponsor:"", companionInstrumentNbr:"", effectiveDateCertain:"", effectiveDateOther:"", firstReadSecondBody:"", secondReadSecondBody:"", direction:"ASC"orderBy:"InstrumentNbr"limit:"15"offset:"0"  search:"" customFilters: {}companionReport:"", ){ ID,SessionYear,InstrumentNbr,InstrumentSponsor,SessionType,Body,Subject,ShortTitle,AssignedCommittee,PrefiledDate,FirstRead,CurrentStatus,LastAction,ActSummary,ViewEnacted,CompanionInstrumentNbr,EffectiveDateCertain,EffectiveDateOther,InstrumentType,IntroducedUrl }}',
-            "operationName": "",
-            "variables": [],
-        }
+        offset = 0
+        limit = 10000
+        while True:
```
johnseekins, line 30: These sorts of loops always make me nervous. Is there any way we can have a "for sure" break?

showerst: Hmm it'll have to break at 
```
            # no need to paginate again if we max the last page
            if len(page["data"]["allInstrumentOverviews"]) < limit:
                return
```

I could hardcode in a guard at 50k or something, but open to more elegant ideas.

scrapers/al/bills.py
```diff
@@ -60,70 +25,133 @@ def scrape(self, session):
...
-                    row["PrefiledDate"], "%m/%d/%Y"
-                )
-                action_date = self.tz.localize(action_date)
-                bill.add_action(
-                    chamber=chamber,
-                    description="Filed",
-                    date=action_date,
-                    classification="filing",
-                )
+        # Can this be ANDED together with the other graphql query?
```
jessemortenson, line 130: @showerst I don't think so: all the properties on the `InstrumentOverviews` type (which is returned by the prior query request, `allInstrumentOverviews`) are scalar, there is no property that links to another object/interface. Same with what I assume is a related type, `InstrumentSummary`. Looking at the AL graphql schema I don't see much support for these kind of nested data that you normally expect in a graphql schema.

jessemortenson:
Overall this looks reasonable to me, Tim, given the constrains of the AL graphql endpoint. My main feedback would be to find a graphql client library to import so that you don't have to construct the POST body JSON in quite such a manual fashion. The ` "query"` parameter of the `json_data` dict is pretty rough to read and pick apart (I loaded the query into Insomnia to play with). I think that would be a lot easier using syntax from, say, the [gql package](https://github.com/graphql-python/gql#basic-usage). I haven't actually used that hands-on, but at a glance it looks nicer to reason about than the raw client-side JSON.

If you want to [introspect](https://graphql.org/learn/introspection/) the schema, a couple helpful graphql queries are:

```
{
  __schema {
    types {
      name
      fields {
        name
        type {
          name
          kind
        }
      }
    }
  }
}
```

```
{
  __type(name: "Query") {
    name
    fields {
      name
      type {
        name
        kind
        ofType {
          name
          kind
        }
      }
    }
  }
}
```

showerst:
Thanks @jessemortenson -- shame we can't distill that down to one call. At least they allow decent paging limits. 

Some rainy day when Govhawk is due for a scraper image rebuild I'll PR adding a decent gql library to the poetry reqs.

PR #4570, opened 2023-03-23
NV bills: shortener for too long bill title

chrisyamas:
Function created to shorten bill title in case when it's length is longer than the acceptable 300 characters.

Only known case in with SJR7-2021, where bill title from source is:
- "Proposes to amend the Nevada Constitution to remove the constitutional provisions governing the election and duties of the Board of Regents of the State University and to authorize the Legislature to provide by statute for the governance of the State University and for the auditing of public institutions of higher education in this State. (BDR C-944)"
- this solution shortens it to:
   - "Proposes to amend the Nevada Constitution to remove the constitutional provisions governing the election and duties of the Board of Regents of the State University and to authorize the Legislature to provide by statute for the governance of the State University and for the auditing... (BDR C-944)"

### DIFF
--- a/scrapers/nv/bills.py
+++ b/scrapers/nv/bills.py
@@ -265,6 +265,7 @@ class BillTabDetail(HtmlPage):
                 )
                 return
 
+        # Only known case where bill title is over 300 characters
         if self.input.identifier == "SJR7-2021" and self.input.session == "82":
             short_title = shorten_bill_title(short_title)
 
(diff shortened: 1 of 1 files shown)


scrapers/nv/bills.py
```diff
@@ -252,6 +265,10 @@ def process_page(self):
                 )
                 return
 
+        # Only known case where bill title is over 300 characters
```
johnseekins, line 268: Would it be better to just do the more general `if len(short_title) > 300:`?

PR #4572, opened 2023-03-24
Expanded address abbreviations in AK events.py

a-cat-named-snowball:
Expanded room names and address abbreviations into full addresses


chrisyamas (approved):
Looks great, nice job catching all cases on this so we could maximize the accuracy of the location data. Only change I made was to make all room references uppercase to ensure consistency of format.

Approving and merging!

PR #4579, opened 2023-03-28
OH events: dedupe key add

chrisyamas:
Adds the event `name` as `dedupe_key` to prevent `DuplicateItemError` during import.


bfossen-ce (approved):
LGTM

PR #4582, opened 2023-03-30
bug fixes IL events.py

a-cat-named-snowball:
Changed event descriptions and added a space to bill ids

### DIFF
--- a/scrapers/il/events.py
+++ b/scrapers/il/events.py
@@ -17,6 +17,8 @@ chamber_names = {
     "lower": "House",
 }
 
+bill_re = re.compile(r"(\w+?)\s*0*(\d+)")
+
 
 class IlEventScraper(Scraper):
     localize = pytz.timezone("America/Chicago").localize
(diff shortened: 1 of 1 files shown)


scrapers/il/events.py
```diff
@@ -37,8 +44,12 @@ def scrape_page(self, url):
...
 
         where = metainf["Location:"]
-        subject_matter = metainf["Subject Matter:"]
-        description = "{}, {}".format(ctty_name, subject_matter)
+
+        description = f"{chamber} {ctty_name}"
+        # Remove committee suffix from names
+        committee_suffix = " Committee"
+        if description.endswith(committee_suffix):
+            description = description[: -len(committee_suffix)]
```
chrisyamas, line 52: The solution is working well for all events except those with "Hearing Notice For" in the description. That string should be removed in a way that also does not result in the chamber being listed twice.

Two examples of event `description` using current solution:
- "Senate Hearing Notice For Environment and Conservation"
- "Senate Hearing Notice For Senate Special Committee on Pensions"

After update to this PR, these cases should instead be:
- "Senate Environment and Conservation"
- "Senate Special Committee on Pensions"

chrisyamas (changes requested):
Just that one requested change. Everything else looks good!

scrapers/il/events.py
a-cat-named-snowball: Fixed with https://github.com/openstates/openstates-scrapers/pull/4582/commits/d4bfa46906d683c2357320c3bd1c23f0246fa7c9

chrisyamas (approved):

PR #4585, opened 2023-04-03
RI events: dedupe key made unique

chrisyamas:
This PR makes each event's `dedupe_key` more unique, which allows for it to function as intended in preventing `DuplicateItemError` during events import.

### DIFF
--- a/scrapers/ri/events.py
+++ b/scrapers/ri/events.py
@@ -29,9 +29,8 @@ replace = {
 class RIEventScraper(Scraper, LXMLMixin):
     _tz = pytz.timezone("US/Eastern")
     found_events = False
-    event_keys = set()
 
-    def scrape_agenda(self, chamber, url):
+    def scrape_agenda(self, chamber, url, event_keys):
         page = self.lxmlize(url)
         # Get the date/time info:
         date_time = page.xpath("//table[@class='time_place']")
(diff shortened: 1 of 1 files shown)


johnseekins:
Typically when we add a dedupe key, we also add a `set()` that holds all the unique items in it so we can more gracefully handle duplicated items. I'm not sure that's actually necessary, though. If you _want_ to pursue that, an example: https://github.com/openstates/openstates-scrapers/pull/4550/files
If not, I think this should work fine.

johnseekins (approved):

NewAgeAirbender (approved):

chrisyamas:
> Typically when we add a dedupe key, we also add a `set()` that holds all the unique items in it so we can more gracefully handle duplicated items. I'm not sure that's actually necessary, though. If you _want_ to pursue that, an example: https://github.com/openstates/openstates-scrapers/pull/4550/files If not, I think this should work fine.

Thanks for that note and code reference, @johnseekins

I went ahead and added a commit with a simple set, conditional check and warning log like your example.

scrapers/ri/events.py
```diff
@@ -92,14 +93,21 @@ def scrape_agenda(self, chamber, url):
...
 
         event.add_document("Agenda", url, media_type="text/html", on_duplicate="ignore")
         event.add_source(url)
-        event.dedupe_key = event_desc
+
+        # Unique key for preventing triggering of duplicate item error
+        event_details_key = f"{event_desc}#{event_start}#{where}"
+
+        if event_details_key in self.event_keys:
+            self.warning(f"Duplicate event {event.dedupe_key}")
```
johnseekins, line 106: Technically, this won't skip the duplicate. We'll just get a warning and then when we go to insert we'll still get the duplicate item error.
We probably want to re-do line 190 like (roughly) so:

```
events = set()
for chamber in chambers:
    for event in self.scrape_chamber(chamber):
        if event in events:
            self.warning(f"Duplicate event: {event}")
            continue
        else:
            events.add(event)
        yield event
````

johnseekins: This ensures we don't send duplicate events to the insert step, but we're still notified that they're happening, which should help with potential debugging later.

chrisyamas: I dig it.

PR #4593, opened 2023-04-17
SC events: event time fix, dedupe key add

chrisyamas:
This PR addresses two issues: one with event times, the other with duplication.

Event times
------------
South Carolina events scraper was failing on a `ValueError` when a string like `"Immediately after EOC Full Committee"` was being passed to `datetime.strptime()` method instead of a time string like `"1:00 pm"`.

This would occur in cases when the time for an event is determined by the start or end time of a prior event. This solution stores a reference to the immediately-prior-event's start time, and has the aforementioned dependent cases simply default to the prior event's start time (as that is the best estimate that can be provided, and we do not want to provide a time that could be later than the dependent event's actual start time).

Duplication
-----------
Dedupe key added to event, with an `event_keys` set to prevent actual duplicates.

### DIFF
--- a/scrapers/sc/events.py
+++ b/scrapers/sc/events.py
@@ -80,6 +80,8 @@ class SCEventScraper(Scraper):
 
     prior_event_time_string = None
 
+    event_keys = set()
+
     def get_page_from_url(self, url):
         """
         Get page from the provided url and change to string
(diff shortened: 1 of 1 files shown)


bfossen-ce (approved):
LGTM

PR #4597, opened 2023-04-19
MT bills: (temporarily) skip committee vote PDFs

chrisyamas:
This PR refactors Montana bills scraper's `_parse_votes()` method into `_parse_pdf_votes()` and `_parse_html_votes()`, and temporarily skips committee vote PDFs (leaving `_parse_pdf_votes()` uncalled) during scraper run.

The idea behind this is that most (if not all) committee vote PDFs were failing to be parsed, and the many failed attempts is likely what has been causing the scraper as a whole to fail.

Note on the *temporary* nature of this change: separately from this PR, I am locally testing the PDF links to determine the parsing issue. Once I have determined it, I will write a solution and create a new PR that adds committee vote PDF parsing back into the scraper's functions.

### DIFF
--- a/scrapers/mt/bills.py
+++ b/scrapers/mt/bills.py
@@ -498,33 +498,43 @@ class MTBillScraper(Scraper, LXMLMixin):
                         chamber=chamber, date=date, action=action, vote_url=vote_url
                     )
 
-                    # Update the vote object with voters..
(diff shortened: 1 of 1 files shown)


bfossen-ce (approved):
LGTM

PR #4606, opened 2023-04-23
MO events: regex for multiple agenda item bill nums

chrisyamas:
Adds regex pattern capture to cover all current cases of when multiple bills items with the same prefix are listed on as items on source agenda "&".

Examples:
- "HJRs 33 & 45" or "HBs 882 & 518"

Regex is a cleaner and more accurate way to extract these bill ids to add them to `Event` object's agenda items.

### DIFF
--- a/scrapers/mo/events.py
+++ b/scrapers/mo/events.py
@@ -9,27 +9,47 @@ from openstates.scrape import Scraper, Event
 from openstates.exceptions import EmptyScrape
 
 
-# Adds bills from text that may contain one or multiple bills
-# e.g. "SB 123" or "SBs 123 & 341"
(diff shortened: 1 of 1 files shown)


scrapers/mo/events.py
```diff
@@ -9,27 +9,47 @@
 from openstates.exceptions import EmptyScrape
 
 
-# Adds bills from text that may contain one or multiple bills
-# e.g. "SB 123" or "SBs 123 & 341"
+class UnknownBillStringPattern(BaseException):
```
johnseekins, line 12: Nice use of exception classes.

chrisyamas: Thanks!

PR #4607, opened 2023-04-23
Fix image scrape for Rep. Charlie Conrad in OR

dsarhadian:
Representative Charlie Conrad's [page](https://www.oregonlegislature.gov/conrad) does not have the representative's image in the selector specified in the scraper. 

This commit manually specifies his image.


NewAgeAirbender (approved):
Thanks for fixing that!

PR #4611, opened 2023-04-25
AR bills: fix yield statements in refactoring

chrisyamas:
Adds `yield bill` command that was missing from `scrape_bill_page()` method, and necessary for scraper to run successfully.

### DIFF
--- a/scrapers/ar/bills.py
+++ b/scrapers/ar/bills.py
@@ -15,11 +15,6 @@ TIMEZONE = pytz.timezone("US/Central")
 
 
 class ARBillScraper(Scraper):
-    """
-    The challenge here is that I need to comment out all
-
-    """
-
     ftp_user = ""
     ftp_pass = ""
     bills = {}
--- a/scrapers/ar/bills.py
+++ b/scrapers/ar/bills.py
@@ -93,7 +93,7 @@ class ARBillScraper(Scraper):
                     primary=True,
                 )
 
-            # need year before if its a fiscal or special session beyond first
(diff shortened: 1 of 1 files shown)


scrapers/ar/bills.py
```diff
@@ -326,6 +326,8 @@ def scrape_bill_page(self, bill):
                 media_type="application/pdf",
             )
 
+        yield bill
```
johnseekins, line 329: If we're only supposed to be returning one item from this function, let's just do a regular `return`. `yield` has specific semantics that can be messy when you only need one object.

showerst:
I might be missing context here, but it looks like the scraper doesn't yield there because we want to add actions to the bills after that step during self.scrape_actions and then yield them all at the end. Is there some bug preventing them from getting emitted at the end, or are you trying to swap the order?

johnseekins (approved):

chrisyamas:
> I might be missing context here, but it looks like the scraper doesn't yield there because we want to add actions to the bills after that step during self.scrape_actions and then yield them all at the end. Is there some bug preventing them from getting emitted at the end, or are you trying to swap the order?

Hi @showerst, I just went about it wrong. Initially it was using `yield from self.scrape_bill()` before then scraping actions, but it was only actually yielding votes (while also instantiating bills and adding them to a dictionary, that would then be accessed to add actions to each bill during the `scrape_actions()` method). After I refactored the scraper to (temporarily) remove vote event processing, I had not removed all yield statements that were inherent to that process. However, this PR has since been updated, and they have been removed accordingly.

PR #4634, opened 2023-05-18
Scrapers Utils: url_xpath exception add

chrisyamas:
This PR adds an exception clause to the `url_xpath()` method in `scrapers/utils/lxmlize.py`. The purpose is to provide more detailed and useful logging, particularly to debug errors that occur with this method when a scraper is run virtually but do not occur when it is run locally (i.e. s in AWS but not on developer's local machine).

### DIFF
--- a/scrapers/utils/lxmlize.py
+++ b/scrapers/utils/lxmlize.py
@@ -9,7 +9,7 @@ def url_xpath(url, path, verify=True, user_agent=None):
     try:
         doc = lxml.html.fromstring(res.text)
     except Exception as e:
-        logging.warning(
+        logging.error(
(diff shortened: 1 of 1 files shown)


showerst:
Rather than a print statement, can this be a ```self.warning``` or ```self.error``` ? Not sure offhand if that will work w/ the current class references.

chrisyamas:
> Rather than a print statement, can this be a `self.warning` or `self.error` ? Not sure offhand if that will work w/ the current class references.

I changed it to a `logging.warning()`, do you think that suffices?

chrisyamas:
@johnseekins I updated it to `logging.error()`

johnseekins (approved):

PR #4638, opened 2023-05-22
ME: Events: Fix for blank events

showerst:


scrapers/me/events.py
```diff
@@ -108,6 +108,11 @@ def scrape(self, session=None, start=None, end=None):
                 classification="committee-meeting",
             )
             event.dedupe_key = event_name
+
+            if name.strip() == "":
```
johnseekins, line 112: `if not name.strip()`?

PR #4639, opened 2023-05-22
NY: dedupe key length limit

chrisyamas:
This PR adds a character limit of 500 to the `dedupe_key` for NY events. This should prevent future cases of `DataImportError:value too long for type character varying(500)`.


scrapers/ny/events.py
```diff
@@ -103,7 +103,7 @@ def scrape_lower_event(self, url):
             start_date=when,
             location_name=location,
         )
-        event.dedupe_key = event_name
+        event.dedupe_key = event_name[:499]
```
jessemortenson, line 106: This is a good solution if you're very confident that event name will be unique before 500 characters long. For NY data that seems like a pretty safe assumption. So I don't think you need to change anything. Below is just for information:

If you ran into any cases where you think the event might have a really long "standard" prefix, shared by other events, you could do an alternate approach of making a hash of `event_name`. MD5 is a quick hash method that is OK to use when you don't have any security concerns (it's not secure any more), just want a unique (shorter) representation of a (longer) string. 

`cryptography` is already included in this project so you could use it to do this: https://cryptography.io/en/latest/hazmat/primitives/cryptographic-hashes/#cryptography.hazmat.primitives.hashes.Hash

jessemortenson (approved):
LGTM

scrapers/ny/events.py
johnseekins: That's a very clever idea. We could even use some code like the following to get consistent hashing:
```
import hashlib
import uuid
m = hashlib.md5()
m.update(event_name.encode("utf-8"))
new_uuid = uuid.UUID(m.hexdigest())
event_name =  str(new_uuid)
```
And this could be added to scraper utils as a general consistent hashing function for _any_ long names.

chrisyamas: I created [a separate PR](https://github.com/openstates/openstates-scrapers/pull/4643) that implements hashing into the scraper utils directory, as per John's suggestion. Once that is approaced and merged, I will push a commit to this `ny_events` branch that uses that hashing utility function.

PR #4643, opened 2023-05-24
Scraper Utils: event key hash add

chrisyamas:
This PR creates a new utility function that can be utilized in event scrapers to create a unique hashed identifier for the event's `dedupe_key`, to help facilitate duplication prevention which also eliminates risk of a `dedupe_key` exceeding allowable length.

The modification made in this PR was undertaken based on [this code suggestion](https://github.com/openstates/openstates-scrapers/pull/4639#discussion_r1201294079) from @johnseekins.

### DIFF
--- a/scrapers/utils/events.py
+++ b/scrapers/utils/events.py
@@ -1,3 +1,21 @@
+import hashlib
+import uuid
+
+
+def hash_event_key(event_key_str):
+    """
+    Used to shorten event identifier strings while maintaining uniqueness.
+    :param event_key_str: type str - event identifier of variable length
+    :return: type str - unique hash of event identifier
+    """
+    hash_val = hashlib.md5()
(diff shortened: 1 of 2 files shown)


scrapers/utils/events.py
```diff
@@ -1,3 +1,21 @@
+import hashlib
+import uuid
+
+
+def hash_event_key(event_key_str):
```
johnseekins, line 5: I could see this being useful across more than just events. Maybe this belongs in the __init__.py file with a more generic name?

chrisyamas: Made generic in `__init__` file in [72ea4a3](https://github.com/openstates/openstates-scrapers/pull/4643/commits/72ea4a3ae7d929ca1af9fcf532074299f48fb438)

PR #4652, opened 2023-06-07
NV: 2023 Second Special

chrisyamas:
The [35th (2023) Special Session](https://www.leg.state.nv.us/Session/35th2023Special/) of the Nevada Legislature began on June 07, 2023, at 8:00 AM.

This PR adds it to NV's `__init__` file.


NewAgeAirbender (changes requested):
Looks good but you also need to add the special session to the `session_slugs` in `common.py` in the nv directory

NewAgeAirbender (approved):

PR #4654, opened 2023-06-13
SC: temporary changes to scrape past 2 sessions

chrisyamas:
Adds functionality to scraper to ingest past session bill data.

This was undertaken to update Open States database after improvements were made to `name` and `abstract` collection for SC bills in [previous PR](https://github.com/openstates/openstates-scrapers/pull/4629).

Note: web archived pages were used to get the past session bill index pages, which allows for this past-session ingestion with minimal changes to the scraper. `TODO` comments have been included in the code as reminders to maintainers that an archive id will need to be manually added in the lead up to every new session start (i.e. every two years) to ensure full past-session-scrape functionality.

### DIFF
--- a/scrapers/sc/__init__.py
+++ b/scrapers/sc/__init__.py
@@ -52,6 +52,8 @@ class SouthCarolina(State):
             "name": "2019-2020 Regular Session",
             "start_date": "2019-01-08",
             "end_date": "2020-06-25",
+            # TODO: Remove below line after successful past-session scrapes
+            "active": True,
         },
         {
             "_scraped_name": "124 - (2021-2022)",
@@ -60,7 +62,8 @@ class SouthCarolina(State):
             "name": "2021-2022 Regular Session",
             "start_date": "2021-01-12",
             "end_date": "2022-05-12",
-            "active": False,
+            # TODO: Revert below value to False after successful past-session scrapes
+            "active": True,
         },
         {
             "_scraped_name": "125 - (2023-2024)",
--- a/scrapers/sc/__init__.py
+++ b/scrapers/sc/__init__.py
@@ -52,8 +52,6 @@ class SouthCarolina(State):
             "name": "2019-2020 Regular Session",
             "start_date": "2019-01-08",
             "end_date": "2020-06-25",
-            # TODO: Remove below line after successful past-session scrapes
-            "active": True,
         },
         {
             "_scraped_name": "124 - (2021-2022)",
@@ -62,8 +60,7 @@ class SouthCarolina(State):
             "name": "2021-2022 Regular Session",
             "start_date": "2021-01-12",
             "end_date": "2022-05-12",
-            # TODO: Revert below value to False after successful past-session scrapes
-            "active": True,
+            "active": False,
         },
         {
             "_scraped_name": "125 - (2023-2024)",
--- a/scrapers/sc/bills.py
+++ b/scrapers/sc/bills.py
@@ -90,6 +90,51 @@ def action_type(action):
     return None
 
 
+def get_index_url(session, chamber, chamber_letter):
+    """
+    Note: a Web Archive ID should be added during session transitions to ensure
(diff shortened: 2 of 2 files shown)


scrapers/sc/__init__.py
```diff
@@ -52,6 +52,8 @@ class SouthCarolina(State):
             "name": "2019-2020 Regular Session",
             "start_date": "2019-01-08",
             "end_date": "2020-06-25",
+            # TODO: Remove below line after successful past-session scrapes
+            "active": True,
```
jessemortenson, line 56: One alternative to doing the `active: True` juggle here, would be to create a new job in the Task Definitions file for SC, that explicitly tells the scraper to scrape the old session(s) only, on a schedule that is just once a week (or something else very sparse). (at least I think you can run a CLI command that executes a scrape for a particualr session, regardless of whether it is marked active or not)

The advantage to doing it that way (rather than here in this config), are a) downstream users of this code won't accidentally be scraping old sessions and maybe get confused, and b) setting to a sparse schedule will ensure our current session scrape still happens on the normal schedule, but the backfill job just runs once or twice before we shut it off

jessemortenson (changes requested):
Overall looks good to me except for two issues:

1. Maybe change the comments to explain the past-session-scrape code, instead of removing that code. Would allow other users to scrape past sessions if they want. Overall, I think the functionality for a scraper to be able to scrape past sessions is a nice value add. It's not worth going way out of our way to support, but we already have it here.
2. Consider scheduling the past sessions scrapes via a new job in Task Definitions, rather than (falsely) setting old sessions to active in the scraper code.

scrapers/sc/__init__.py
chrisyamas: Commit [0bc7f30](https://github.com/openstates/openstates-scrapers/pull/4654/commits/0bc7f30deade9eb71723317f6ef99855b9d16543) makes this an ongoing component of the scraper, and cleans it up into a helper function.

Tasks definitions added in [this PR](https://github.com/openstates/task-definitions/pull/84).

scrapers/sc/bills.py
```diff
@@ -90,6 +90,51 @@ def action_type(action):
...
+    # from the prior two sessions. (Archives needed or index_url leads to current session bills.)
+    web_archive_ids = {
+        "2017-2018-lower": "20181101144929",
+        "2017-2018-upper": "20181101144422",
+        "2019-2020-lower": "20201101155143",
+        "2019-2020-upper": "20201101152857",
+        "2021-2022-lower": "20221110101038",
+        "2021-2022-upper": "20221110101352",
+        # TODO: Add archive id values during transition between sessions (i.e. Dec 2024).
+        "2023-2024-lower": None,
```
jessemortenson, line 117: I can see what you're doing by padding out future sessions, which works fine. To make it a little less verbose, you could instead have a little logic that checks to see if the key exists in the `web_archive_ids` dict, and then provides a default value of `None` if it's not present: https://stackoverflow.com/a/1602964

Something like:

```
web_archive_ids = {
        "2017-2018-lower": "20181101144929",
        "2017-2018-upper": "20181101144422",
        "2019-2020-lower": "20201101155143",
        "2019-2020-upper": "20201101152857",
        "2021-2022-lower": "20221110101038",
        "2021-2022-upper": "20221110101352",
}
web_archive_id = web_archive_ids.get(f"{session}-{chamber}", None)

jessemortenson (approved):
LGTM, good improvement and I like the extended comment providing pointers. One suggestion if you want to make the code a little less verbose.

scrapers/sc/bills.py
chrisyamas: That's a really convenient optional parameter for the `get()` method that I completely forgot about. Thanks! Added in [49c5f72](https://github.com/openstates/openstates-scrapers/pull/4654/commits/49c5f722578b3298a2a71a4d1b7ed3f39dbde681).

PR #4679, opened 2023-08-16
add CITATION.cff, top 10 contributors to start

jamesturk:
Recently became aware of CFF files (see:
https://github.blog/2021-08-19-enhanced-support-citations-github/) which, similar to LICENSE, adds a helpful citation-helper to the GitHub UI.  I've been told this repo has been used/cited in papers in an ad hoc way before, so this could prove useful.

This adds a very simple software citation & I started with the top 10 contributors, glad to add others if they'd like & if anyone has an orcid & contributes they should be welcome to add to this!


showerst:
:+1:

PR #4691, opened 2023-08-30
remove bad chamber

johnseekins:


showerst:
FYI @johnseekins i think we use either "legislature" or upper in the unicameral / edge cases.

PR #4695, opened 2023-09-05
Events: Misc Add locations, improved addresses

showerst:


showerst:
@NewAgeAirbender @johnseekins FYI current NH hearing titles have a bunch of junk in them (location, ==REVISED==, etc) so this patch cleans up the names a bit. 

I tried to keep the dedupe keys the same as pre-patch, but a bunch of events will change title.

PR #4720, opened 2023-11-15
Chapter247 Bills.py is converted into scrapy

rohitb-c247:
I have studied the bills.py in North Dakota Scrapper and converted the bills.py into the scrapy .


jessemortenson:
Thank you but I'm looking for this to be a pull request into the https://github.com/openstates/scrapy-test repository, which is already set up for working with scrapy. We're still experimenting with this, and not ready to merge scrapy-style scrapers into our main repository (openstates-scrapers)

PR #4757, opened 2023-12-18
USA: votes scraper fix missing senate vote classification

jessemortenson:
Also set default year when argument not passed explicitly


showerst (approved):
Looks good to me, good idea on the default.

PR #4758, opened 2023-12-19
MO: comment out subjects temporarily

NewAgeAirbender:
We're getting 403 errors hitting the [House Bill Subject](https://house.mo.gov/LegislationSP.aspx?code=R&category=subjectindex&year=2024) page with either Bobsled or Airflow so this temporarily skips it so that we can at least get most of the bill data in for prefiles. I have a contact with the team of House Developers, so I sent an email asking if they'll have that information available on the Bill Detail XML page & will update the scraper based on his information.

### DIFF
--- a/scrapers/mo/bills.py
+++ b/scrapers/mo/bills.py
@@ -684,7 +684,8 @@ class MOBillScraper(Scraper, LXMLMixin):
             )
 
     def scrape(self, chamber=None, session=None):
-        self._scrape_subjects(session)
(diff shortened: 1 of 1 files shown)


jessemortenson:
Ah, bummer, thanks for fixing. Good idea making the request. I believe the MO House 403s come from a network block that is based on client IP address, FWIW (I am able to scrape those subject URLs from a different IP).

PR #4766, opened 2024-01-01
CT: 2024

showerst:
Don't merge until bills post.


jessemortenson:
reminder for when bills drop, we have a votes scraper for CT now that can be merged at that time as well: https://github.com/openstates/openstates-scrapers/pull/4777

PR #4772, opened 2024-01-03
NM: update votes scraper

braykuka:
## Description

`convert_pdf` function in openstates didn't parse the headers correctly in some PDF files.
I've updated NM vote scraper using PyMuPDF library.

## Dependency
- PyMuPDF 
- pandas


braykuka:
@jessemortenson you have to install 2 dependent libraries before running. Please review it and let me know if you have issues.
Thanks.

jessemortenson:
Thanks, this looks pretty good! No data issues with the votes I spot checked. 

One request, could you please change to output the URLs as `https` (instead of `http`) unless the server actually will not take an https connection? eg

```"sources": [
    {
      "url": "http://www.nmlegis.gov/Sessions/23%20Regular/votes/HJM007SVOTE.PDF",
      "note": ""
    }
  ],
  "extras": {},
  "dedupe_key": "http://www.nmlegis.gov/Sessions/23%20Regular/votes/HJM007SVOTE.PDF",
  "_id": "106205df-aa84-11ee-85a1-f956a030e9e9"
```

when I put that URL in a browser it seems the server changes it to https. We want to avoid `http` URLs unless absolutely necessary.

jessemortenson:
Also, I added a few commits:

- Add the required dependencies to pyproject.toml (used `poetry add` to add them, needed to upgrade pytz)
- Reformatted the code using PyCharm + `black` integration (our linting is done by flake and black)
- Added 2024 session, which is now up on the website to some degree, otherwise scraper won't run

braykuka:
@jessemortenson Thanks for your review. I've fixed 'https' issue.

jessemortenson:
Great, thanks! I will merge.

PR #4787, opened 2024-01-09
NJ: add 2024-2025 metadata

NewAgeAirbender:
Only merge when bills can be scraped successfully


NewAgeAirbender:
I think there may be an issue with new 2024-2025 bills being added to the bulk data file for 2022-2023, shooting them an email but updating this to just have 2022-2023 active for now.

PR #4788, opened 2024-01-10
Indiana - Remove Proxy for versions and documents.

showerst:
I haven't done a full scrape here so this might need more cleanup.


showerst:
@jessemortenson flagging that IN now makes their version PDFs public, so hopefully we can dispense with the proxies. 

I haven't checked to see if they have public vote pages yet.

PR #4792, opened 2024-01-12
MO: scrape events for MO House from XML source

braykuka:
## Description

I've updated the events scraping logic from XML source.


jessemortenson:
Looks good to me, merged!

PR #4795, opened 2024-01-16
HI: disable SSL verify, update for some site HTML changes, add 2024

showerst:
@jessemortenson @NewAgeAirbender this is a patch for HI that will work w/ any transparent proxy, to minimize code changes and requiring a specific vendor.

Using zenrows, my test command was:

```shell
HTTPS_PROXY="http://API_KEY_HERE:premium_proxy=true&proxy_country=us@proxy.zenrows.com:8001" HTTP_PROXY="http://API_KEY_HERE:premium_proxy=true&proxy_country=us@proxy.zenrows.com:8001" PYTHONPATH=scrapers poetry run os-update hi bills --scrape
```

The same also works for events.

You might also want to set  ```PYTHONWARNINGS="ignore:Unverified HTTPS request"``` to shut up the insecure request warnings.

If we decide to go this route, i'll update the documentation too.

### DIFF
--- a/scrapers/hi/__init__.py
+++ b/scrapers/hi/__init__.py
@@ -1,4 +1,5 @@
-from utils import url_xpath
+import lxml.html
+import requests
 from openstates.scrape import State
 from .events import HIEventScraper
 from .bills import HIBillScraper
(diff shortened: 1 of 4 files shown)


NewAgeAirbender:
I could get to the session list & first bill detail page last night but it failed with unexpected formatting & this morning I can't even get to the session page so they may have got me or I'm doing something, but if it works on scrape for you I say we can go for it.

showerst:
I've got a few more HTML updates now that they're posting more, let me get those into this branch.

showerst:
@NewAgeAirbender give it a try w/ the update I just pushed, that's working ok on our fork.

NewAgeAirbender:
Awesome! Got it working, something was odd with my setup. Looks good!

PR #4818, opened 2024-01-29
VT: event scraper associating same bills with all meetings of a committee

braykuka:
VT: event scraper associating same bills with all meetings of a committee [#1150](https://github.com/openstates/issues/issues/1150)

### DIFF
--- a/scrapers/vt/events.py
+++ b/scrapers/vt/events.py
@@ -1,51 +1,80 @@
+import datetime
+import re
 import dateutil.parser
 import json
 import pytz
 
-from dateutil.parser import ParserError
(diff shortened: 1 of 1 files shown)


braykuka:
@jessemortenson Please review it.

jessemortenson (approved):
Thanks for the work! Looks good - I did make a couple quick changes:

- Moved the timeframe for processing events into the past (not just datetime.today()), because I noticed that some agenda documents include dates that are a few days in the future, even though the JSON data `MeetingDate` is in the past
- Used a set to fix a bill being added to an event twice

PR #4820, opened 2024-01-30
HI: Add option to scrape subset of days

showerst:
I found a listing page on HI's site that includes all the bills with last update date.

This allows you to do a scrape with a 'days ago' argument, scrape_since=N and only scrape bills with updates in the last N days.

This drastically cuts down on the volume of proxied requests needed to keep up in HI. You should set it to at least two though, as they do sometimes sneak bills super late their time.

I'm open to renaming or better documenting that argument flag, just passing this for feedback.


NewAgeAirbender (approved):
Nice find! Makes sense to me

jessemortenson (approved):
Looks Good To Me as well, nice find!

PR #4831, opened 2024-02-09
WIP: KS: new RSS/API based scraper

showerst:
The old KS scraper would sometimes start dying at the end of session because the state's API listing endpoint would begin to time out.

This scraper is based on the RSS listing page (which seems to be fast), and then the individual bill API pages, which have enough info to scrape w/out loading the HTML pages and getting banned.

NOTE: This removes the votes scraper, whose data isn't available in the API. I'm hoping we can create a new one and get that data somehow without scraping the HTML pages.

### DIFF
--- a/scrapers/ks/bills.py
+++ b/scrapers/ks/bills.py
@@ -1,10 +1,9 @@
 import re
+import feedparser
 import json
-import datetime
-import requests
-
-import lxml.html
-from openstates.scrape import Scraper, Bill, VoteEvent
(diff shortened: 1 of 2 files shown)


NewAgeAirbender:
Ran this & looks good! I think we should merge it & I can assign the vote scraper out since we're missing bills since the current scraper keeps failing

showerst:
@NewAgeAirbender thanks for the reminder!

PR #4842, opened 2024-02-16
FL: Fix for errant vote lines

showerst:


showerst:
Waiting on a test scrape here.

PR #4870, opened 2024-03-06
WY Events: Strip leading zeros from bill number

ryan-ricard:
WY Events scraper was adding agenda items with name/identifier that looked like `SF0013`, these do not match bill numbers in this jurisdiction, which cause later pipeline steps matching events to bills to fail. The fix presented here is to separate out the numerical part of the "bill number" and strip leading zeros before adding the agenda item.

### DIFF
--- a/scrapers/wy/events.py
+++ b/scrapers/wy/events.py
@@ -3,6 +3,7 @@ from openstates.exceptions import EmptyScrape
 from dateutil import parser, relativedelta
 import datetime
 import pytz
+import re
 from utils.events import set_location_url, match_coordinates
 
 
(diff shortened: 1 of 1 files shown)


scrapers/wy/events.py
```diff
@@ -88,7 +89,12 @@ def scrape(self):
                         bills_agenda_item = event.add_agenda_item(
                             "Bills under Consideration"
                         )
-                    bills_agenda_item.add_bill(bill["billNumber"])
+                    # Separate "SF" from "0012" in "SF0012" to strip leading zeroes
+                    num_match = re.search('([A-Z]+)(\d+)',bill["billNumber"] )
```
NewAgeAirbender, line 93: Have to add an `r` before the regex string so it knows it's regex. Also space after comma in parameters & not before closing parentheses

scrapers/wy/events.py
```diff
@@ -88,7 +89,12 @@ def scrape(self):
...
+                    if (num_match):
```
NewAgeAirbender, line 94: Also don't need parentheses here, it'll work without them

NewAgeAirbender (commented):
Just some edits for the lint to pass but spot checks look good on the new bill format!

ryan-ricard:
@NewAgeAirbender Whew! Python is a bit rusty but I've got pre-commit installed now so I should be able to manage :)

NewAgeAirbender (approved):

PR #4894, opened 2024-03-26
MN: take house out of committee

NewAgeAirbender:


showerst:
@NewAgeAirbender -- pretty sure I added 'House' here because otherwise some committee names are ambiguous, like 'Taxes'. Is there a reason you removed it?

NewAgeAirbender:
It's still unmatching those committees, testing out some fixes for core so will be adding it back!

showerst:
No prob. 

FYI in jurisdictions where the committee is ambiguous I think we modded a few of the event scrapers to write <house|assembly> X or <senate> X. 

I _think_ the designed way to do this is to add it as an alternate name in the committee file in openstates/people but I have no clue how/if that stuff is hooked up in core.

NewAgeAirbender:
Yeah, there is an `other_name` that can get added to the committee file in `people`, but found out that's not actually included in the query to match the committee entity in `core` so hoping to have a deploy that does that soon.

PR #4904, opened 2024-04-01
MI: Bills: Fix for chapter law regex

showerst:


showerst:
Chapter law regex was blowing up on 3 digit chapter laws, like PA 103'23

https://legislature.mi.gov/Bills/Bill?ObjectName=2023-SB-0173

PR #4930, opened 2024-04-25
ND bills: new scraper from API

chrisyamas:
North Dakota bills scraper was rewritten to access bill data from [their new API](https://ndlegis.gov/legislative-api-documentation).
- new scraper extracts from JSON rather than from HTML, as before.
- still written using `spatula` library

### DIFF
--- a/scrapers/nd/bills.py
+++ b/scrapers/nd/bills.py
@@ -16,22 +16,18 @@ class BillList(JsonPage):
         super().__init__()
         self.input = input_data
         self.source = self.create_source_url()
-        # Initialize a logger for this class instance to avoid AttributeError
-        self.logger = logging.getLogger(__name__)
 
     def create_source_url(self):
         """
-        Dynamically retrieves session_year for proper url path to API endpoint.
+        Retrieves year for specified session.
+
+        Returns proper url path to API endpoint.
         """
-        assembly_session_id = self.input.get("session")
-        # Extract numeric-only part of the identifier if necessary
-        assembly_num = int("".join(filter(str.isdigit, str(assembly_session_id))))
-        # This formula reflects assembly number and session year relationship.
-        #  Ex. 67th Assembly (2021): assembly_num = 67 --> session_year = 2021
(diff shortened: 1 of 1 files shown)


scrapers/nd/bills.py
```diff
@@ -1,164 +1,125 @@
...
+from spatula import JsonPage
+from .actions import NDCategorizer
 
 
-class BillList(HtmlListPage):
+class BillList(JsonPage):
+    # TODO: remove this hard-coded {assembly-session_year} below, there must be
+    #  a way to get access to the "start_date" value for the active leg session
+    #  being scraped, so that the year could be pulled from that and this could
+    #  then be assembled in a variable way in the `NDBillScraper()` class.
```
chrisyamas, line 13: @NewAgeAirbender curious for your thoughts on how this could be achieved.

chrisyamas: I know that the active sessions' `identifier` value (from the dictionary in the `legislative_sessions` list in the `__init__` file) is accessible within the `NDBillScraper`, as it's getting passed to its `scrape()` method as the `session` argument.

I began looking more into OS-core to try to find an answer to how those other values in the active session's dictionary can be accessed. Didn't find answer as of last night, but I gotta imagine there are other scrapers that pass more than just the session identifier, right?

showerst: @chrisyamas it's kinda hacky, but to get that value you can do --

```python
        for i in self.jurisdiction.legislative_sessions:
            if i["identifier"] == session:
                start_date = i["start_date"]
```

showerst: Here's a few cases in the existing code where we do variations on that --

https://github.com/search?q=repo%3Aopenstates%2Fopenstates-scrapers%20self.jurisdiction&type=code

chrisyamas: Amazing, that's what I didn't realize, that I had access to `self.jurisdiction`. Thanks, @showerst! I'll incorporate that in my next commit.

PR #4950, opened 2024-05-10
VA: set special session live

NewAgeAirbender:
Session starts Monday but no bills file is on the site yet so holding off merging until that's live


NewAgeAirbender:
Update: the other file path doesn't have a csv for the new special session that started today but the sftp site does so added that back in.

PR #4954, opened 2024-05-19
FL: add headers for bill page request

chrisyamas:
Primary change is a modification of the url that is constructed from input in the `HouseSearchPage()` object.
- The HTTP request of the source did not have appropriate headers, and consequently was not getting the page data which would include the target bill page link, leading the `XPath()` selector to fail.
- This PR utilizes a `URL()` object to add the appropriate headers for the `GET` request to get the full page data necessary.

Secondary change adds the `media_type` for bill analysis documents, which was discovered to be missing while doing spot checks on the bill JSON data.


NewAgeAirbender (approved):
Looks good!

PR #4995, opened 2024-07-26
MA: new votes scraper

chrisyamas:
Senate votes scraper is written.

House votes scraper is somewhat built out, but most votes recorded in journals are in the Senate Journals, so that has taken priority to ensure we get most MA vote data in as soon as possible.


NewAgeAirbender:
Solid start! Vote basics are looking good, please take out the print statements

NewAgeAirbender (approved):

PR #4998, opened 2024-07-29
IL: Sponsorship Cleanup

braykuka:
## Description

- [#1213 Sponsorship Cleanup](https://github.com/openstates/issues/issues/1213)
- I've checked the current scraper and found that we don't need refine_sponsor_list function because the sponsor list in the source website is the last list after happened all actions. so the first sponsor is the Chief Sponsor and others are Co-sponsors. Please let me know if you have opinion for my thought.


braykuka:
@NewAgeAirbender Please review it for my last changes.

PR #5000, opened 2024-08-01
SD: Add room to location

braykuka:
- [issues 1214](https://github.com/openstates/issues/issues/1214)
- add room to the current location.
- switch the location to that, if API provides a building name

### DIFF
--- a/scrapers/sd/events.py
+++ b/scrapers/sd/events.py
@@ -1,8 +1,10 @@
-from openstates.scrape import Scraper, Event
+import re
 import pytz
 import lxml
 import scrapelib
 import dateutil.parser
+
+from openstates.scrape import Scraper, Event
 
 
 class SDEventScraper(Scraper):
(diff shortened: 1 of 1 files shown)


scrapers/sd/events.py
```diff
@@ -179,12 +182,16 @@ def scrape(self):
...
-    def create_event(self, committee, agenda_document):
+    def create_event(self, committee, agenda_document, location=""):
         name = committee["FullName"]
 
         start_date = dateutil.parser.parse(agenda_document["DocumentDate"])
 
-        location = "500 E Capitol Ave, Pierre, SD 57501"
+        base_loc = "500 E Capitol Ave, Pierre, SD 57501"
+        if location and re.match(r"Room \d+", location):
+            location = base_loc + ", " + location
```
NewAgeAirbender, line 192: Think this should be `location = location + ", " + base_loc` just for readability

NewAgeAirbender (changes requested):
Looks good besides one little change

scrapers/sd/events.py
braykuka: @NewAgeAirbender I've fixed this issue.

NewAgeAirbender (approved):

PR #5006, opened 2024-08-13
VA people: new session id, fix phone and address handling

chrisyamas:
This PR updates the Virginia people scraper to:
- get the new session data
- update handling for phone and address retrieval, as prior code was not getting capitol office phone numbers


NewAgeAirbender (approved):
Looks good!

PR #5015, opened 2024-08-27
VA committees refactoring, bypassing source errors

chrisyamas:
This PR refactors the Virginia committees scraper to bypass the frequent source errors that were occurring due to a confluence of how the page was structured and how spatula seeks sources.

Note: while the `CommitteeList()` class will get all the committees, the `FindSubcommittees()` class gets all the subcommittees. They are not linked in one action, and the job runner will need to execute those separately (i.e. include a `scrapers_next.va.subcommittees` execution for the ` FindSubcommittees()` class.


NewAgeAirbender (changes requested):
Looks good, just need to fix the lint issues before we merge

NewAgeAirbender (approved):

PR #5022, opened 2024-09-09
DC: update bill check on events

NewAgeAirbender:
They've updated how they post related bills on events. 

Heads up @showerst, I think it may also be worth moving over to the API for Events soon, looks like we're missing a bunch of them because the calendar on the site vs the file is different. Can access the same data with the API key with an event list & detail page.
`POST https://lims.dccouncil.gov/Hearings/API/Public/GetUpcomingHearings`
`GET https://lims.dccouncil.gov/Hearings/API/Public/GetHearingDetailsForPublic/503`


showerst:
Good flag, someone from govhawk or I will take this one tomorrow.

PR #5037, opened 2024-09-18
Skip specific link that return 404

alexobaseki:
This [link](http://webserver1.lsb.state.ok.us/2023-24HB/CMN-AGRI-20241207-01000000.pdf) fails with a 404. So we are going to skip it for now.


NewAgeAirbender (approved):

DAndrea56:
bitch get off my stuff

***@***.***

my account was compromised by an unauthorized user

PR #5048, opened 2024-10-08
USA: reduce duplicate 'Subcommittee Hearings Held' events

jessemortenson:
cc @showerst  in case you are doing event collation and this might impact you

The USA bill scraper has been yielding hearing events related to bills since April. However recently I have noticed a number of `DuplicateItemError` in our data processing pipeline, I believe because these events have sparse information and generic titles often, thus yielding more frequently detected duplicates.

My goal here is to reduce these duplicates, but also provide a more helpful title to the event.


showerst:
@jessemortenson thanks for the flag, makes sense to me. I'll check on the output of the events scraper later and make sure this doesn't break any detection there.

PR #5049, opened 2024-10-09
WA: fix bug where bill version notes are long repeated strings

jessemortenson:
We were getting string repeats concatenated into bill version notes field, like `Second Substitute Substitute Substitute Substitute Substitute Second Substitute Joint Resolution` because the loop didn't reliably initialize a variable. Some bill versions failed to import because the notes got so long, which is how I noticed.

A default of "Passed Legislature" is a little weird for passed bill versions ... but to fix we'd need to scrape each bill version page and hunt around for the right text. Which I don't think is really worth it but could be a good improvement in the future.

### DIFF
--- a/scrapers/wa/bills.py
+++ b/scrapers/wa/bills.py
@@ -124,7 +124,9 @@ class WABillScraper(Scraper, LXMLMixin):
 
                 bill_id = chamber[0] + bill_types[bill_type] + " " + bill_num
 
-                if bill_type != "Passed Legislature":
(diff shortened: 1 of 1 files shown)


showerst:
"Passed Legislature" comes from the state, they call it "Bill as Passed Legislature". e.g. https://app.leg.wa.gov/billsummary/?BillNumber=1017&Year=2023&Initiative=false so I think that's probably ok.

jessemortenson:
Good point, thanks Tim!

PR #5053, opened 2024-10-11
LA: add third extraordinary session slated for November

jessemortenson:
Don't actually have solid dates for this but it seems like it's happening:

- [Listed on website](https://www.legis.la.gov/legis/BillSearch.aspx?sid=243ES) (no bills yet)
- [Some reportage](https://www.usnews.com/news/us/articles/2024-10-01/louisiana-governor-plans-to-call-third-special-session-to-overhaul-the-states-tax-system)

### DIFF
--- a/scrapers/la/__init__.py
+++ b/scrapers/la/__init__.py
@@ -256,6 +256,15 @@ class Louisiana(State):
             "name": "2024 Second Extraordinary Session",
             "start_date": "2024-02-19",
             "end_date": "2024-03-06",
+            "active": False,
+        },
+        {
+            "_scraped_name": "2024 Third Extraordinary Session",
+            "classification": "special",
+            "identifier": "2024s3",
+            "name": "2024 Third Extraordinary Session",
+            "start_date": "2024-11-01",
+            "end_date": "2024-11-29",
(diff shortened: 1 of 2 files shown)


scrapers/la/__init__.py
```diff
@@ -256,6 +256,15 @@ class Louisiana(State):
...
+            "active": False,
+        },
+        {
+            "_scraped_name": "2024 Third Extraordinary Session",
+            "classification": "special",
+            "identifier": "2024s3",
+            "name": "2024 Third Extraordinary Session",
+            "start_date": "2024-11-01",
+            "end_date": "2024-11-29",
             "active": True,
```
showerst, line 268: @jessemortenson when there's no bills yet for a session, add it with "active" set to False. States occasionally drop test bills, broken bills, the wrong session's, etc in the first day of a session, so somebody should do a quick manual review before flipping to active so the scrapers don't pick up a bunch of junk that needs to be deleted.

jessemortenson: Cool, thanks for the reminder here - I'm a little rusty and always appreciate the feedback :)

PR #5054, opened 2024-10-13
New PA Events Scraper

austinhouck:
Updated the PA events scraper to use the new https://www.palegis.us/ website.

Addresses https://github.com/openstates/issues/issues/1239

### DIFF
--- a/scrapers/pa/events.py
+++ b/scrapers/pa/events.py
@@ -1,6 +1,5 @@
 import re
 import pytz
-import urllib
 import datetime
 
 import lxml.html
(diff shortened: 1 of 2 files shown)


austinhouck:
One sad thing is that the new classes for these tags are less descriptive, we used to be able to query on things like `CMS-MeetingDetail-Agenda-CommitteeName` but alas, this is no longer the case.

showerst:
@austinhouck thanks for the submission! We'll review this later today.

showerst:
Looks good!

PR #5056, opened 2024-10-18
MD & WY: add chamber to bill scrapers add_sponsorships()

alexobaseki:
In Maryland House bills can only be sponsored by House Members and the same applies to Senate bills. [Example - HB0901](https://mgaleg.maryland.gov/mgawebsite/Legislation/Details/hb0901?ys=2024RS).
* Adding chamber will help improve matching, especially in Maryland where several legislators share last name.


jessemortenson (approved):
LGTM

PR #5063, opened 2024-10-23
MO: Improve Senate sponsor psuedo_person_id

alexobaseki:
MO: House and Senate source website are quite different. House Sponsors mostly have full name as seen [here](https://documents.house.mo.gov/xml/241-HB2.xml) while senate can sometimes be just the last name [for example](https://www.senate.mo.gov/24info/BTS_Web/Bill.aspx?SessionType=R&BillID=11). So the change only apply to Senate Sponsor scrape.


jessemortenson (approved):
LGTM

PR #5064, opened 2024-10-24
TX: strip title from event participants name

alexobaseki:
Title in legislators name causes event participants mismatch for TX.


alexobaseki (commented):
Oops, mistakenly merged this.

PR #5067, opened 2024-10-28
ND: remove now-obsolete votes scraper

jessemortenson:
Improved vote logic is incorporated into the Bills scraper for ND. This separate votes scraper was only getting the most recent ~35 Vote Events, and it didn't fetch individual voters' votes (only aggregate counts). 

So I can't think of any reason to keep this one around. @showerst heads up in case you want to keep this (or need to disable in your setup)


showerst (commented):
LGTM, thanks for the head's up.

PR #5069, opened 2024-10-31
IN: fix event exhibits and minutes url

braykuka:
- add the exhibit and minute urls
- remove the proxy-based urls


braykuka:
@jessemortenson Please review it.

PR #5070, opened 2024-11-01
USA: Update action classification

alexobaseki:
There is a new action code(Intro-H) that is not documented above. Also adding (Intro-S) to mitigate against any surprise in Senate


jessemortenson (approved):
LGTM

PR #5078, opened 2024-11-12
TX: enable 2025 regular session 89

jessemortenson:
Also marks last two special sessions as inactive

### DIFF
--- a/scrapers/tx/__init__.py
+++ b/scrapers/tx/__init__.py
@@ -196,7 +196,7 @@ class Texas(State):
             "_scraped_name": "89(R) - 2025",
             "classification": "regular",
             "identifier": "89R",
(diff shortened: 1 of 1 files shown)


scrapers/tx/__init__.py
```diff
@@ -190,6 +190,15 @@ class Texas(State):
             "start_date": "2023-11-08",
             # TODO: correct end date
             "end_date": "2023-11-17",
+            "active": False,
+        },
+        {
+            "_scraped_name": "89(R) - 2025",
+            "classification": "regular",
+            "identifier": "89R",
+            "name": "88th Legislature (2025)",
```
showerst, line 199: Should be 89th, not 88th

showerst, line 199: Should be 89th.

PR #5089, opened 2024-11-19
VA Bills: bill identifier is None error

braykuka:
[1259](https://github.com/openstates/issues/issues/1259)


jessemortenson (approved):
LGTM

PR #5090, opened 2024-11-19
IN Events: scraper fails on parsing api on response

braykuka:
[1260](https://github.com/openstates/issues/issues/1260)

- API sometimes returns this error.
  'NoneType' object does not support the context manager protocol

### DIFF
--- a/scrapers/in/apiclient.py
+++ b/scrapers/in/apiclient.py
@@ -89,11 +89,16 @@ class ApiClient(object):
         headers["Accept"] = "application/json"
         headers["User-Agent"] = self.user_agent
         url = urljoin(self.root, f"/{session}")
-        resp = requests.get(url, headers=headers).json()
-        session_no_regex = re.search(r"Session\s+(\d+).+", resp["name"])
 
-        if session_no_regex:
-            session_no = session_no_regex.group(1)
+        while not session_no:
+            resp = requests.get(url, headers=headers)
+            resp = resp.json()
+            if "message" in resp:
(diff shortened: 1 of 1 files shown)


scrapers/in/apiclient.py
```diff
@@ -89,11 +89,16 @@ def get_session_no(self, session):
         headers["Accept"] = "application/json"
         headers["User-Agent"] = self.user_agent
         url = urljoin(self.root, f"/{session}")
-        resp = requests.get(url, headers=headers).json()
-        session_no_regex = re.search(r"Session\s+(\d+).+", resp["name"])
 
-        if session_no_regex:
-            session_no = session_no_regex.group(1)
+        while not session_no:
```
jessemortenson, line 93: I'm a little worried that we need some kind of backoff logic in here, so that we don't risk having a scraper that indefinitely repeats a request to the source website without any delay between requests. I think there's an example in `ga/util.py`.

We need to avoid creating a scenario where a scraper can hammer a source site with a (possibly) infinite loop of requests.

jessemortenson (changes requested):
Thanks for the explanation of how the API is being flaky. I'm requesting a change to how the case is handled

scrapers/in/apiclient.py
braykuka: @jessemortenson Thanks for your review. I added a backoff function as you mentioned. Please review it.

jessemortenson (approved):
LGTM, thanks!

PR #5120, opened 2024-12-06
IA: Add ability to scrape 2025 prefiles

showerst:
@jessemortenson -- tagging you because this one will require a config change.

IA posted prefiles, but they haven't added the session to the main logic yet, and if you try to hit the various normal bill listing pages with session 91 in the URL argument, it returns the session 90 bills. So I made this default to 2023-24, and used an extra argument to ONLY scrape the prefiles page.

To get the live prefiles, set your scraper up to --

```python
poetry run os-update ia bills --session=2025-2026 prefiles=True
```


jessemortenson:
Cool, thanks Tim!

PR #5126, opened 2024-12-11
WIP: Alabama gql updates

showerst:


showerst:
@jessemortenson just a flag on the state of AL -- The scraper is working with prefiles for the basic data. It's not currently doing votes, enacted versions, or code cites. Could use a code cleanup too.

I'll get to those as time allows if you don't beat me to it. There shouldn't be any big votes or enacted/code cites for at least a few weeks.

jessemortenson:
Gotcha, thanks Tim

PR #5136, opened 2024-12-16
TX: improve vote scraping

braykuka:
- add vote scrpaing code to bill scraper
- need to add the code for a voice vote scraping
- if we use the llm and openai for this scraper, then we can use pdf text and next pdf text.
- still more tweaks for this pull request


jessemortenson:
Thanks again, Bray. I found a way to improve a lot of the missing votes in the existing scraper. Gonna close this since that is now merged in.

PR #5146, opened 2024-12-19
Do not merge -- AR, debugging a decoding issue

showerst:


jessemortenson:
Seems like we can close now that we have https://github.com/openstates/openstates-scrapers/pull/5150

PR #5171, opened 2025-01-03
TX: votes: fix session year selection bug

jessemortenson:
@showerst well I broke the session naming convention in TX for this year (`89R` instead of `89`). Was that a dumb decision? I'd like to blame it on late night coding or something but the commit log says otherwise. 

Think I should revert the session identifier to `89` or do this fix (and any other likely following ones to adjust other assumptions about the session identifier string)?

### DIFF
--- a/scrapers/tx/votes.py
+++ b/scrapers/tx/votes.py
@@ -639,15 +639,11 @@ class TXVoteScraper(Scraper):
             yield vote
 
     def get_session_year(self, session):
-        if "R" in session:
-            session_num = session.strip("R")
-        else:
-            session_num = session
         session_instance = next(
             (
                 s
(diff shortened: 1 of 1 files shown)


showerst:
I'd say once we've got data propagated with your session ID don't change it, since that means all downstream consumers need to go change their IDs. So I'd fix the scraper without changing the ID.

FWIW personally I think any case where changing the id out of some unadvertised pattern 'breaks' the scraper is bad practice, and we should move those kind of IDs into the session metadata extras. 

You have to do a convoluted lookup to get at that session metadata right now, maybe once things cool down we can think of a better way to make that accessible to the scrapers.

jessemortenson:
Thanks for the feedback Tim, I'll merge this fix in and leave the new session ID as is.

PR #5191, opened 2025-01-08
MA: Fix invalid 194th session end date

nickrecchi:
Massachusetts: The 194th session end date had an invalid date of 2025-11-31. This was changed to the correct end date of 2026-07-31. Here's the source of that information:
https://malegislature.gov/ClerksOffice/Senate/Deadlines


jessemortenson:
Thanks for the correction!

PR #5213, opened 2025-01-13
PR bill scraper rewrite

braykuka:
[1283](https://github.com/openstates/issues/issues/1283)

### DIFF
--- a/scrapers/pr/__init__.py
+++ b/scrapers/pr/__init__.py
@@ -1,8 +1,10 @@
 import requests
-import lxml.html
 from openstates.scrape import State
 from .bills import PRBillScraper
 from .votes import PRVoteScraper
+import re
+import json
+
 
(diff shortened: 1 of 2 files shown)


jessemortenson:
Thanks Bray! I pushed a change because it seems like there are not actually any votes on the new site. The scraper was creating hundreds of votes and when I spot checked the documents, they all seemed to be bill versions (not votes).

If I'm wrong please link me to some examples of vote documents you found!

braykuka:
> Thanks Bray! I pushed a change because it seems like there are not actually any votes on the new site. The scraper was creating hundreds of votes and when I spot checked the documents, they all seemed to be bill versions (not votes).
> 
> If I'm wrong please link me to some examples of vote documents you found!

@jessemortenson , Thanks for your feedback.
You are correct. There are no vote section on site now.

PR #5242, opened 2025-01-21
MS Events update and improvement for 2025

braykuka:
[1290](https://github.com/openstates/issues/issues/1290)


jessemortenson:
Thanks Bray - I needed to fix to use the correct URL for House events

- https://www.legislature.ms.gov/calendars-and-schedules/house-calendar/ is the URL for the schedule of when bills will be heard/discussed (I believe)
- https://www.legislature.ms.gov/calendars-and-schedules/house-schedule/ is the schedule for committee meetings

PR #5243, opened 2025-01-22
IN State Committees

jojech:
[Issue](https://github.com/openstates/issues/issues/1292)

**Changes**
- Updating an outdated session_year variable
- Updating the url for the indiana state site to reflect the currrent session


jessemortenson (approved):
Thanks for the contribution, jojech!

PR #5247, opened 2025-01-23
CA events Senate site seems to need rewrite

braykuka:
[1291](https://github.com/openstates/issues/issues/1291)


jessemortenson:
Thanks, Bray. Looks like you missed a couple pieces from my WIP branch that I shared in the ticket. Also found a status case to handle, and added a missing call to `add_committee()`

PR #5261, opened 2025-01-30
VA event: Fix date parse error

alexobaseki:
Quick fix for current error in VA events. This doesn't fix the missing or/and invalid event bill IDs.

### DIFF
--- a/scrapers/va/events.py
+++ b/scrapers/va/events.py
@@ -163,7 +163,11 @@ class VaEventScraper(Scraper):
             if when_time == "Invalid date":
                 when_time = ""
 
-            when = dateutil.parser.parse(f"{when_date} {when_time}")
+            try:
+                when = dateutil.parser.parse(f"{when_date} {when_time}")
+            except ValueError:
+                # Handle cases where when_time is not a valid time.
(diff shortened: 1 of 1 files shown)


scrapers/va/events.py
```diff
@@ -163,7 +163,11 @@ def scrape(self, start_date=None):
             if when_time == "Invalid date":
                 when_time = ""
 
-            when = dateutil.parser.parse(f"{when_date} {when_time}")
+            try:
+                when = dateutil.parser.parse(f"{when_date} {when_time}")
+            except ValueError:
+                # Handle cases where when_time is not a valid time.
+                when = dateutil.parser.parse(when_date)
```
jessemortenson, line 170: I think we need one more change here: when an event is assigned a datetime that actually doesn't have a time input (it's just a day, and we don't know the time), our standard is to do two things:

1. mark that event as `all_day=True` 
2. make sure that the value of `start_time` is a `date` object, not a `datetime` object. So ie `start_time=when.date()`

example of this handling is here: https://github.com/openstates/openstates-scrapers/blob/fb8fa9aa2bb0aba590f3c7a53fa7247384b09105/scrapers/pa/events.py#L60

PR #5262, opened 2025-01-30
VA: event add all_day flag

alexobaseki:
It seems like the 2 cases with invalid `when_time` the `when_date` actually had a reasonable datetime object. `'2025-01-30T16:30:00'` So I decide to modify it like this.
Note: In line 160 above we were already doing this. `when_date = row["ScheduleDate"].replace("T12:00:00", "")`

Let me know if I should just stick to the default.


jessemortenson (approved):
LGTM

alexobaseki:
> LGTM

Made an update to be able grab house bill coreectly.

PR #5275, opened 2025-02-05
NH: Add a second source link for bills

alexobaseki:
Bill Status link doesn't work for current session. So I am adding the Bill Docket link.

### DIFF
--- a/scrapers/nh/bills.py
+++ b/scrapers/nh/bills.py
@@ -306,6 +306,12 @@ class NHBillScraper(Scraper):
 
     def add_source(self, bill, lsr, session):
         bill_url = (
+            "https://www.gencourt.state.nh.us/bill_status/legacy/bs2016/bill_docket.aspx?"
+            + "lsr={}&sy={}&sortoption=&txtsessionyear={}".format(lsr, session, session)
+        )
+        bill.add_source(bill_url)
+
+        bill_url = (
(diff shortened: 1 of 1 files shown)


jessemortenson:
If the status link does not work, should the code that adds the status link be commented out? I guess another way to put it: do you think the status link will start working again sometime this session?

FYI if there are two source URLs, there will need to be code in the NH adapter on the Plural side to select the correct one.

jessemortenson:
Oh - and I just remembered - NH seems to have a domain name change recently - https://gc.nh.gov/ - did you try the status link with that domain name? Might be worth checking both with it. Seems like NH is redirecting https://www.gencourt.state.nh.us to https://gc.nh.gov, so it would not surprise me if the www.gencourt.state.nh.us domain fails on certain redirects now (or in the future)

alexobaseki:
> If the status link does not work, should the code that adds the status link be commented out? I guess another way to put it: do you think the status link will start working again sometime this session?
> 
> FYI if there are two source URLs, there will need to be code in the NH adapter on the Plural side to select the correct one.

I felt I should keep the existing link since the bill status URL will work at some point either when they have it fixed or when this  session (2025) is archived.

jessemortenson (approved):
ok LGTM

alexobaseki:
> https://gc.nh.gov,

Yes. I did. I am still working on fixing the bad pdf link. I can update the url to the new one in the follow up PR too

PR #5282, opened 2025-02-10
Change next-action for request headers

alexobaseki:
It looks like this values will keep changing and causing failing


showerst:
If this keeps breaking every time they rebuild and that action ID changes, I don't think it's bad practice to just hard-code this list. They're only adding one session per 3 years, and they don't do specials, so i'd personally be fine with skipping the request entirely and just making the function.

```python
return ["2025-2028", "2021-2024", ..etc] 
```

alexobaseki:
I agree! I thought of doing that today.

PR #5298, opened 2025-02-19
WA: bills: fix scraping of bill actions from bill page

jessemortenson:
fyi @showerst looks like we've been missing quite a few WA bill actions as the scraping logic for the list on the bill details page stopped working. We continued to get a limited set of actions from the `scrape_hearings()` method.

But once the primary actions scrape logic was fixed, then the `scrape_hearings()` method starts adding duplicate bill actions. Spot checking a few bills, it seems `scrape_hearings()` is not helpful anymore. Leaving this up overnight in case this looks unwise in the morning.

### DIFF
--- a/scrapers/wa/bills.py
+++ b/scrapers/wa/bills.py
@@ -239,7 +239,7 @@ class WABillScraper(Scraper, LXMLMixin):
             self.scrape_chamber(chamber, session)
 
         # uncomment the line below to scrape a single bill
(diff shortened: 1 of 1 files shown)


jessemortenson:
Yes, I've also been dreaming of ways to incorporate more statistical/aggregate automated quality checks. Have looked at Great Expectations (a while ago) and SQLMesh audits (recently). But we haven't taken the plunge into a project yet.

PR #5299, opened 2025-02-19
VA: bills: better handle action actor re: missing ChamberCode value

jessemortenson:


showerst:
@jessemortenson chamber = None causes some pain for my system, does it matter to you at all if I change this to default to chamber = 'legislature' ? I'll submit the patch.

jessemortenson:
In fact, I'm looking at these edge cases now, I'll make that change myself

showerst:
@jessemortenson future note to us, maybe "None" chamber on a bill, action, or vote should throw a validation error. I started poking around our DB and the few cases where they weren't be either upper/lower/joint, legislature, or executive were junk.

PR #5309, opened 2025-02-24
xlmlize utilities should respect VERIFY_CERTS

showerst:
The shortcut methods for grabbing urls (self.lxmlize and url_xpath) weren't checking for a `VERIFY_CERTS` env var, so you couldn't globally set verify to false using the env var in the scrapers that used these methods. This should fix that that.

@jessemortenson courtesy ping, i'm changing the default function args for lxmlize and url_xpath from True to None, but then defaulting the `verify` keyword to `True` lower down in the code, so behavior shouldn't change.

### DIFF
--- a/scrapers/utils/lxmlize.py
+++ b/scrapers/utils/lxmlize.py
@@ -4,8 +4,12 @@ import logging
 import os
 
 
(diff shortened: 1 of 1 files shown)


jessemortenson:
Makes sense, thanks!

PR #5314, opened 2025-02-27
Update scraper docker entrypoint

alexobaseki:
Updating Docker Entrypoint here so scheduler will be able to run people-repo-to-database.  see core [PR](https://github.com/openstates/openstates-core/pull/161)


jessemortenson:
Is `COMMAND_MODE` a standard thing or like an env var that you would set for the container? I don't recognize it. Smells like an obscure special case, but maybe I'm not understanding the intent/context.

Do we need to modify the openstates-scrapers docker image's entrypoint to support this? We do already regularly build an [openstates-core docker image](https://hub.docker.com/r/openstates/core) - can we run the relevant command in that image? the openstates-core Docker image is not bundled with the helper docker_entrypoint.sh script currently, but do we need Google Cloud credentials for the people repo update script? 

I would guess the command would run OK via the `openstates-core` docker image, in which case we don't need to modify the scrapers docker image to support this

alexobaseki:
> Is `COMMAND_MODE` a standard thing or like an env var that you would set for the container? I don't recognize it. Smells like an obscure special case, but maybe I'm not understanding the intent/context.
> 
> Do we need to modify the openstates-scrapers docker image's entrypoint to support this? We do already regularly build an [openstates-core docker image](https://hub.docker.com/r/openstates/core) - can we run the relevant command in that image? the openstates-core Docker image is not bundled with the helper docker_entrypoint.sh script currently, but do we need Google Cloud credentials for the people repo update script?
> 
> I would guess the command would run OK via the `openstates-core` docker image, in which case we don't need to modify the scrapers docker image to support this

COMMAND_MODE is an env var that I am setting for the container. I am open to leaving the docker image entrypoint as is if there is a way to `poetry run os-update` without having to modify `openstates.cli.update`. Because my thought here was to have this run independent of `update.py`

PR #5322, opened 2025-03-05
chore: handle blockages from fl website

elseagle:
This PR enhances the Florida scraper to handle connection issues that occur when the state website blocks our scraping attempts. After about 15 minutes of continuous scraping, the FL website starts abruptly closing connections without responses, effectively blocking our data collection.

Key improvements:

- Added retry logic with exponential backoff to gracefully handle connection failures
- Implemented user agent rotation to make our requests appear to come from different browsers
- Added random delays between requests to mimic human browsing patterns
- Created a circuit breaker pattern that pauses scraping after consecutive failures
- Added periodic connection pool resets to establish fresh connections every 10 minutes
- Specifically targeted the "Remote end closed connection without response" error that FL uses to block scrapers

These changes should allow us to reliably scrape Florida's legislative data even though they've implemented measures to prevent automated access. Rather than trying to scrape everything at once (which triggers their blocking), we now use a more patient approach with built-in resilience against their defensive measures.

The scraper now behaves more like a human user browsing the site, making it less likely to trigger their anti-scraping mechanisms while still collecting all the data we need.

**Without this approach, I got nuked at ~8mins, with this approach, the scraper has been running for 5+ hours on my machine without any interruption and got completed**

Result summary below:
```
fl (scrape)
  bills: {}
bills scrape:
  duration:  4:59:38.252845
  objects:
    bill: 1856
    vote_event: 95
jurisdiction scrape:
  duration:  0:00:00.005200
  objects:
    jurisdiction: 1
    organization: 3
```

NOTE: This approach and code may not be so clean, but it's to ensure we get FL unblocked for now.

### DIFF
--- a/scrapers/fl/utils.py
+++ b/scrapers/fl/utils.py
@@ -1,6 +1,142 @@
+import logging
+import random
+import time
+from http.client import RemoteDisconnected
+from urllib.error import URLError
+
+from requests.exceptions import ConnectionError, Timeout, RequestException
+
+logger = logging.getLogger(__name__)
+
+
 def fix_name(name):
     # handles cases like Watson, Jr., Clovis
     if ", " not in name:
         return name
     last, first = name.rsplit(", ", 1)
     return first + " " + last
+
+
+def get_random_user_agent():
+    """
+    Return a random user agent to help avoid detection.
+    """
+    user_agents = [
+        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.124 Safari/537.36",
+        "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.1 Safari/605.1.15",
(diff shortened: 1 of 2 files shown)


elseagle:
Roger that, merging now. Thanks!

scrapers/fl/utils.py
```diff
@@ -1,6 +1,142 @@
...
+    Args:
+        min_seconds: Minimum delay in seconds
+        max_seconds: Maximum delay in seconds
+    """
+    delay = random.uniform(min_seconds, max_seconds)
+    logger.debug(f"Adding random delay of {delay:.2f} seconds")
+    time.sleep(delay)
+
+
+def retry_on_connection_error(func, max_retries=5, initial_backoff=2, max_backoff=60):
```
jessemortenson, line 49: (also in broader context re: scrapy: seems like this logic could fit neatly between scrapy retry settings https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#std-setting-RETRY_TIMES and a custom downloader middleware https://github.com/scrapy/scrapy/blob/master/docs/topics/downloader-middleware.rst)

PR #5327, opened 2025-03-07
Log external IP of each scraper run to assist with debug

jessemortenson:
In this age of more states adding anti-bot measures to their web services, it seems like it could be useful to just have the external IP address used by the scraper logged out on every run. Florida just responded by asking for an IP address re: their blocking behavior this week, for example. 

I added two checks here, one using `curl` and one using python `requests` since there might be a difference if `HTTP_PROXY`/`HTTPS_PROXY` is set. 

I don't 100% love using a third party service (https://ipecho.net) but also that's a lot easier than having infra for our own. I tried to double check bash recommendations for securely operating on unknown content.

Making you as a reviewer here @showerst in case this would introduce any wrinkle/concerns for you.

### DIFF
--- a/docker_entrypoint.sh
+++ b/docker_entrypoint.sh
@@ -21,5 +21,14 @@ elif [[ -n "${GOOGLE_CREDENTIAL_FILE}" ]]; then
     export GOOGLE_APPLICATION_CREDENTIALS
 fi
 
+###
+# Print external IP used: helpful for debugging connectivity/IP-based blocks
+# both OS-level and requests library (in case requests is using proxy)
+###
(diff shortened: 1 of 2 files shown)


showerst (approved):
Thanks for the flag -- we run our own fork with a custom dockerfile so you're good to go here. TBH I will likely merge this approach into our downstream too, it's not a bad idea. I don't have a better idea than hitting an external service either. 

If the free ones go away or there's privacy/rate limit concerns we can always stand one up real quick on a cheap vm, and convert the hardcoded URL to an ENV var.

jessemortenson:
Cool, thanks for the feedback Tim!

PR #5337, opened 2025-03-21
IA: bills: improve countermeasures vs HTTP disconnect

jessemortenson:
IA bill scraper still intermittently fails on HTTP disconnects. Seemed worth a hacky attempt to add in the countermeasures that Sogo implemented in FL bills here and see how that works.

Sogo - how incompetently did I adapt here? :)

If this improves in both cases, then probably a good candidate to refactor to a lower level in scraping.

### DIFF
--- a/scrapers/ia/bills.py
+++ b/scrapers/ia/bills.py
@@ -1,12 +1,29 @@
 import re
 import datetime
+from http.client import RemoteDisconnected
 import lxml.html
+import random
 import requests
 import time
 from openstates.scrape import Scraper, Bill
 from .actions import Categorizer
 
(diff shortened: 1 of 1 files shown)


elseagle (approved):
LGTM!

showerst:
@jessemortenson was brainstorming for if/when we try to make this code more universal, this looks kinda promising:

https://pypi.org/project/backoff/

Maybe with some combo of backoff, jitter, and custom event handlers that swap out the user agent.

jessemortenson:
Agree, that does look promising. I'm gonna merge this in and see if it reduces our IA fail/retries (at the whole-scrape-run level)

PR #5372, opened 2025-04-23
Bump Openstates version

alexobaseki:
With this bump in `openstates` version, scraper output will now have metadata `jurisdiction` and `scraped_at` keys like;
```
...
"jurisdiction": {"id": "ocd-jurisdiction/country:us/state:al/government", "name": "Alabama", "classification": "state", "division_id": "ocd-division/country:us/state:al"},
"scraped_at": "2025-04-14T19:59:48+00:00",
...
```
See related [version](https://github.com/openstates/openstates-core/pull/171).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1551,13 +1551,13 @@ et-xmlfile = "*"
 
 [[package]]
 name = "openstates"
-version = "6.21.6"
+version = "6.22.0"
 description = "core infrastructure for the openstates project"
 optional = false
 python-versions = "<4.0,>=3.9"
 files = [
-    {file = "openstates-6.21.6-py3-none-any.whl", hash = "sha256:bb7cca3fced222f6ef7c4af0edb6b8b5124a75014bc42b9b94c63217ef8759ad"},
-    {file = "openstates-6.21.6.tar.gz", hash = "sha256:23d6328fab6d0e626d2072df2a78a30362079801dc0c146bd6c0888c1d915350"},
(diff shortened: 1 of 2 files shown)


showerst:
Thanks for the head's up, the format looks good.

showerst (approved):

jessemortenson:
> I haven't dug into it, but how does the scraped_at play with your caching/idempotency? Are you removing that later before doing a cache check?

We actually run the os-update "import" logic (to get data into the Open States DB, which I *think* just updates every entity on every scrape (not efficient, but it's been that way a while). Then our Plural import stuff is downstream of that. I don't think adding these properties will cause the os-update "import" logic to trigger an improper last-update-at date to be stored, but it is worth a double check ...

jessemortenson:
@alexobaseki I was expecting an explicit version change in pyproject.toml here and I'm only seeing poetry.lock changed. Maybe ok via semantic versioning but my pref is to explicitly set the new version in pyproject.toml as well

showerst:
> We actually run the os-update "import" logic (to get data into the Open States DB), which I think just updates every entity on every scrape (not efficient, but it's been that way a while). Then our Plural import stuff is downstream of that. I don't think adding these properties will cause the os-update "import" logic to trigger an improper last-update-at date to be stored, but it is worth a double check ...

You don't have to check on my behalf, we're already modifying the object before the hash check, so we'll just take that key out. Because we're doing a fresh --scrape every time, each object gets a fresh UID that has to go. Shame, that's a nice feature, maybe i'll add it back between the hash check and emit.

At some point i'll take a look at the logic y'all have built into core for this and my patches and see if we can unify it or just switch to what's in core, I think we came to super similar code.

jessemortenson (approved):

PR #5381, opened 2025-05-02
MN: bills: clean up MN testing mode a bit

jessemortenson:


showerst:
@jessemortenson yeah but then you went back and actually documented the change, so you win, lol.

PR #5490, opened 2025-11-05
MD committees: make lists discoverable to os-scrape, tweak name normalization

michaelmulley:
See openstates/issues#1344

MD committee data is out-of-date on v3.openstates.org, but the commitee scraper appears to still work well.

One guess I have as to the cause is that running `os-scrape md committees` results in `ERROR spatula: found no list pages in module scrapers_next.md.committees`. This PR switches to uses spatula's `ListPage` in order to make the scraper classes discoverable, and the `os-scrape` command now appears to work for me.

### DIFF
--- a/scrapers_next/md/committees.py
+++ b/scrapers_next/md/committees.py
@@ -1,3 +1,5 @@
+import re
+
 from spatula import HtmlPage, HtmlListPage, XPath, SelectorError, URL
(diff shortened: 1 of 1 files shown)


michaelmulley:
Thanks for letting me know. I updated the commits with `black`.

PR #5524, opened 2025-12-05
NH: Scrape prefiles from web

showerst:
@jessemortenson h/u -- We got tired of waiting to see if NH will update their bulk data, and found an RSS feed with the new data. I opted to leave the main scraper intact hoping to switch back later, and added a `scrape_from_web` flag to the scrape, to pull the new stuff.

This whole scraper could use some refactoring some rainy day, but for now it's getting all the new bills.

### DIFF
--- a/scrapers/nh/bills.py
+++ b/scrapers/nh/bills.py
@@ -50,8 +50,8 @@ extended_type_map = {
     "PET": "petition,",
 }
 
-VERSION_URL = "https://www.gencourt.state.nh.us/legislation/%s/%s.html"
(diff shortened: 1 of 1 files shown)


jessemortenson:
Thanks, Tim! Looks like the scraper is emitting duplicates, are you just specifying a single chamber as well as this flag? seems like running the scrape_from_web func twice doesn't add anything

showerst:
@jessemortenson that's probably an oversight, i'll take a look tomorrow or feel free to change sooner if you get to it.

PR #5655, opened 2026-05-04
VA: align SESSION_SITE_IDS with declared sessions

Shoyayeb:
scrapers/va/__init__.py declares the 2025, 2026, and 2026 Special I sessions with extras.session_code values 20251, 20261, and 20262 respectively, but scrapers/va/common.py's SESSION_SITE_IDS map only contains "2025": "243" — which is also the wrong code (243 is what 2025 used during pre-launch; the live LIS site uses the new 5-digit session codes that match extras.session_code).

The first time the scraper runs against any of these sessions, csv_bills.py crashes with KeyError before fetching any bills:

    File "scrapers/va/csv_bills.py", line 209, in scrape
        session_id = SESSION_SITE_IDS[session]
    KeyError: '2026S1'

Fix: update "2025" to "20251" and add the missing "2026" and "2026S1" entries so the map matches the active session declarations.


showerst:
Thanks!

PR #5656, opened 2026-05-04
KS: fall back to active legislative_sessions if live nav lookup fails

Shoyayeb:
Kansas.get_session_list() scrapes kslegislature.org's nav for the "Senate Bills" anchor and parses the session slug out of its href. The kslegislature.org nav has been rewritten and that anchor is no longer where the xpath looks, so url_xpath returns an empty list and the unconditional [0] crashes with IndexError on the session-list sanity check before any bill is fetched.

Observed in production:

    File "scrapers/ks/__init__.py", line 110, in get_session_list
        url = url_xpath(
              ^^^^^^^^^^
    IndexError: list index out of range

Fix: wrap the live lookup in try/except and fall back to the _scraped_name(s) of whatever's marked active in legislative_sessions so openstates can still validate the session and proceed with the scrape. Behaviour when the nav lookup succeeds is unchanged.

### DIFF
--- a/scrapers/ks/__init__.py
+++ b/scrapers/ks/__init__.py
@@ -107,8 +107,22 @@ class Kansas(State):
     ignored_scraped_sessions = []
 
     def get_session_list(self):
-        url = url_xpath(
-            "https://www.kslegislature.org/li",
-            '//div[@id="nav"]//a[contains(text(), "Senate Bills")]/@href',
-        )[0]
-        return [url.split("/")[2]]
+        # The kslegislature.org nav has been rewritten and the "Senate Bills"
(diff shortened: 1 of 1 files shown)


showerst:
Hi, we don't want the logic to work like this for two reasons:

1. We use that get_session_list as a signal that the site markup hasn't changed and that the site is still up, if the markup has changed there it has likely changed elsewhere
2. If they added a new session and we do a big scrape with all the bills assigned the incorrect fallback session, that's a big mess to undo, and could result in sending out incorrect alerts to end users.

The correct fix here is to update the markup parsing code, which we'd be happy to take a PR for.

PR #5662, opened 2026-05-05
TN: fix for special index pages

showerst:
@alexobaseki ping so we don't step on each other's toes. I moved the special session index page logic to a new function to clean things up a bit, this will correctly scrape both specials.


alexobaseki (commented):
Looks good to me, thanks!

PR #5669, opened 2026-05-13
WI: add missing special session slug

showerst:
@alexobaseki head's up WI will pull all the regular bills into the special unless this is added too.


alexobaseki:
Thank you @showerst

PR #5670, opened 2026-05-13
DE: drop per_page down to 20

showerst:
I can't get a DE scrape to run unless I drop this down to 20.

### DIFF
--- a/scrapers/de/bills.py
+++ b/scrapers/de/bills.py
@@ -62,7 +62,7 @@ class DEBillScraper(Scraper, LXMLMixin):
         # Cannot scrape a particular chamber, since the search-by-chamber
         # returns bills that are currently _active_ in a chamber,
         # and is not a search by chamber-of-origin
-        per_page = 200
+        per_page = 20
 
         bills_and_votes = []
         page_number = 1


showerst:
@alexobaseki -- when you have a chance, can you try running this with per_page set at 200 vs 20 to confirm it's not just something weird about my setup?

If you can confirm that 200 fails and 20 is right i'll merge.

alexobaseki:
> @alexobaseki -- when you have a chance, can you try running this with per_page set at 200 vs 20 to confirm it's not just something weird about my setup?
> 
> If you can confirm that 200 fails and 20 is right i'll merge.

You are right this fails with a ```raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)``` error when I use the 200 page size.

alexobaseki:
> > @alexobaseki -- when you have a chance, can you try running this with per_page set at 200 vs 20 to confirm it's not just something weird about my setup?
> > If you can confirm that 200 fails and 20 is right i'll merge.
> 
> You are right this fails with a `raise RequestsJSONDecodeError(e.msg, e.doc, e.pos) requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` error when I use the 200 page size.

Just to confirm. 20 works right. So good to merge.

PR #5673, opened 2026-05-17
Drop redundant (?i) prefixes from utils.actions.Rule patterns

c-tonneslan:
Follow-up to #5672. Once that lands and `Rule` defaults to `re.IGNORECASE`, the inline `(?i)` prefixes scattered across state action categorizers become no-ops. Strip them out so the regexes are a little easier to read and the convention stays consistent.

Only touched files that import `Rule` from `utils.actions` (`ca`, `co`, `de`, `ma`, `me`, `ok`, `wa`, `wv`). `ny`, `pa`, and `tn` have their own local categorizer implementations that still rely on `(?i)`, so they're untouched.

Marking as a follow-up: please don't merge until #5672 is in.


jessemortenson:
Thanks for the contributions!

PR #5687, opened 2026-06-04
DATA-5684: Handle missing effective date in CO bill session laws

harshitasingh360:
This PR fixes a crash in the Colorado bill scraper session law parsing by handling rows where the session law row has an empty or missing effective-date cell in td[1]/span. It updates clean() so empty XPath results return an empty string and changes scrape_laws() to read effective_text first, parse it only when non-empty, and set effective=None when the date is missing, while still adding the citation with chapter, title, and URL. The fix was verified by reviewing the HTML parsing logic and confirming the updated file compiles cleanly with no parser call on empty date strings, making Colorado session law parsing more robust against optional or incomplete row data.

### DIFF
--- a/scrapers/co/bills.py
+++ b/scrapers/co/bills.py
@@ -90,7 +90,9 @@ class COBillScraper(Scraper, LXMLMixin):
         self.page_number += 1
 
     def clean(self, text):
-        if type(text) is list:
+        if isinstance(text, list):
+            if not text:
+                return ""
             return text[0].text_content().strip()
         return text.text_content().strip()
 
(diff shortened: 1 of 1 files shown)


harshitasingh360:
@jessemortenson Please review this.

scrapers/co/bills.py
```diff
@@ -90,7 +90,9 @@ def scrape_bill_list(
         self.page_number += 1
 
     def clean(self, text):
-        if type(text) is list:
+        if isinstance(text, list):
```
jessemortenson, line 93: I think this proposed change to `clean()` could be simplified a bit. The original function assumes the `text` variable is either an lxml DOM Element , or a list of lxml DOM Elements. If the value of `text` is None (or an empty list), then either `return` statement will fail (both are calling lxml instance method `text_content()` on the var).

I think we may also want the function to explicitly return None, so that any calling code has to handle that (whereas there might be a silent failure if the calling code gets an empty string and just assumes it is not empty).

So I think this would be clearer:

```
def clean(self, text_element):
  if not text_element:
    # if text_element is an empty list, empty string or None, signal that we could not clean it
    return None
  if type(text_element) is list:
    return text_element[0].text_content().strip()
  return text_element.text_content().strip()
```

What do you think - am I missing an aspect that you saw?

jessemortenson (changes requested):
Thanks for the fix! I have a suggestion for refactoring the `clean()` function a bit. My biggest concerns there were having the unexpected-input-handling (`if text is None: return ""`) nested in the function, and the possibility of returning an empty string and client code just assuming it's a "good" string.

jessemortenson:
Closed as Tim fixed underlying issue in another PR

PR #5688, opened 2026-06-05
DE: experiment with skipping bills when core fetch fails

jessemortenson:
Continuing to wrestle with this one. Trying to create a pathway here where the scraper can totally fail on a bill and just skip it. I think this should be temporary, or else needs to be refined further (I'm just catching any exception). But still seeing a lot of failures caused by just intermittent retry fails, so trying to do something to get the full scrape to run mostly-consistently.

### DIFF
--- a/scrapers/de/bills.py
+++ b/scrapers/de/bills.py
@@ -12,6 +12,12 @@ from utils import LXMLMixin
 from .actions import Categorizer
 
 
+class FailedBillFetch:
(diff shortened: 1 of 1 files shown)


showerst:
FWIW i think this is a good approach, the usual strategy of 'fail the whole scrape' doesnt' make sense with DE and its goofy intermittent failures.

PR #5705, opened 2026-06-25
Updated KS Scraper for website change

harshitasingh360:


jessemortenson:
Thanks, @harshitasingh360  - I'm about to do a more detailed review, but two quick pieces of feedback first:

1. Since there are two separate scrapers here, I'd prefer they be split into two separate PRs. I know it's all under the same ticket, so you made a reasonable assumption. Not a big deal, but a general guideline going forward.
2. Looks like you have a linting issue in CI, above. Also not a big deal, catches me often. Please check for that after you make a PR (this repo runs both the Black formatter and the flake8 formatter, so two chances to get linting issues). You can run the same commands locally if you want to be proactive before pushing the PR  `poetry run flake8 scrapers scrapers_next` and `poetry run black --check --diff scrapers scrapers_next` (change to `poetry run black scrapers scrapers_next` to actually reformat). You can also integrate Black with PyCharm to either automatically re-format, or re-format on the "format code" command, if you want, under settings -> Python -> Tools -> Black

harshitasingh360:
Thanks @jessemortenson for pointing that out. I will have a look into it.

scrapers/ks/events.py
```diff
@@ -1,6 +1,9 @@
 import datetime
```
jessemortenson, line 1: Looks like `datetime` library is no longer used in this file, so import should be removed

scrapers/ks/events.py
```diff
@@ -15,76 +18,191 @@ def scrape(self, session, start=None):
...
+
+        landing_page = self.get(url_root).content
+        landing_page = lxml.html.fromstring(landing_page)
+
+        available_dates = landing_page.xpath(
+            "//div[@id='hearings-cal-root']/@data-available"
+        )
+
+        if not available_dates:
+            self.warning("No available hearing dates found")
```
jessemortenson, line 52: This `self.warning()` is fine, because it's warning about a possible problem.

scrapers/ks/events.py
```diff
@@ -15,76 +18,191 @@ def scrape(self, session, start=None):
...
-                event_name = f"{chamber}#{com_name}#{title}#{when}"[:500]
-                if event_name in events:
-                    self.warning(f"Skipping duplicate event {event_name}")
-                    continue
-                events.add(event_name)
-                event = Event(
-                    start_date=when,
-                    name=f"{chamber} {com_name}",
-                    location_name=location,
+        self.warning(f"Using hearings URL: {url_root}")
```
jessemortenson: It's OK to log some contextual information about the scrape, but please change to`self.info()` for any log statements that are not actually warning about a possible problem.

scrapers/ks/events.py
```diff
@@ -15,76 +18,191 @@ def scrape(self, session, start=None):
...
+                    f"&date={date_str}"
+                    f"&sort=date"
+                    f"&dir=desc"
+                    f"&per_page=20"
                 )
-                event.dedupe_key = event_name[:499]
-                event.add_participant(
-                    f"{chamber} {com_name}", type="committee", note="host"
+
+                self.warning(f"Fetching {url}")
```
jessemortenson: This log entry shouldn't be needed at all, because the underlying framework generally logs out URLs that are requested (but otherwise this would be another good example of something that should not be a `warning`)

scrapers/ks/events.py
jessemortenson, line 17: Looks like the old implementation took an incoming optional parameter, `start`. Looks like your new implementation does not use this. 

Event scrapers often scrape future events by default, or at least recent events + future events. (Our system is generally not interested in old past events, though they can be useful in certain situations). The old implementation set `start` to a default value of 30 days ago.

I recommend following the spirit of that prior implementation, by filtering days in `available_dates` to only dates that are greater than or equal to `start`. A similar default value of "now - 30 days" is a good policy: that way the scraper would continue to function as the prior implementation.

jessemortenson (changes requested):
Here is the review on the events scraper code. I wanted to share this before I go on to provide comments on the votes scraper, so you're not waiting for all the feedback at once. Overall the events scraper is looking good: it follows a reasonable logic to find events data and handle the grouping issue. I do request changes in a few areas.

harshitasingh360:
@jessemortenson I have noted all the pointers and would be rewriting the scraper accordingly. thank you for the feedback

scrapers/ks/votes.py
```diff
@@ -1,152 +1,226 @@
...
+
+        while True:
+            # Vote listings are loaded via the HTMX fragment endpoint.
+            # Paginate until no vote rows are returned.
+            list_url = (
+                f"{self.BASE_URL}/{biennium}/votes/fragment/"
+                f"?page={page}&per_page=20"
+            )
+
+            self.info(f"Get {list_url}")
```
jessemortenson: This is another log message that is redundant, as the scrape framework will log out URLs as they are fetched (at least when standard methods like `self.get()` are used)

scrapers/ks/votes.py
```diff
@@ -1,152 +1,226 @@
...
-        # Server sometimes sends proper error headers,
-        # sometimes not
-        try:
-            self.info("Get {}".format(link))
-            response = requests.get(link)
-            text = response.text
-        except requests.exceptions.HTTPError as err:
-            self.warning("{} fetching vote {}, skipping".format(err, link))
-            return
+        self.info(f"Fetching vote {link}")
```
jessemortenson: Also a redundant log message

jessemortenson (changes requested):
Events scraper now looks good 👍 

Votes scraper needs at least a couple tiny changes (log messages), and still needs resolution to the error where motion_text is empty (I believe https://www.kslegislature.gov/b2025_26/vote/?apn=b2025_26/year2/house/legislative_days/day018_20260205/vote_records/0038_SB254.odt is an example of that).

jessemortenson:
Looks good! Merged

PR #5706, opened 2026-06-26
fix: DC scraper crashes mid-run on non-PDF attachments, dropping bills

tamara-builds:
Fix: use get_media_type() for non-PDF attachments in DC bill details

When a DC bill's action attachment is not a PDF, mimetype is set to None and passed to add_version_link(). This causes OCD schema validation to fail with None is not of type 'string' for media_type, crashing the scraper mid-run. All bills after the failing bill in the API response are dropped from the output.

The legislationHistory attachment block above already uses get_media_type() for non-PDF files — this fix makes the leg_details["actions"] block consistent with it.

### DIFF
--- a/scrapers/dc/bills.py
+++ b/scrapers/dc/bills.py
@@ -49,7 +49,6 @@ class DCBillScraper(Scraper):
             leg_listing = resp.json()
 
             for leg in leg_listing:
-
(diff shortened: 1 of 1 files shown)


showerst:
Thanks!

PR #5714, opened 2026-07-01
Data 5696/wv events scraper missing interim events

harshitasingh360:


scrapers/wv/events.py
```diff
@@ -11,18 +11,23 @@ class WVEventScraper(Scraper, LXMLMixin):
     verify = False
     _tz = pytz.timezone("US/Eastern")
 
+    interims_url = "http://www.wvlegislature.gov/committees/Interims/interims.cfm"
```
jessemortenson: Since this a constant value, please instead define it outside of the class instance (ie up on line 9 or so) following the [python convention](https://peps.python.org/pep-0008/#constants): something like `INTERIMS_URL = "http://www.wvlegislature.gov/committees/Interims/interims.cfm"`

scrapers/wv/events.py
```diff
@@ -42,6 +47,172 @@ def scrape_committees(self, chamber, url):
                 event.dedupe_key = event_name
                 yield event
 
+    def scrape_interim_schedules(self, url):
```
jessemortenson: Since `url` value that is used here is a constant value (client code calls `self.scrape_interim_schedules(self.interims_url)`), the `url` parameter here does not really add value: I struggle to think of a near future where the function would be called again with a different `url` value passed in. At the same time, passing `url` as a parameter to this function does add opacity/indirection: it makes it a touch harder to figure out where the `url` value is coming from. And we generally only want to add opacity/indirection when doing so gives use some real value in exchange (modularity, re-usability, composibility).

So, please remove the `url` parameter from the function definition, and instead just have a first line in the function that sets `url` from the suggested `INTERIMS_URL` constant

scrapers/wv/events.py
```diff
@@ -42,6 +47,172 @@ def scrape_committees(self, chamber, url):
...
+                ]
+                if event_name in event_objects:
+                    self.warning(f"Found duplicate {event_name}. Skipping.")
+                    continue
+                event_objects.add(event_name)
+                event.dedupe_key = event_name
+                yield event
+
+    def scrape_interim_schedule_page(self, url):
+        self.info(f"GET {url}")
```
jessemortenson: Please remove redundant log entry, as framework will log requested URLs

scrapers/wv/events.py
```diff
@@ -42,6 +47,172 @@ def scrape_committees(self, chamber, url):
...
+        cells = row.xpath("./td")
+        if len(cells) < 4:
+            return
+
+        convene = cells[0].text_content().strip()
+        adjourn = cells[1].text_content().strip()
+        where = cells[3].text_content().strip()
+
+        # The committee cell holds either a linked committee name plus a
+        # separate "- Agenda" link, or plain text for site tours/presentations
```
jessemortenson: Looking at these interim pages more closely, I *think* the right approach is probably to ignore the events that are not linked. Things like "Dolly sods" or "Floor Session" (rows that do not have a link) mostly do *not* look like actual committee hearings. 

Starting at a more inclusive look (leaning towards getting more than strictly necessary) is a good approach, generally. In this case I think we should narrow a bit to just the events that have the link/agenda format.

(Otherwise the data produced by the scraper looks good!)

scrapers/wv/events.py
```diff
@@ -42,6 +47,172 @@ def scrape_committees(self, chamber, url):
...
+            event.add_committee(com, note="host")
+            event.add_source(com_link[0].get("href"))
+
+        if agenda_link:
+            self.scrape_interim_agenda_page(event, agenda_link[0])
+
+        yield event
+
+    def scrape_interim_agenda_page(self, event, url):
+        self.info(f"GET {url}")
```
jessemortenson: redundant log

jessemortenson (changes requested):
Thanks for the work here, it is in pretty good shape. A couple significant changes requested, please.

jessemortenson:
Thanks for the code changes. Looking at the code, I believe you have resolved the issues I raised. I'm going to go ahead and merge this in. 

You'll need to check the other PR on WV events (https://github.com/openstates/openstates-scrapers/pull/5727/) to see if there are any merge conflicts that need to be resolved there (once I merge this one in).

PR #5724, opened 2026-07-03
FL: stream bills as yielded, handle flhouse.gov bot detection gracefully

tamara-builds:
## Problem

**The FL scraper produces less than 30 bills despite over 5 hours of active scraping.**

This is running on a self-hosted home-network runner (not Azure/cloud IPs), so IP blocking is not the issue. flhouse.gov uses application-layer bot detection — HTTP 200 responses with a "Request Rejected" HTML body after sustained scraping. Spatula's accept_response() correctly identifies these. The problem is the list() wrapper in scrape(), which prevents any bills from being written to disk until the entire session generator completes. Since the session never completes (bot detection hits at ~bill 160), only the 5-bill special session is saved — the regular session produces zero bills across all 3 retry attempts.

The scraper wraps its session generator in `list()`:

```python
yield from retry_on_connection_error(
    lambda: list(do_scrape_with_retry()),  # forces full materialization
    max_retries=3,
    ...
)
```

`list()` forces the **entire session** (~8,000 bills for FL 2026) to fully materialize before a single bill is yielded to OpenStates and written to disk. When `RejectedResponse` is raised at bill ~160, `list()` propagates the exception and **zero bills are saved** despite all the work done before it. The scraper retries 3 times with the same result.

Evidence from a self-hosted home-network runner:
```
spatula.pages.RejectedResponse: Response was rejected (4x) by accept_response: <Response [200]>
⚠️ scrape attempt 3 failed; sleeping 20s...
Found 22 JSON files in _working/_data/fl
```
(22 files = 5 bills from the special session + org/jurisdiction metadata — zero regular session bills)

## Fix

**1. Remove `list()` in `scrape()`** — yield directly from the generator so bills are written to disk as they complete:

```python
# before
yield from retry_on_connection_error(
    lambda: list(do_scrape_with_retry()),
    max_retries=3,
    initial_backoff=10,
    max_backoff=120,
)

# after
yield from do_scrape_with_retry()
```

**2. Catch `RejectedResponse` in `_process_bill_list()`** — wrap the `for` loop so the scraper exits cleanly with partial data instead of crashing:

```python
try:
    for item in bill_list.do_scrape():
        try:
            ...
            yield item
            ...
        except Exception as e:
            ...  # existing per-item error handling
except Exception as e:
    # flhouse.gov application-layer bot detection raises a spatula rejection error
    if "reject" in str(type(e).__name__).lower() or "reject" in str(e).lower():
        self.logger.warning(
            f"flhouse.gov bot detection triggered — stopping session early. "
            f"Bills processed before this point have been saved. ({e})"
        )
    else:
        raise
```

## Result

- Bills are saved to disk as they're yielded — a partial scrape produces partial output rather than zero output
- The scraper exits cleanly when bot detection triggers, allowing downstream tooling to use what was captured
- `BillDetail.process_page()` already yields `self.input` (the bill) before `HouseSearchPage`, so bills with senate data are saved even if the house vote fetch never completes

## Notes

- `flsenate.gov` does not block — only `flhouse.gov` triggers bot detection
- The bot detection appears to be rate/pattern-based, not IP-based; home-network IPs are affected after sustained scraping of `SessionId=113` (2026 regular session)
- A separate discussion about checkpoint/resume support for scrapers that regularly hit mid-session rate limits is being tracked in openstates/issues

### DIFF
--- a/scrapers/fl/bills.py
+++ b/scrapers/fl/bills.py
@@ -730,6 +730,12 @@ class HouseSearchPage(HtmlListPage):
             },
             retries=3,
             verify=False,
+            # spatula's URL defaults to timeout=None (wait forever). flhouse.gov
+            # occasionally accepts the connection but never finishes the response,
+            # which without an explicit timeout hangs the whole scrape indefinitely
+            # instead of raising the ConnectionError/Timeout our retry logic already
+            # handles (see patched_get_response above).
+            timeout=30,
         )
 
     def accept_response(self, response: requests.Response):
@@ -759,7 +765,7 @@ class HouseSearchPage(HtmlListPage):
             return True
 
     def process_item(self, item):
-        source = URL(f"{item}", verify=False)
+        source = URL(f"{item}", verify=False, timeout=30)
         return HouseBillPage(self.input, source=source)
 
     # Override so that we can handle occasional bill that does not show up in search
@@ -787,7 +793,7 @@ class HouseBillPage(HtmlListPage):
     )
 
     def process_item(self, item):
-        source = URL(f"{item}", verify=False)
+        source = URL(f"{item}", verify=False, timeout=30)
         return HouseComVote(self.input, source=source)
 
 
@@ -886,7 +892,15 @@ class FlBillScraper(Scraper):
             "scrapers/fl/__init__.py"
         )
 
-    def scrape(self, session=None):
+    def scrape(self, session=None, allow_partial=None):
+        # Off by default: a scrape that hits flhouse.gov bot detection mid-session
+        # fails loudly rather than quietly returning a subset of bills, matching
+        # every other scraper's assumption that a successful run means a complete
+        # dataset. Pass allow_partial=true on the command line (e.g.
+        # `fl bills --scrape allow_partial=true`, same style as VA's
+        # `csv_bills --scrape session=2026`) to opt into keeping whatever bills
+        # were saved before the block instead of failing the whole run.
+        self.allow_partial_scrape = str(allow_partial).lower() in ("1", "true", "yes")
         self.raise_errors = False
         self.retry_attempts = 5
         self.retry_wait_seconds = 5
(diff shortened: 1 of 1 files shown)


jessemortenson:
I'm not going to merge this in as it stands. Our scrapers (and related systems) generally operate on the assumption that a complete run of the scraper yields a complete set of data (as of that time) about the legislative source. If the scraper exits successfully with only a subset of the data, then downstream users might consistently fail to get all data, without seeing the failure/error to alert them to that fact.

I ran this code and got ~1400 bills output. Successful scrapes of the status quo code yield ~1900. 

Now, the behavior about yielding scraped objects/files as the scraper progresses IS desired. There are some scrapers that yield all objects/files at the end, but most do so incrementally as they go. But this is a separate (though related) behavior from the question of a scraper completing successfully (not failing even though it actually failed to get all data).

If you want to refactor to put this behavior behind a scraper parameter that makes it optional, I'm open to that. By default the scraper should attempt to get all relevant data from the legislative source, or fail so that the user knows it failed to do so.

scrapers/fl/bills.py
```diff
@@ -480,9 +487,53 @@ def process_votes(self):
...
+    page, move on" handling spatula already applies to rejected responses, just for
+    transient network errors instead.
+
+    Deliberately NOT applied to BillList/BillDetail/SubjectPDF -- those are session-wide
+    or a bill's own core data, where silently swallowing a failure risks looking like a
+    complete, successful run when it isn't. This is only for data whose absence just
+    means "this one bill is missing some supplementary info," not "this bill or the
+    whole session is broken."
+
+    This scrape runs nightly, so a bill skipped tonight for a transient network blip
```
jessemortenson, line 532: Please remove comments re: scheduling frequency, as that depends on each downstream user's operations. (can give people the wrong expectations)

scrapers/fl/bills.py
```diff
@@ -731,6 +785,17 @@ def get_source_from_input(self):
...
+            # spatula's URL defaults to timeout=None (wait forever). flhouse.gov
+            # occasionally accepts the connection but never finishes the response,
+            # which without an explicit timeout hangs the whole scrape indefinitely
+            # instead of raising the ConnectionError/Timeout our retry logic already
+            # handles (see patched_get_response above). 10s applies separately to
+            # both the connect and read phases; every real request we observed in
+            # testing completed in 1-4s, so this is well clear of normal latency
+            # while still failing fast -- this fetch happens up to 3x per bill
+            # across ~1,900 bills, so a longer timeout compounds fast if a hang
+            # recurs. Tune freely if a different value fits better.
```
jessemortenson, line 849: This comment is pretty verbose and includes LLM text that is intended for the LLM user ("tune freely") - such comemnt content should not be in a final PR. Noting that this is called often and increasing the timeout value has ramifications is fine, though.

jessemortenson (commented):
Thanks for the update. I am requesting a couple changes here re: comments. Overall this direction looks good to me. I'm doing my own run of the scraper to give it further look.

PR #5725, opened 2026-07-06
VA Bill Actions and Statuses Missing fix

harshitasingh360:
Root cause: The VA scraper crashed with a KeyError on governor/executive actions because they use a space-prefixed format (" Vetoed by Governor", " Approved by Governor") instead of a chamber-letter prefix. This halted action ingestion, freezing bills at "passed."
- Scraper runs cleanly (3,646 bills, 38,049 votes — previously crashed)
- SB542 → executive-veto, SB518 → executive-signature, HB229 → executive-signature

### DIFF
--- a/scrapers/va/csv_bills.py
+++ b/scrapers/va/csv_bills.py
@@ -347,9 +347,7 @@ class VaCSVBillScraper(Scraper):
                 date = dateutil.parser.parse(action_date).date()
                 vote_id = hist["history_refid"]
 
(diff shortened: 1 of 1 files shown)


showerst:
I'd change this flow, because we don't know that all unseen actions will be executive. "legislature" is a better fallback.

Maybe something like:
```python
elif 'governor' in cleaned_action.lower():
    chamber = "executive"
    ...
else:
   chamber = "legislature"
```

PR #5730, opened 2026-07-10
IL, ID Misc fixes

showerst:
IL was missing bill actions due to a change in the markup of the new site.

ID didn't have a classifier for line item vetos.


jessemortenson:
Thanks, Tim - looks like I should have checked more holistically when I fixed https://github.com/openstates/openstates-scrapers/pull/5721/changes !

showerst:
No prob, those random site markup changes that don't actually break the scrapers are a real joy.

PR #5733, opened 2026-07-14
WY: extract bill abstracts from digest/summary HTML

harshitasingh360:
The API's summary/digest fields only contain PDF paths (e.g. 2026/Summaries/HB0002.pdf), which were being stored as the bill abstract. Extract the actual text from the digestHTML/summaryHTML fields instead: the digest 'AN ACT...' clause as the primary abstract and the plain-language summary as a secondary abstract.

### DIFF
--- a/scrapers/wy/bills.py
+++ b/scrapers/wy/bills.py
@@ -1,6 +1,9 @@
 import pytz
 import datetime
 import json
+import re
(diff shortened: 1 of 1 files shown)


jessemortenson (approved):
LGTM

PR #5740, opened 2026-07-23
TX: events: fix missing events due to bad selector, flipped boolean

jessemortenson:
Well I can't see who to blame but looks like this is_duplicate boolean logic has been flipped for years :grimacing: fyi @showerst


showerst:
lol.

Well it's somebody running arch so for once I can definitely say this one's not my fault =).

PR #5741, opened 2026-07-24
IL: events: fix missing events #5728

sinchana00:
The IL events scraper (scrapers/il/events.py) is reporting “no objects returned" from IL EventScraper scrape, but there IS an event it should be getting.

There was a change in the XPath that caused the issue in the IL events scraper. The heading tag changed from h4 to h2.


jessemortenson:
Looks good to me (LGTM)! Will merge

PR #5743, opened 2026-07-24
fix(fl): scrape bills incrementally instead of batching the whole session

agent-smith-ddp:
## Summary
- FL's `scrape()` wrapped the entire session's bill generator in `retry_on_connection_error(lambda: list(do_scrape_with_retry()))`, forcing every bill to fully process and sit in memory before any got saved. A crash near the end of a long session lost the whole session's progress, not just what was left.
- That whole-session retry was largely redundant: `_process_bill_list` already has its own per-item circuit breaker, and every HTTP request already retries via the `patched_get_response` monkeypatch in this file.
- Restructured to retry just the bill-list pagination walk directly, without collapsing it into a list. Bills now save to `bill_*.json` as soon as each finishes, matching every other jurisdiction's scraper.

## Test plan
- [x] Verified against real live FL 2026F data in an isolated dev checkout (not production): scrape completed with the same 5 bills / 13 vote events as before the fix.
- [x] Confirmed `bill_*.json` files now appear at different timestamps throughout the run (19:28:18, 19:29:26, 19:30:19, 19:30:30, 19:31:10) instead of all at the very end.
- [x] Confirmed `--import` and `os-text-extract archive` still work correctly against the result.
- [ ] Not yet tested against a real multi-hour production-scale session (e.g. FL's full 2026 regular session) or a genuine mid-run network failure -- worth doing before merging if that scenario matters.

### DIFF
--- a/scrapers/az/bills.py
+++ b/scrapers/az/bills.py
@@ -343,6 +343,15 @@ class AZBillScraper(Scraper):
                 """
                 ayes = action["Ayes"] if "Ayes" in action and action["Ayes"] else 0
                 nays = action["Nays"] if "Nays" in action and action["Nays"] else 0
+                # Skip summary/procedural passage actions that carry no recorded
+                # vote: AZ's feed includes "Passed" lines with no Ayes/Nays and an
+                # empty Votes[] list. Emitting a VoteEvent for those creates an empty
+                # motion that can shadow the real recorded vote downstream.
+                votes_list = (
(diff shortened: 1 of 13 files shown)


agent-smith-ddp:
Opened in error against the wrong repo -- this is an internal DDP fork change, not intended for upstream. Closing; the correct PR is against Digital-Democracy-Project/openstates-scrapers.

PR #5745, opened 2026-07-29
TX: events: Improvements

sinchana00:
Problem: 
1. The agenda was being added as one long text string that basically contains all of the text from the event page. Instead, we can capture main agenda text for that event.
2. Events are marked cancelled. However in the scraped data output these have a status property of “confirmed” (should be cancelled)
3. Every event listed on at least the has a link to a PDF document shown in the listing table. This PDF link should be added as a “document” on the event.

Fixes:
1. Added shorter format of agenda to the scraped data present in bold text on the website.
2. Correction of the events status.
3. Appended the PDF link to the scraped data as document.

### DIFF
--- a/scrapers/tx/events.py
+++ b/scrapers/tx/events.py
@@ -94,7 +94,7 @@ class TXEventScraper(Scraper, LXMLMixin):
             chair = re.sub(r"(Rep\. |Senator |Representative |Sen\. )", "", chair)
 
         plaintext = re.sub(r"\s+", " ", plaintext).strip()
-        bills = bill_re.findall(plaintext)
+        # bills = bill_re.findall(plaintext)
 
         event = Event(
             name=committee,
@@ -135,11 +135,11 @@ class TXEventScraper(Scraper, LXMLMixin):
                 text = " ".join("".join(span.itertext()).split())
                 if text and text not in results:
                     agenda_text = text.rstrip(" :")
-                    agenda = event.add_agenda_item(agenda_text)
+                    event.add_agenda_item(agenda_text)
 
-        for alpha, num in bills:
-            bill_id = f"{alpha} {num}"
-            agenda.add_bill(bill_id)
+        # for alpha, num in bills:
+        #     bill_id = f"{alpha} {num}"
+        #     agenda.add_bill(bill_id)
 
         day = datetime.strftime("%Y-%m-%d")
         videos = []
--- a/scrapers/tx/events.py
+++ b/scrapers/tx/events.py
@@ -95,13 +95,18 @@ class TXEventScraper(Scraper, LXMLMixin):
         bills = bill_re.findall(plaintext)
 
         event = Event(
-            name=committee, start_date=self._tz.localize(datetime), location_name=where, status=status
+            name=committee,
+            start_date=self._tz.localize(datetime),
+            location_name=where,
+            status=status,
         )
         event.dedupe_key = url
 
         pdf_document = url.replace("html", "pdf").replace("HTM", "pdf")
 
-        event.add_document(note="Agenda PDF", url=pdf_document, media_type="application/pdf")
+        event.add_document(
+            note="Agenda PDF", url=pdf_document, media_type="application/pdf"
+        )
 
         event.add_source(url)
 
@@ -131,7 +136,7 @@ class TXEventScraper(Scraper, LXMLMixin):
                     results.append(text.rstrip(" :"))
 
         results = ", ".join(results)
-        agenda = event.add_agenda_item(results if results else 'Agenda Not found')
+        agenda = event.add_agenda_item(results if results else "Agenda Not found")
 
         for alpha, num in bills:
             bill_id = f"{alpha} {num}"
@@ -199,7 +204,11 @@ class TXEventScraper(Scraper, LXMLMixin):
                 datetime = dt.datetime.strptime(datetime, "%A, %B %d, %Y %I:%M %p")
 
                 yield from self.scrape_event_page(
-                    session, chamber, event.attrib["href"], datetime, status="cancelled" if cancelled else "confirmed"
+                    session,
+                    chamber,
+                    event.attrib["href"],
+                    datetime,
+                    status="cancelled" if cancelled else "confirmed",
                 )
 
     def scrape_committee_upcoming(self, session, chamber):
--- a/scrapers/tx/events.py
+++ b/scrapers/tx/events.py
@@ -101,11 +101,7 @@ class TXEventScraper(Scraper, LXMLMixin):
 
         pdf_document = url.replace("html", "pdf").replace("HTM", "pdf")
 
-        event.add_document(
-            note="Agenda PDF",  
-            url=pdf_document,
-            media_type="application/pdf",        
-        )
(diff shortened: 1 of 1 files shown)


scrapers/tx/events.py
```diff
@@ -95,19 +95,48 @@ def scrape_event_page(self, session, chamber, url, datetime):
...
+        .//span/b[substring(normalize-space(), string-length(normalize-space()))=':']
+        """
+        results = []
+        for p_element in page.xpath("//p"):
+            for span in p_element.xpath(xpath):
+                text = " ".join("".join(span.itertext()).split())
+                if text and text not in results:
+                    results.append(text.rstrip(" :"))
+
+        results = ", ".join(results)
```
jessemortenson: Looks like you are creating a single agenda item, the description of which is produced by concatenating multiple strings from the bold elements on the source page. Would it be hard to make these separate agenda items? 

Looking at the source page (like https://capitol.texas.gov/tlodocs/89R/schedules/html/C5942026081309001.HTM), it looks to me that each bolded/underlined element in the main text is a separate agenda item. So I would expect "Securing Texas Land Ports of Entry," "Combating Criminal Activities on Oversize and Overweight Routes," "Shoring Up the Integrity of Export Fuel", and "Combating Public Corruption Tied to Fuel Crimes" all to be separate agenda items (ie four calls to the `add_agenda_item()` function).

scrapers/tx/events.py
```diff
@@ -95,19 +95,48 @@ def scrape_event_page(self, session, chamber, url, datetime):
...
         event = Event(
-            name=committee, start_date=self._tz.localize(datetime), location_name=where
+            name=committee,
+            start_date=self._tz.localize(datetime),
+            location_name=where,
+            status=status,
         )
         event.dedupe_key = url
 
+        pdf_document = url.replace("html", "pdf").replace("HTM", "pdf")
```
jessemortenson: Did you double check that this URL pattern always holds (swapping the file extension)? Can be a ilttle risky, vs. scraping the PDF link from the listing page

scrapers/tx/events.py
```diff
@@ -158,6 +187,11 @@ def scrape_upcoming_page(self, session, chamber, url):
             if time_elem:
                 time = time_elem[0].text_content()
 
+            # cancelled = row.xpath("./td/span[contains(@class, 'redText')]").  # Can be used as a backup if the text changes in the future.
+            cancelled = row.xpath("./td/span[contains(text(), 'Canceled')]")
```
jessemortenson: I think this is probably fine, so these are just thoughts for consideration:

Another option would be to shorten the text it is checking to just `Cancel` so if there are variant spellings of Canceled/Cancelled both would be caught.

A more elaborate option: grab the relevant text strings wtih xpath, but then use a regex (`re`) to test whether they match possible strings: something like `"/canceled|cancelled/", re.IGNORECASE`

jessemortenson (changes requested):
Overall this is pretty close, just one change requested (multiple agenda items instead of concatenating into one agenda item), and two suggestions for consideration

jessemortenson:
HI Sinchana, a few housekeeping details to please take care of on PRs in this state (a review has been made, and you've made some changes):

1. Click the Re-request Review button up in the reviewers section (looks like two arrows in a circle), which notifies me that the PR is ready for re-review. Also makes it so I can quickly filter my PRs list to those waiting on me. 
2. If I ask a question in a comment ("Did you double check that this URL pattern always holds (swapping the file extension)?"), please respond to that with your own comment. 
3. Usually a good practice to just hit the "resolve comment" button on any review comments that you have addressed with code changes. Makes it easy to see that you've checked those off.

scrapers/tx/events.py
jessemortenson: Looks like you're adding all related bills to the last agenda item that was created. Is there a reasonable way to associated the bill with the specific agenda item it relates to? I can't remember if the page structure allow for that or not. 

If it's not reasonable, then just associating all bills with an arbitrary agenda item is reasonable. But want to check on that before just going forward.

jessemortenson (approved):
Overall Looks Good to Me (LGTM). But please share a response for one question where there may be a possible improvement.

scrapers/tx/events.py
sinchana00: Hi @jessemortenson 

This looks like it was something that I might have missed, thank you for pointing that out.
 
I would require some clarity here.
Analysing the issue I was able to identify one of those scenarios and the relative PDF is https://capitol.texas.gov/tlodocs/89R/schedules/pdf/C6402026082509001.PDF
 
so in this case there is a bill thats getting attached and its attached to the last item and hence the overall agenda looks like this
 
"agenda": [
    {
      "description": "Prioritizing Public Safety at Rail-Grade Crossings",
      "classification": [],
      "related_entities": [],
      "subjects": [],
      "media": [],
      "notes": [],
      "order": 0,
      "extras": {}
    },
    {
      "description": "Studying the Deployment of Autonomous Vehicle Technology",
      "classification": [],
      "related_entities": [],
      "subjects": [],
      "media": [],
      "notes": [],
      "order": 1,
      "extras": {}
    },
    {
      "description": "Strengthening Commercial Driver License Standards and Oversight",
      "classification": [],
      "related_entities": [
        {
          "name": "SB 2807",
          "entity_type": "bill",
          "note": "consideration",
          "bill_id": "~{\"identifier\": \"SB 2807\"}"
        }
      ],
      "subjects": [],
      "media": [],
      "notes": [],
      "order": 2,
      "extras": {}
    }
  ],
 
should the related entities be attached to the corresponding agenda item (which is necessary)?
which in case the agenda could look like
 
"agenda": [
    {
      "description": "Prioritizing Public Safety at Rail-Grade Crossings",
      "classification": [],
      "related_entities": [],
      "subjects": [],
      "media": [],
      "notes": [],
      "order": 0,
      "extras": {}
    },
    {
      "description": "Studying the Deployment of Autonomous Vehicle Technology",
      "classification": [],
      "related_entities": [
 {
          "name": "SB 2807",
          "entity_type": "bill",
          "note": "consideration",
          "bill_id": "~{\"identifier\": \"SB 2807\"}"
        }],
      "subjects": [],
      "media": [],
      "notes": [],
      "order": 1,
      "extras": {}
    },
    {
      "description": "Strengthening Commercial Driver License Standards and Oversight",
      "classification": [],
      "related_entities": [
      ],
      "subjects": [],
      "media": [],
      "notes": [],
      "order": 2,
      "extras": {}
    }
  ],
 
 
OR
 
was it something that was initially added because the entire agenda was in a raw text format ? (not necessary with the current changes and can keep the related entities blank for all)

jessemortenson: You asked

> should the related entities be attached to the corresponding agenda item (which is necessary)?

Yes, ideally SB 2807 would be attached to the "Studying the Deployment of Autonomous Vehicle Technology" agenda item, because that is the agenda item that mentions that bill being considered.

If that is reasonable to accomplish given state of the source site and code, then I'd so go for that. If it would be days of work, then it's probably not worth it.

sinchana00: As discussed, I would like to take up this task under a new ticket (DATA-5739)

jessemortenson (approved):
LGTM, merging as is, given the follow up ticket

PR #5746, opened 2026-07-30
Fix for missing sponsers

harshitasingh360:
NH's General Court changed several bulk-data files the scraper depends on. Pushing a fallback fix so sponsor data doesn't silently drop.

Details from Claude:
NH's legislators file was removed entirely, and several other files the scraper depends on (bill records, actions, docket data, roll-call data) started returning HTTP 200 with empty (0-byte) bodies instead of erroring — only the sponsor records file still had real data. Since sponsors are matched against bill and legislator entries, this caused sponsor rows to silently drop or trigger "can't find person" warnings. NH's newer members roster page was checked as a replacement but lacks the ID fields needed to join against sponsor records.

The fix adds a fallback that parses sponsor info directly from each bill's status page HTML (already fetched for version info, so no extra network cost), verified against real bills with single and multiple sponsors. Also added a safeguard that checks the bill records file size before processing, so future empty-data situations fail loudly instead of silently completing with zero bills and no sponsors.

Output - [nh 3.zip](https://github.com/user-attachments/files/30792270/nh.3.zip)

### DIFF
--- a/scrapers/nh/bills.py
+++ b/scrapers/nh/bills.py
@@ -447,12 +447,8 @@ class NHBillScraper(Scraper):
                         version_id = self.versions_by_lsr[lsr]
                         self.add_source(self.bills[lsr], version_id)
                         bill_page = self.get_bill_page(version_id)
-                        self.scrape_versions_from_web_2026(
-                            self.bills[lsr], bill_page
-                        )
-                        self.scrape_sponsors_from_bill_page(
-                            self.bills[lsr], bill_page
-                        )
+                        self.scrape_versions_from_web_2026(self.bills[lsr], bill_page)
+                        self.scrape_sponsors_from_bill_page(self.bills[lsr], bill_page)
                     else:
                         self.warning(
                             f"Missing version_id for {bill_id}, can't build bill page"
@@ -520,9 +516,7 @@ class NHBillScraper(Scraper):
                 # table, regardless of whether we're adding the sponsorship
                 # itself: it's useful metadata for downstream consumers.
                 if employee in self.legislators:
-                    self.bills[lsr].extras["_code"] = self.legislators[employee][
-                        "seat"
-                    ]
+                    self.bills[lsr].extras["_code"] = self.legislators[employee]["seat"]
 
                 # Skip the fallback add if the HTML-page scrape already
                 # populated sponsors for this bill.
--- a/scrapers/nh/bills.py
+++ b/scrapers/nh/bills.py
@@ -104,6 +104,58 @@ class NHBillScraper(Scraper):
             except ValueError:
                 continue
 
+        # Fallback: scrape sponsors from the same bill page.
+        # NH's bulk-data legislators.txt was removed from /dynamicdatadump/ in 2026,
+        # so the bulk-data sponsor lookup no longer resolves employee IDs to names.
+        # The per-bill HTML page still lists sponsors reliably.
+        self.scrape_sponsors_from_bill_page(bill_with_sources, bill_page)
+
+    def scrape_sponsors_from_bill_page(self, bill, bill_page):
+        """
+        Parse sponsors from a NH bill status page (billinfo.aspx).
+
+        Sponsors live inside a div with id="dvSponosrs" (sic - misspelled in NH HTML).
+        Each sponsor is an <a> whose title attribute contains the fully qualified
+        name (e.g. "Rep. Michael Cahill (D)"). The prime sponsor is marked by
+        leading "(Prime)" text inside the <a>.
+        """
+        sponsor_links = bill_page.xpath(
+            "//div[@id='dvSponosrs']//a[contains(@class, 'Member')]"
+        )
+        added = 0
+        for link in sponsor_links:
+            # Prefer the title attribute; it has the clean "Rep./Sen. Name (Party)" form
+            title_attr = link.get("title", "").strip()
+            link_text = link.text_content().strip()
+
+            is_primary = "(Prime)" in link_text
+
+            if title_attr:
+                # Strip leading honorific ("Rep. " / "Sen. ") and any trailing " (D)" / " (R)" party marker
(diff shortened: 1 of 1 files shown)


jessemortenson:
Thanks for sharing the nh.zip of the scraper output. However i'm finding the data to be quite unexpected: seems like none of the bills have any sponsorships (and also no actions, no versions, no abstracts...). I definitely cannot evaluate the sponsorships fix if there is no sponsor data in the scraper output. Is there a problem with this output data?

scrapers/nh/bills.py
```diff
@@ -296,15 +360,24 @@ def scrape_chamber(self, chamber, session, scrape_from_web):
...
 
-            last_line = []
-            for line in (
-                self.get(
-                    f"https://gc.nh.gov/dynamicdatadump/LSRs.txt?x={self.cachebreaker}",
-                    verify=False,
+            # Guardrail: NH's bulk data dump has been intermittently returning empty
+            # files (e.g. LSRs.txt returning 0 bytes). Detect that upfront so the run
+            # fails loudly rather than silently emitting an empty scrape.
+            lsrs_content = self.get(
```
jessemortenson, line 361: This is a good call

scrapers/nh/bills.py
```diff
@@ -104,6 +104,56 @@ def scrape_versions_from_web_2026(self, bill_with_sources):
             except ValueError:
                 continue
 
+        # Fallback: scrape sponsors from the same bill page.
+        # NH's bulk-data legislators.txt was removed from /dynamicdatadump/ in 2026,
+        # so the bulk-data sponsor lookup no longer resolves employee IDs to names.
+        # The per-bill HTML page still lists sponsors reliably.
+        self.scrape_sponsors_from_bill_page(bill_with_sources, bill_page)
```
jessemortenson: I think the overall approach to fall back to scraping the sponsors from the bill page makes sense. 

The downside of fitting this into the existing flow, however, is that the logic gets nested pretty deep. That can make it hard for someone to understand how the scraper is working later. The nesting, as I read it, is basically (pseudocode):

```
  if NOT scrape_from_web:
    for line in lsrs_content.split("\n"):
      if chamber/session_yr/no expanded_bill_id
        if lsr in self.versions_by_lsr:
          self.scrape_versions_from_web_2026(self.bills[lsr])
            self.scrape_sponsors_from_bill_page()
```

It's not obvious why someone would assume that scraping sponsors is nested inside of `scrape_versions_from_web_2026()`

So I recommend some refactor here as well, so that this sponsors logic is not so deeply nested. In my mind, the flow would look something like:

```
  if NOT scrape_from_web:
    for line in lsrs_content.split("\n"):
      if chamber/session_yr/no expanded_bill_id
        if lsr in self.versions_by_lsr:
          bill_page = get_bill_page()
          self.scrape_versions_from_web_2026(self.bills[lsr], bill_page)
          self.scrape_sponsors_from_bill_page(bill_page)
```

I wanted to examine whether the sponsor function truly needs to be nested within the check for `if lsr in self.versions_by_lsr:` I'm not sure, because as noted the NH page is hard to work with (wasn't loading any of the data sources when I tried just now). But a possible variant would be:

```
  if NOT scrape_from_web:
    for line in lsrs_content.split("\n"):
      if chamber/session_yr/no expanded_bill_id
        bill_page = get_bill_page()
        self.scrape_sponsors_from_bill_page(bill_page)
        if lsr in self.versions_by_lsr:
          self.scrape_versions_from_web_2026(self.bills[lsr], bill_page)
```

jessemortenson (changes requested):
Overall the approach looks good, but please take a look at a) the question I raised about the attached data zip, and b) the suggestion to refactor so that the sponsorship logic is not so deeply nested in the overall control flow of the scrape. Happy to talk through that second one if you want.

harshitasingh360:
> Thanks for sharing the nh.zip of the scraper output. However i'm finding the data to be quite unexpected: seems like none of the bills have any sponsorships (and also no actions, no versions, no abstracts...). I definitely cannot evaluate the sponsorships fix if there is no sponsor data in the scraper output. Is there a problem with this output data?

Thanks for pointing it out Jesse, let me try verifying this with tha fresh run and producing a new zip file. The website seems down now because I am not able to reproduce the new output anymore. I have also take the refactor into consideration and would be changing the code to implement that.

harshitasingh360:
> Thanks for sharing the nh.zip of the scraper output. However i'm finding the data to be quite unexpected: seems like none of the bills have any sponsorships (and also no actions, no versions, no abstracts...). I definitely cannot evaluate the sponsorships fix if there is no sponsor data in the scraper output. Is there a problem with this output data?

The zip was produced during a broader NH outage, which is why every bill came back with no sponsors, actions, or versions. I checked and NH's endpoints are all currently degraded:

- dynamicdatadump/LSRs.txt: 1 byte
- dynamicdatadump/LsrsOnly.txt: 0 bytes
- dynamicdatadump/legislators.txt: 3 bytes (BOM only)
- rssFeeds/rssQueryResults.aspx: 60s read timeouts
- bill_status/legacy/bs2016/bill_docket.aspx: returning anti-bot JS blob (empty <head></head>)

I'll re-run once NH's endpoints recover and post a fresh zip so the sponsors fix can actually be evaluated.

showerst:
Just a side note, i've noticed that NH's bulk data sometimes is broken during the day (eastern US time) then shows back up at night. I'm not sure if this is policy or coincidence, but it's worth a check.

jessemortenson:
Thanks for the note, Tim.

@harshitasingh360 Thanks for that note as well: and that's a good reminder to check the data before you upload it as an example of the code change. Let's proceed with what you suggested re: trying to get the scraper to run - let's both try to run this like once a day until we can get a good scraper output based on this branch. We won't be able to close this PR out until we can verify the data output.

jessemortenson (approved):
LGTM, merging

PR #5752, opened 2026-08-06
OK: guard against null title/datetime in events scraper

harshitasingh360:
The OK House API sometimes returns a null title field. Pushing a fix that skips events where title/date is missing. Output - [ok.zip](https://github.com/user-attachments/files/30792814/ok.zip)

Details from Claude:
The OK events scraper was crashing whenever the upstream House API or Senate meeting pages returned records with missing fields. This PR adds defensive guards so that malformed records are skipped with a warning instead of taking down the whole scrape.
On the Senate side, the scraper previously assumed every meeting page would have both a title and a start datetime. When a page loaded without either of those fields, it would raise an error. The scraper now checks for those fields before using them and logs a warning if they are missing.
On the House side, the API occasionally returns event records where the title or start datetime is null. Without those fields, a valid event cannot be built, and even attempting to save such a record causes the whole scrape to fail. The new guards catch these records early and skip them with a descriptive warning.
There was also a small latent bug in how missing locations were handled. The fallback string "See agenda" was being written into the wrong place, so events with no location weren't actually getting the fallback — it was silently overwritten. That is now fixed.
Finally, a few other fields returned by the House API (such as document links and agenda items) were assumed to always be lists, but the API sometimes returns null instead. And individual agenda items were sometimes missing their HTML content, which crashed the HTML parser. Both of these are now handled defensively.

### DIFF
--- a/scrapers/ok/events.py
+++ b/scrapers/ok/events.py
@@ -87,7 +87,6 @@ class OKEventScraper(Scraper):
         page.make_links_absolute(url)
 
         title_nodes = page.xpath("//span[contains(@class,'field--name-title')]/text()")
-        
         if not title_nodes or not title_nodes[0].strip():
             self.warning(f"Skipping senate event with no title: {url}")
             return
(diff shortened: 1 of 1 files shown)


jessemortenson (approved):
Looks Good to Me (LGTM)!

PR #5753, opened 2026-08-06
MO: Bills: Index Error issue

sinchana00:
Problem:
The bill_sponsor_link selector no longer returns any data. The expected anchor (<a href>) no longer exists at the  location in the updated page structure, so the resulting list is empty and hence the error.

Fixes:
Extracting the sponsor link at the same point where we extract the bill identifier, since both values are available in that section of the page.

### DIFF
--- a/scrapers/mo/bills.py
+++ b/scrapers/mo/bills.py
@@ -235,21 +235,21 @@ class MOBillScraper(Scraper, LXMLMixin):
         if bill_lr:
             bill.extras["MO_BILL_LR"] = bill_lr
 
-        for cosponsor_link in bill_page.xpath(
-            '//div[contains(@class, "detail-grid__item") and contains(string(.), "Co-Sponsors")]/div[1]/a'
-        ):
-            if "Senators" in cosponsor_link.xpath("@href")[0]:
-                chamber = "upper"
-            else:
-                chamber = None
-
-            bill.add_sponsorship(
-                cosponsor_link.text_content(),
-                entity_type="person",
-                classification="cosponsor",
-                primary=False,
-                chamber=chamber,
-            )
+        # Based on the analysis of the MO website, co-sponsors are typically Senators,
+        # so we assume the chamber is "upper" for co-sponsor entities.
+        chamber = "upper"
+        co_spons = '//div[contains(@class, "detail-grid__item") and contains(string(.), "Co-Sponsors")]/div/span'
+        co_sponsors_elements = bill_page.xpath(co_spons)
+        if co_sponsors_elements:
+            for element in co_sponsors_elements:
+                name = element.text_content()
+                bill.add_sponsorship(
+                    name,
+                    entity_type="person",
+                    classification="cosponsor",
+                    primary=False,
+                    chamber=chamber
+                )
 
         # get the actions
         actions_url = f"{bill_url}&handler=Actions"
@@ -344,7 +344,7 @@ class MOBillScraper(Scraper, LXMLMixin):
         index_page = lxml.html.fromstring(index_page)
 
         for card_header in index_page.cssselect("div.bill-card-header"):
-            bill_sponsor_handler = None
+            bill_sponsor_link = None
             bill_identifier_element = card_header.cssselect("div.bill-number a")
             if not bill_identifier_element:
                 self.warning("Missing bill identifier element. Skipping.")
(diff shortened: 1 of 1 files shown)


scrapers/mo/bills.py
```diff
@@ -344,10 +343,30 @@ def _scrape_upper_chamber(self, session):
         index_page = self.get(index_url).text
         index_page = lxml.html.fromstring(index_page)
 
-        for link in index_page.cssselect("div.bill-number a"):
-            bill_identifier = link.text_content().strip()
+        for card_header in index_page.cssselect("div.bill-card-header"):
+            bill_sponsor_handler = None
```
jessemortenson: Kind of a small thing, but I would call this variable `bill_sponsor_link`, since that is what the function that consumes it as an input (`_parse_senate_billpage()`) calls it. That creates one less translation my brain has to do when reading the scraper. 

Also: `handler` in the software context, in my experience, has a connotation of being  a function, not a literal value (a string in this case). So I'd avoid naming a variable "handler" unless it is a function that handles something, or unless the domain demands it ("baggage handler" in the airport domain, for example).

jessemortenson (changes requested):
I only have partial feedback now because I'm running the scraper, and I'm waiting for full data to come back. However, based on partial data, I think we need at least one change before we can merge this in. 

It looks like this logic is failing to fetch Senate co-sponsors. (Possibly House also, but I don't have data examples yet). In the output, I'm seeing just one bill sponsor for each bill, but there are MO bills that have multiple sponsors.

Examples:

https://www.senate.mo.gov/BillTracking/Bills/BillInformation?year=2026&billid=479

Sponsorships data as output by this scraper looks like:

```json
[{'name': 'Brattin, Rick', 'classification': 'primary', 'entity_type': 'person', 'primary': true, 'person_id': '~{"chamber": "upper", "name": "\r\n                                            Brattin, Rick\r\n                                "}', 'organization_id': NULL}]
```

As you can see, the cosponsor `Coleman, Mary Elizabeth` is missing. Checking the Open States DB, we do have both sponsors:

```sql
SELECT b.identifier, s.name AS sponsor_name, s.entity_type, s.primary, s.*
FROM opencivicdata_billsponsorship s
INNER JOIN opencivicdata_bill b on s.bill_id = b.id
INNER JOIN opencivicdata_legislativesession ls ON b.legislative_session_id = ls.id
INNER JOIN opencivicdata_jurisdiction j ON ls.jurisdiction_id = j.id
WHERE j.name = 'Missouri'
AND ls.identifier = '2026'
AND b.identifier IN ('SB 856', 'SB 877')
```

```text
+----------+-----------------------+-----------+-------+
|identifier|sponsor_name           |entity_type|primary|
+----------+-----------------------+-----------+-------+
|SB 856    |Brattin, Rick          |person     |true   |
|SB 856    |Coleman, Mary Elizabeth|person     |false  |
+----------+-----------------------+-----------+-------+
```

SB 877 is another example, where sponsorships from scraper looks like

```json
[{'name': 'Fitzwater, Travis', 'classification': 'primary', 'entity_type': 'person', 'primary': true, 'person_id': '~{"chamber": "upper", "name": "\r\n                                            Fitzwater, Travis\r\n                                "}', 'organization_id': NULL}]
```

but https://www.senate.mo.gov/BillTracking/Bills/BillInformation?year=2026&billid=275 also shows Trent Curtis as a co-sponsor.

Please modify so that we are getting both sponsors and co-sponsors.

jessemortenson:
FYI looks like Tim merged in a contributor's fix - that does not address the Senate cosponsor issue - here: https://github.com/openstates/openstates-scrapers/pull/5758

So now there is a conflict that needs to be resolved before this gets merged in.

scrapers/mo/bills.py
```diff
@@ -236,21 +235,21 @@ def _parse_senate_billpage(self, bill_url, bill_identifier, year):
...
-                chamber = None
-
-            bill.add_sponsorship(
-                cosponsor_link.text_content(),
-                entity_type="person",
-                classification="cosponsor",
-                primary=False,
-                chamber=chamber,
-            )
+        # Based on the analysis of the MO website, co-sponsors are typically Senators,
```
jessemortenson: I would change the language here a bit, just to specify that we're talking about the MO Senate webpage for this logic (` _parse_senate_billpage()`). The MO House website, in contrast, lists only House members as co-sponsors. I'd suggest:

> MO Senate website lists Senate sponsors, so we assume chamber is upper

jessemortenson (approved):
Looks good to me! Please fix the merge conflict, ideally adjust the comment language as noted below, and go ahead and merge in.

PR #5754, opened 2026-08-06
CO events scrapper code change for website redesign

harshitasingh360:
Fixed the Colorado events scraper, which was failing on every run after the source site was redesigned and the old selectors and pagination no longer matched. Rewrote the parser against the new page structure, added a fallback for the in-session layout, and switched empty results to raise EmptyScrape instead of erroring. 

Output -[co.zip](https://github.com/user-attachments/files/30799111/co.zip)

### DIFF
--- a/scrapers/co/events.py
+++ b/scrapers/co/events.py
@@ -3,7 +3,6 @@ import pytz
 
 import lxml
 from openstates.scrape import Scraper, Event
-from openstates.exceptions import EmptyScrape
 
 from utils.events import match_coordinates
 from utils import LXMLMixin
@@ -78,12 +77,6 @@ class COEventScraper(Scraper, LXMLMixin):
                         yielded += 1
                         yield event
 
-        if yielded == 0:
-            # Neither layout matched or the schedule is genuinely empty
-            # (e.g. between interim and session). Raise EmptyScrape so
-            # openstates records a clean empty run instead of ScrapeError.
-            raise EmptyScrape
-
     def scrape_event_row(self, row: lxml.html.HtmlElement, start_day: str):
         start_time = self.clean(row.xpath("td[1]"))
 
@@ -95,17 +88,17 @@ class COEventScraper(Scraper, LXMLMixin):
             com_name = self.clean(row.xpath("td[2]"))
 
         if not com_name or not start_time:
+            self.warning(
+                f"Skipping event row with missing name or time: "
+                f"name={com_name!r}, time={start_time!r}"
+            )
             return
 
         location = self.clean(row.xpath("td[3]"))
         location = f"{location}, 200 E Colfax Ave, Denver, CO 80203"
 
         start = f"{start_day} {start_time}"
-        try:
-            start = dateutil.parser.parse(start, fuzzy=True)
-        except (ValueError, OverflowError):
-            self.warning(f"Could not parse date/time for {com_name}: {start!r}")
-            return
+        start = dateutil.parser.parse(start, fuzzy=True)
         start = self._tz.localize(start)
 
         event = Event(com_name, start, location, status="tentative")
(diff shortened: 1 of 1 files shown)


scrapers/co/events.py
```diff
@@ -39,72 +35,97 @@ def scrape(self):
...
+                heading = day_block.cssselect("h3")
+                if not heading:
+                    continue
+                event_date = self.clean(heading)
+                for row in day_block.cssselect("tbody tr"):
+                    for event in self.scrape_event_row(row, event_date):
+                        yielded += 1
+                        yield event
+
+        if yielded == 0:
```
jessemortenson: Given that the layout is *possibly* in flux, I worry a little bit about the possibility that the "interim" layout goes away, and then "normal" layout comes back - but maybe different. In that case I'm worried that raising EmptyScrape here will give us a false positive: the scrape finishes "successfully" but maybe there were events on the page we're just missing because the "normal" layout came back with different markup.

So my request is to drop the `raise EmptyScrape` logic here, so instead we just let the scrape fail normally if no events are found. We can always come back and adjust later.

scrapers/co/events.py
```diff
@@ -39,72 +35,97 @@ def scrape(self):
...
+
+        # Committee/activity name: prefer the link text, fall back to cell text
+        com_link = row.xpath("td[2]//a")
+        if com_link:
+            com_name = self.clean(com_link)
+        else:
+            com_name = self.clean(row.xpath("td[2]"))
+
+        if not com_name or not start_time:
+            return
```
jessemortenson, line 95: Do we expect to be missing `com_name` and/or `start_time` on a regular basis? If not, then please add a warning log  above this `return` so there is some messaging in log in case someone is investigating later about whether events are being captured.

scrapers/co/events.py
```diff
@@ -39,72 +35,97 @@ def scrape(self):
...
+
         location = self.clean(row.xpath("td[3]"))
         location = f"{location}, 200 E Colfax Ave, Denver, CO 80203"
 
         start = f"{start_day} {start_time}"
-        start = dateutil.parser.parse(start, fuzzy=True)
+        try:
+            start = dateutil.parser.parse(start, fuzzy=True)
+        except (ValueError, OverflowError):
+            self.warning(f"Could not parse date/time for {com_name}: {start!r}")
```
jessemortenson: Are you seeing this happen with the current scrape? 

I think this is likely a case where we *do* want the scrape to fail. If there is a datetime listed in the event, and our logic fails to parse it, I think we want the scrape to fail. That's the signal we use to investigate and fix so we don't miss events. 

Websites putting up weird datetime strings for legislative events is pretty common. If we just return+log on all of those we're gonna end up missing a significant fraction of events (not big overall, but in a given jurisdiction it can be like 20% sometimes).

So unless I'm missing something (hence my question above), my request here is to drop the try/catch and just let the scrape fail on ValueError or OverflowError

jessemortenson (changes requested):
This is looking pretty good, good job! There are two changes I'd like to request before signing off

scrapers/co/events.py
harshitasingh360: Sure Jesse, that does make sense. I've dropped the raise EmptyScrape logic.

scrapers/co/events.py
harshitasingh360: The pattern is visible across other legislature websites hence the changes Although in CO, all 22 events in the last run had valid names and time. I have added a warning log above the return as requested.

scrapers/co/events.py
harshitasingh360: I agree with what you are saying. I have dropped the try/catch so the scrape fails on ValueError or OverflowError

jessemortenson (approved):
LGTM! Merging

PR #5755, opened 2026-08-07
TX: Events: Improvements to agenda, related entities addition 

sinchana00:
Problem: Related bills were missing from the agenda items

Fixes: Updated regex to identify and match the bill entities, then appended the matched bills to their respective agenda items.

### DIFF
--- a/scrapers/tx/events.py
+++ b/scrapers/tx/events.py
@@ -152,14 +152,7 @@ class TXEventScraper(Scraper, LXMLMixin):
 
                         bills = bill_re.findall(normalized)
                         for alpha, num in bills:
-                            orig = re.search(
-                                rf"(Senate Bill|House Bill|SB|HB|SJR|HJR|SCR|HCR|SR|HR)\s+{num}",
-                                full_text,
-                                re.IGNORECASE,
-                            )
-                            bill_id = (
-                                f"{orig.group(1)} {num}" if orig else f"{alpha} {num}"
-                            )
+                            bill_id = f"{alpha} {num}"
                             if bill_id not in seen_bills:
                                 seen_bills.add(bill_id)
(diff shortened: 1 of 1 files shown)


scrapers/tx/events.py
```diff
@@ -135,11 +137,32 @@ def scrape_event_page(
...
+
+                        bills = bill_re.findall(normalized)
+                        for alpha, num in bills:
+                            orig = re.search(
+                                rf"(Senate Bill|House Bill|SB|HB|SJR|HJR|SCR|HCR|SR|HR)\s+{num}",
+                                full_text,
+                                re.IGNORECASE,
+                            )
+                            bill_id = (
+                                f"{orig.group(1)} {num}" if orig else f"{alpha} {num}"
```
jessemortenson: I'm noticing that the bill identifiers (`bill_id` here) are sometimes in a long format like "Senate Bill 1814" - can you change so that it is always added to the agenda item as the short form (ie "SB 1814")?

Here's the data output I'm seeing:

Image

jessemortenson (changes requested):
Good work, I think this is close. One request re: the form of the bill identifier string that is added as a bill to the agenda item. Matching the bill identifier format that we use in the bill scraper (ie "SB 1234") will make it so downstream logic can match the string to an existing bill.

scrapers/tx/events.py
sinchana00: Yes, thank you for letting me know. I was not sure so I had added the exact value from the website. I have made the changes and now it has the shortened version which is all the same format now.

[tx.zip](https://github.com/user-attachments/files/30845004/tx.zip)

jessemortenson (approved):
LGTM - merging

PR #5758, opened 2026-08-10
MO: make the Senate sponsor link optional (fixes IndexError on every upper-chamber bill)

robshrum:
### What breaks

`_parse_senate_billpage` reads the sponsor's chamber out of a hyperlink:

```python
bill_sponsor_link = bill_page.xpath(
    '//div[contains(@class, "detail-grid__item") and contains(string(.), "Sponsor")]/div[1]/a/@href'
)[0]
```

The Missouri Senate site no longer renders the sponsor as a link — the value is now a bare `<span>`. The xpath legitimately matches nothing, `[0]` raises `IndexError`, and because it raises before `add_sponsorship` the exception takes down the **entire upper-chamber scrape**, not just one bill.

### Why this fix

The sponsor **name** xpath was never broken — only the link was. That link fed nothing except the chamber assignment, and this is `_parse_senate_billpage`, so a Senate bill's sponsor is a Senator.

So the link becomes optional and `chamber` defaults to `"upper"` when it is absent. When the link *is* present the behaviour is byte-for-byte what it was, so this is safe if MO restores the markup.

### Verification

Run live against the current site: **52 bills scraped**, where the scrape previously died on the first one.

### DIFF
--- a/scrapers/mo/bills.py
+++ b/scrapers/mo/bills.py
@@ -216,14 +216,20 @@ class MOBillScraper(Scraper, LXMLMixin):
         bill_sponsor = bill_page.xpath(
             '//div[contains(@class, "detail-grid__item") and contains(string(.), "Sponsor")]/div[1]'
         )[0].text_content()
+        # The Senate site formerly rendered the sponsor as a hyperlink; it now
+        # renders a bare <span>, so this xpath legitimately matches nothing.
+        # Indexing [0] unconditionally raised IndexError and took down the whole
+        # upper-chamber scrape.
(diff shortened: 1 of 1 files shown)


jessemortenson:
FYI this code doesn't include co-sponsors on the Senate side (Open States database had some senate cosponsors last week, now it has none). We're working on a fix that includes the co-sponsors data https://github.com/openstates/openstates-scrapers/pull/5753

showerst:
Sorry about that @jessemortenson -- I didn't notice there was another stream going.

jessemortenson:
No worries, it's a work in progress

PR #5760, opened 2026-08-12
MO: Bills: Incorrect title on Bills

sinchana00:
Problem:
For MO Senate bills, the current bill summary is captured as the bill title

Fixes:
The correct title was extracted but during bill object creation title was assigned with the bill_desc, which is now changed to bill_title


jessemortenson (approved):
LGTM, merging

PR #5766, opened 2026-08-21
NC: Events: Missing events and incorrect time

sinchana00:
Time defaulting to midnight: Time extraction used /text() which returned empty when the time was wrapped in an <a> tag. This caused events to fall back to midnight. Switched to .text_content() to handle this.

Certain event types being dropped: The scraper only processed events with a /Committees/ link and skipped anything using an /LegislativeCalendarEvent/ "Event details" link. The current live month only had the latter type, resulting in zero scraped events. Added a fallback to handle these.

Badge text leaking into location: Strip logic was only catching "STREAM" (uppercase). The actual badge text is "Stream", and "Ended" badges were also slipping through. Fixed both.

Type mismatch in time parsing fallback: When time couldn't be parsed normally, the fallback returned a date-only value while the rest of the code worked with full date-and-time values. Fixed to ensure both paths return the same type.

### DIFF
--- a/scrapers/nc/events.py
+++ b/scrapers/nc/events.py
@@ -65,21 +65,13 @@ class NCEventScraper(Scraper, LXMLMixin):
                         'span[contains(@class, "text-dark font-weight-bold")]/text()'
                     )[0].strip()
                     chamber = chamber.replace(":", "")
-                committee_links = event_row.xpath('a[contains(@href,"/Committees/")]')
-                is_committee = bool(committee_links)
-                if is_committee:
-                    com_link = committee_links[0]
-                    com_name = com_link.text_content().strip()
-                    com_name = f"{chamber} {com_name}".strip()
-                    com_url = com_link.xpath("@href")[0]
-                else:
-                    details_links = row.xpath('.//a[@title="Event details"]')
-                    if not details_links:
-                        continue
-                    com_link = details_links[0]
-                    com_name = event_row.text_content().strip()
-                    com_name = f"{chamber} {com_name}".strip()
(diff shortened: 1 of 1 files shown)


scrapers/nc/events.py
```diff
@@ -61,23 +65,29 @@ def scrape(self):
                         'span[contains(@class, "text-dark font-weight-bold")]/text()'
                     )[0].strip()
                     chamber = chamber.replace(":", "")
-
-                # sometimes there are unlinked events, usually just press conferences
-                if not event_row.xpath('a[contains(@href,"/Committees/")]'):
```
jessemortenson, line 66: I think we actually want to keep this piece of logic that checks if the event is a Committee hearing. Even though that means the scraper will not output any events at this time of year. We generally want legislative events to consist of hearings held by actual committees of the legislature. Sometimes this is a judgement call, but the fact that a prior author made the judgement call to limit the events is persuasive to me.

jessemortenson (changes requested):
Overall Looks Good to Me, but there is one change I want to request before we merge in. The date and other small fixes look good, but I think we want to keep the filter for just events that are committee events.

scrapers/nc/events.py
sinchana00: I revered the changes I had made to this logic

sinchana00:
[nc.zip](https://github.com/user-attachments/files/31364413/nc.zip)

Ran the scraper after the changes and it picked up one valid event that was there, time looks correct too, no more midnight defaulting.

jessemortenson (approved):
LGTM, merging!

PR #5768, opened 2026-08-24
DE: fix bill text/version links broken by site relabel + relative URLs

alexkost819:
FULL DISCLAIMER: This is an AI PR, but 

## Summary
- `legis.delaware.gov` renamed the bill-text label from `Original Text:` to `Original / Not Amended:`, so the xpath in `scrape_bill` matched zero links for every DE bill regardless of session or final status (`became_law` included) — this is the root cause of the reported 1,055 bills with zero text/resource links.
- The site also switched several previously-absolute hrefs (bill text, session-law chapter citation, legislator detail) to site-relative URLs. Even with the label fixed, these would have produced broken relative URLs in the scraped data, so links are now resolved against `https://legis.delaware.gov` via `urljoin`.
- Confirmed no documented DE API for bill versions/fiscal notes exists — the TODO in the file was speculative; HTML scraping is still the only path, it just needed the label/URL fixes above.
- Spot-checked live pages for the reported `became_law` examples (SB 220, etc.) — the "Original / Not Amended:" section with working HTML/PDF links exists in current HTML but the old xpath matched nothing.
## Before / After

Ran `os-update de bills --scrape session=153 start_page=1 end_page=2` in the project's docker scrape image, once before this fix and once after, same 36 bills / 47 vote events both runs (no bill-count regression):

| | bills | bills with `versions` | bills with `documents` (fiscal notes) |
|---|---|---|---|
| before | 36 | 8 | 8 |
| after | 36 | **36** | 8 |

(fiscal-note document links were already absolute on-page, so their count is unchanged — only the version-link xpath and relative-URL handling were broken.)

### Example: `HB 447`

**Before** — `versions` field:
```json
[]
```

After — versions field:
```json
[
  {
    "note": "Bill Text",
    "links": [
      {
        "url": "https://legis.delaware.gov/json/BillDetail/GenerateHtmlDocument?legislationId=143414&legislationTypeId=1&docTypeId=2&legislationName=HB447",
        "media_type": "text/html"
      },
      {
        "url": "https://legis.delaware.gov/json/BillDetail/GeneratePdfDocument?legislationId=143414&legislationTypeId=1&docTypeId=2&legislationName=HB447",
        "media_type": "application/pdf"
      }
    ],
    "date": "",
    "classification": ""
  }
]
```
Before the fix, the scraper's> reading "Original Text:".DE's site now renders that landed:", so the xpath matched nothing on every bill — versions was empty across the board, regardless of session or bill status. Full output zips are attached in the release linked above if you want to diff more examples.

Output zips (pupa-format JSON) for both runs:
- before: https://github.com/alexkost819/openstates-scrapers/releases/download/de-bill-text-test-scrape/de-scrape-before.zip
- after: https://github.com/alexkost819/openstates-scrapers/releases/download/de-bill-text-test-scrape/de-scrape-after.zip

- [x] `python3 -m py_compile scrapers/de/bills.py`
- [x] Docker test-scrape run before/after, diffed version/document coverage

### DIFF
--- a/scrapers/de/bills.py
+++ b/scrapers/de/bills.py
@@ -5,6 +5,7 @@ from json.decoder import JSONDecodeError
 from requests.exceptions import JSONDecodeError as RequestsJSONDecodeError
 import random
 import time
+from urllib.parse import urljoin
 
 import requests
 from openstates.scrape import Scraper, Bill, VoteEvent
@@ -20,6 +21,8 @@ class FailedBillFetch:
 
 class DEBillScraper(Scraper, LXMLMixin):
     verify = False
+
+    base_url = "https://legis.delaware.gov"
 
     categorizer = Categorizer()
     chamber_codes = {"upper": 1, "lower": 2}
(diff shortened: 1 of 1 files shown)


jessemortenson:
Thanks, @alexkost819 - I'm taking a look. Just want to note, the scrape output (AFTER) zip seems like a very partial scrape, only 36 bills included (https://github.com/alexkost819/openstates-scrapers/releases/download/de-bill-text-test-scrape/de-scrape-after.zip). FWIW it's very helpful to have a full scrape output, that way I don't have to run the scraper myself to do data quality checks.

alexkost819:
> Thanks, @alexkost819 - I'm taking a look. Just want to note, the scrape output (AFTER) zip seems like a very partial scrape, only 36 bills included ([alexkost819/openstates-scrapers@`de-bill-text-test-scrape` de-scrape-after.zip (download)](https://github.com/alexkost819/openstates-scrapers/releases/download/de-bill-text-test-scrape/de-scrape-after.zip)). FWIW it's very helpful to have a full scrape output, that way I don't have to run the scraper myself to do data quality checks.

Gotcha @jessemortenson , sorry, I was trying to be more narrow and easier for review, but I hear you and will provide full scrapes moving forward. I'll do it for this PR too upon request / if I have a moment on the side to do it.

jessemortenson:
I hear where you're coming from, I think that intention is helpful and a good approach for most of the iterations you're doing. The tricky part is the risk of some unintended negative effect on the part of the data that isn't being sampled. I've definitely seen cases of regression that wouldn't be caught on a single sample.

However, i think this code looks good to me, so I'm merging in.

FYI @harshitasingh360 since you're also working on a DE issue

PR #5773, opened 2026-08-25
Data 5715/de missing bills fix

harshitasingh360:
Some Delaware bills show no action history because the state's actions API returns no data for them, even though the bill detail page clearly shows an introduction date and current committee status. This change adds a fallback that pulls those data points from the bill detail page and records them as actions whenever the API comes back empty, ensuring users always see at least the bill's introduction and current status.

Output - 
[de_25_08_26.zip](https://github.com/user-attachments/files/314
Updated output - 
[de_26_08_26.zip](https://github.com/user-attachments/files/31479550/de_26_08_26.zip)
15606/de_25_08_26.zip)

### DIFF
--- a/scrapers/de/bills.py
+++ b/scrapers/de/bills.py
@@ -355,6 +355,7 @@ class DEBillScraper(Scraper, LXMLMixin):
             allow_redirects=True,
             verify=False,
             headers=self.headers,
+            timeout=self.timeout,
         ).content
 
         page = json.loads(page)
@@ -381,6 +382,7 @@ class DEBillScraper(Scraper, LXMLMixin):
             allow_redirects=True,
             verify=False,
             headers=self.headers,
+            timeout=self.timeout,
         )
         page = self.decode_and_retry_request(
             "scrape_votes", request_method, retries=1, raise_exception=False
@@ -403,6 +405,7 @@ class DEBillScraper(Scraper, LXMLMixin):
             allow_redirects=True,
             verify=False,
             headers=self.headers,
+            timeout=self.timeout,
         )
         page = self.decode_and_retry_request(
             "scrape_vote", request_method, retries=1, raise_exception=False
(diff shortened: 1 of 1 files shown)


scrapers/de/bills.py
```diff
@@ -548,6 +553,83 @@ def scrape_actions(self, bill, legislation_id):
...
+                    if "Senate" in description:
+                        action_chamber = "upper"
+                    elif "House" in description:
+                        action_chamber = "lower"
+                    elif "Governor" in description:
+                        action_chamber = "executive"
+                    else:
+                        action_chamber = home_chamber
+
+                    categorization = self.categorizer.categorize(description)
```
jessemortenson, line 617: I think we need one more tweak here. It's great that you're calling the categorizer, but only one of two actions is actually getting the classification/categorization that it should. This code produces:

- "Introduced" -> classification set correctly to `introduction`
- "House Natural Resources & Energy" -> classification is empty

Since the source website says that "status' is "House Natural Resources & Energy 6/30/26", I think that action should be classified as `referral-committee`. I believe it indicates that the bill has been referred to the House (lower) Natural Resources & Energy committee.

SO, either the action description text needs to match what the classifier is expecting (`["assigned to (?P<committees>.+?) Committee in"]` from actions.py)... OR a new rule should be added to the classifier. 

I think it's probably simpler, since this is a fallback to handle an edge case, to go with modifying the description text to match, so something like:

"Assigned to Natural Resources & Energy Committee in House"

jessemortenson (changes requested):
This is looking good, but needs one more tweak

scrapers/de/bills.py
harshitasingh360: That makes sense Jesse, pushing in the changes for this fix.

jessemortenson (approved):
LGTM! Merging in

PR #5774, opened 2026-08-26
VA: fix bill text extraction for 2025+ sessions

### LINKED ISSUE #1395: VA: bill text extraction returns is_error for all recent sessions after the LIS relaunch (2025+)
### Summary

Bill text extraction for Virginia appears to have stopped working at the 2024/2025 session boundary. `opencivicdata_searchablebill` rows for recent Virginia sessions carry `is_error = true` with empty `raw_text`, where 2024 rows extracted normally.

### Measurement

Sampling 300 random Virginia bills per session from the bulk dump, matched by `bill_id`:

| session | rows found | `is_error = true` | usable `raw_text` |
|---|---|---|---|
| VA 2024 | 300 | 15 | 285 |
| VA 2025 | 300 | **300** | **0** |

Spot check on a specific bill (VA HB 2094, 2025 — the high-risk AI bill):

```
ocd-bill/4776e188-f65f-46a7-80d9-586e1cddf0cd   is_error = true    length(raw_text) = 0
ocd-bill/bdc5f164-9346-4f50-a47c-908b91c6a3b2   is_error = false   length(raw_text) = 13764   (HB 802, 2024)
```

Sessions 2026, 2026S1 and 2027 look the same as 2025 from our side.

### Likely cause

Virginia relaunched LIS, and the version URL format changed at exactly that boundary:

- **2023-2024** — `https://lis.virginia.gov/cgi-bin/legp604.exe?241+ful+HB802` (100% of rows)
- **2025+** — `https://lis.blob.core.windows.net/files/1076496.PDF` (~80% PDF, ~8% HTML, remainder null)

Two things that may matter:

1. **The new URLs are healthy.** Fetching one directly returns `HTTP 200`, `Content-Type: application/pdf`, and a valid PDF — so this does not look like a fetch/blocking problem. We ran a PDF text extractor over one and got clean output (`2025 SESSION / HOUSE RESOLUTION NO. 766 / Commending the Timberlake Tavern`, ~2.4k characters).

2. **`media_type` disagrees with the content.** In the dump, Virginia's version rows report `text/html` for *every* row, including the PDFs. If extraction dispatches on the declared media type, PDFs would be routed to an HTML parser and yield nothing — which would match the observed `is_error` pattern. This may be the whole story, or may be a second bug sitting behind the first.

### Impact for downstream users

The bills themselves are present and correct — identifiers, titles, sponsors, actions and statuses all look right, so the `csv_bills` scraper is clearly working. It is specifically the body text that is missing, which means full-text search silently returns nothing for recent Virginia legislation rather than erroring.

Happy to provide more samples or test a fix against our mirror if that would help.

### LINKED ISSUE #5763: VA: two sessions marked active, neither scrapeable (2027 not yet started; 2026S1 ended)
### Summary

`scrapers/va/__init__.py` currently marks **two** sessions `active: true`, and neither can be scraped:

| identifier | name | active | start_date | end_date |
|---|---|---|---|---|
| `2026` | 2026 Regular Session | **false** | 2026-01-14 | 2026-03-14 |
| `2026S1` | 2026, 1st Special Session | **true** | 2026-04-23 | 2026-05-01 |
| `2027` | 2027 Regular Session | **true** | 2027-01-13 | 2027-02-27 |

- `2027` has not started — it begins five months from now. Scraping it yields nothing.
- `2026S1` concluded on 2026-05-01. Scraping it raises `ScrapeError: no objects returned from VaBillScraper scrape`, which is a *correct* scrape of a finished session.
- The session with VA's actual current corpus, **`2026` Regular**, is marked `active: false`.

So any resolver that picks "an active session" gets one of the two unscrapeable ones, and the observable symptom is a scraper that looks broken. It isn't — it faithfully reports that a future session has no bills.

### Suggested fix

Either mark the 2027 session `active: false` until it convenes, or reflect the actual current session. Happy to send a PR for whichever you prefer — I didn't want to guess your convention for how far ahead a session gets flipped to active.

### Note on a fix we tried and rejected

The obvious consumer-side workaround is "prefer active sessions that have started and not ended." We tested that across every jurisdiction in the repo before adopting it: **it changes the resolved session for 41 of ~50 jurisdictions, in every case to none**, because many `__init__.py` files carry an `end_date` in the past on the session that is genuinely current. Sharing that because it looks obviously correct until you measure it.

Also minor, same file: `classification` is `null` on every VA session, where most states set `"primary"`. Not a bug on its own, but consumers that filter regular sessions on `classification == 'primary'` will match nothing for VA.

alexkost819:
## Summary

- Fixes VA bill-text extraction, broken since the LIS relaunch (2025+ sessions).
- Fixes VA's active-session flags so default scrapes hit the real current session.

### `scrapers/va/bills.py`

`getlegislationtextbyidasync` now returns versions that can carry **multiple** PDF/HTML files (new `lis.blob.core.windows.net` storage), which hit an unhandled `raise Exception`. Fixed by looping over all files instead of assuming at most one.

Root-cause note: live-testing against the real 2025 session showed this exception never actually fires across the full session (before/after scrapes were byte-identical for it). It's kept as a real defensive fix, but it wasn't the cause of the reported symptom.

The actual cause of the reported "402 bills, zero recorded version" symptom: some versions (mostly still-prefiled bills) have `PDFFile` and `HTMLFile` both `null` on the new API -- no document was ever generated, only inline `DraftText` in the JSON response, which isn't a URL we can hand to `add_version_link`. VA still runs `legacylis.virginia.gov`, its pre-relaunch site, which server-renders the same text as HTML. Added a fallback to that URL for versions with no file. Verified live against real bill data (see test plan).

### `docker-compose.yml`

Swapped the unused `VIRGINIA_FTP_USER`/`VIRGINIA_FTP_PASSWORD` env vars for `VA_API_KEY`, which is what `bills.py` actually reads.

Closes openstates/issues#1395
Closes #5763

## Test plan

- Standalone script mocking `getlegislationtextbyidasync` responses confirms: (1) a multi-file version row no longer raises and captures all links, (2) a no-file version row falls back to the legacylis URL.
- Ran the full VA 2025 session live against `scrapers/va/bills.py`, before and after the fix (VA_API_KEY required, chunked via the scraper's built-in `scrape_chunk_number` to survive a flaky connection to `lis.virginia.gov` from CI-like sandboxes):
  - **Before:** 3537 bills, 402 with zero recorded versions -- matches the reported symptom exactly, including all 8 example bills (SB 284, SJ 10, SB 230, SB 492, SB 633, HB 1311, SB 251, HB 658).
  - **After:** same 3537 bills, same 9306 vote events, **0** bills with zero versions. All 8 example bills now have a version pointing at a working `legacylis.virginia.gov` URL.
- Before/after scrape output zips: 
[va-2025-before.zip](https://github.com/user-attachments/files/31445642/va-2025-before.zip)
[va-2025-after.zip](https://github.com/user-attachments/files/31445641/va-2025-after.zip)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,8 +25,7 @@ services:
       - NEW_YORK_API_KEY
       - INDIANA_API_KEY
       - DC_API_KEY
-      - VIRGINIA_FTP_USER
-      - VIRGINIA_FTP_PASSWORD
+      - VA_API_KEY
       - S3_REALTIME_BASE=s3://openstates-realtime-bills
       - STATS_ENABLED
       - STATS_RETRIES=1
--- a/scrapers/va/bills.py
+++ b/scrapers/va/bills.py
@@ -5,7 +5,6 @@ import json
 import lxml
 import os
 import pytz
-import requests
 import urllib3
 
 from openstates.scrape import Scraper, Bill, VoteEvent
@@ -65,7 +64,7 @@ class VaBillScraper(Scraper):
         # If we don't IncludeFailed, we will only get a subset of legislation
         body = {"SessionCode": self.session_code, "IncludeFailed": True}
 
-        page = requests.post(
+        page = self.post(
             f"{self.base_url}/Legislation/api/getlegislationlistasync",
             headers=self.headers,
             json=body,
@@ -135,7 +134,7 @@ class VaBillScraper(Scraper):
             "legislationID": legislation_id,
         }
 
-        page = requests.get(
+        page = self.get(
             f"{self.base_url}/LegislationEvent/api/getlegislationeventbylegislationidasync",
             params=body,
             headers=self.headers,
@@ -200,7 +199,7 @@ class VaBillScraper(Scraper):
             bill.add_related_bill(bill.identifier, f"{prior_session}", "prior-session")
 
     def add_sponsors(self, bill: Bill, legislation_id: str):
-        page = requests.get(
+        page = self.get(
             f"{self.base_url}/LegislationPatron/api/GetLegislationPatronsByIdAsync/{legislation_id}",
             headers=self.headers,
             verify=False,
@@ -221,7 +220,7 @@ class VaBillScraper(Scraper):
             "sessionCode": self.session_code,
             "legislationID": legislation_id,
         }
-        response = requests.get(
+        response = self.get(
             f"{self.base_url}/LegislationText/api/getlegislationtextbyidasync",
             params=body,
             headers=self.headers,
@@ -293,7 +292,7 @@ class VaBillScraper(Scraper):
         }
 
         vote_page_url = f"{self.base_url}/Vote/api/getvotebyidasync"
-        page = requests.get(
+        page = self.get(
             vote_page_url,
             params=body,
             headers=self.headers,
--- a/scrapers/va/bills.py
+++ b/scrapers/va/bills.py
@@ -234,25 +234,42 @@ class VaBillScraper(Scraper):
         page = response.json()
 
         for row in page["TextsList"]:
-            if (row["PDFFile"] and len(row["PDFFile"]) > 1) or (
-                row["HTMLFile"] and len(row["HTMLFile"]) > 1
-            ):
-                self.error(json.dumps(row))
-                self.error("Add code to handle multiple files to VA Scraper")
-                raise Exception
-
-            if row["PDFFile"] and len(row["PDFFile"]) > 0:
+            # Since the 2025 LIS relaunch, a version can carry multiple
+            # PDF/HTML files (e.g. lis.blob.core.windows.net/files/*.PDF).
+            # Add a version link for each one instead of just the first.
+            for pdf in row["PDFFile"] or []:
                 bill.add_version_link(
                     row["Description"],
-                    row["PDFFile"][0]["FileURL"],
+                    pdf["FileURL"],
                     media_type="application/pdf",
+                    on_duplicate="ignore",
                 )
 
-            if row["HTMLFile"] and len(row["HTMLFile"]) > 0:
+            for html in row["HTMLFile"] or []:
                 bill.add_version_link(
                     row["Description"],
-                    row["HTMLFile"][0]["FileURL"],
+                    html["FileURL"],
                     media_type="text/html",
+                    on_duplicate="ignore",
+                )
+
+            # Some versions (commonly ones that never got a generated PDF,
+            # e.g. still-prefiled bills) have no PDFFile/HTMLFile at all --
+            # getlegislationtextbyidasync only returns their text inline as
+            # DraftText, which isn't a URL we can hand off. The old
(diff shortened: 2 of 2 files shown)


jessemortenson:
I don't have time to evaluate all of this, but I for sure can't merge in the __init__ changes. I'm not sure I understand this:

> 2026S1 (ended 2026-05-01) and 2027 (doesn't convene until 2027-01-13) were both marked active: true, while 2026 Regular -- the actual current session -- was active: false. Any resolver picking "the active session" got an unscrapeable one. Fixed: 2026 is now active: true, the other two false.

VA does have bills posted for 2027 (prefiles), so that session has to be active so we can pick those bills up when/if more get posted before the session actually starts. There are a lot of jurisdictions that do prefiles, so we are zealous about trying to ingest those as soon as they're available.

alexkost819:
> I don't have time to evaluate all of this, but I for sure can't merge in the **init** changes. I'm not sure I understand this:
> 
> > 2026S1 (ended 2026-05-01) and 2027 (doesn't convene until 2027-01-13) were both marked active: true, while 2026 Regular -- the actual current session -- was active: false. Any resolver picking "the active session" got an unscrapeable one. Fixed: 2026 is now active: true, the other two false.
> 
> VA does have bills posted for 2027 (prefiles), so that session has to be active so we can pick those bills up when/if more get posted before the session actually starts. There are a lot of jurisdictions that do prefiles, so we are zealous about trying to ingest those as soon as they're available.

Understood - I was trying to flip some bits to fix the previous sessions but it should not have been committed. I will manually scrape all the currently-false sessions since 2025 and paste the zip files here. The hope is somebody with privileges can fixup the existing data to include the extracted bill text.

EDIT:

Verified by running a full local scrape of the 2026 Regular session end-to-end: 3956 bills, 12374 vote events, zero errors, ~13hr runtime (throttled RPM). Confirms the version-link fix works against live data.

[va_2026_scrape.zip](https://github.com/user-attachments/files/31773503/va_2026_scrape.zip)

alexkost819:
Update: dropped the `__init__.py` session-flag change entirely — you're right that `active` means "keep polling," not "current session," and VA's prefile behavior means 2026S1/2027 need to stay active regardless of calendar dates. That file now matches `main` exactly; this PR only touches `bills.py` (version-link extraction) and `docker-compose.yml` (env var rename).

Also found and fixed a real bug while verifying: `bills.py` was calling `requests.get()`/`requests.post()` directly for all VA API calls instead of the scraper's `self.get()`/`self.post()`, which bypasses scrapelib's built-in retry/backoff. Any transient connection reset from `lis.virginia.gov` (which happens somewhat often) killed the whole scrape instead of retrying. Switched all 5 call sites to `self.get`/`self.post`.

Verified by running a full local scrape of the 2026 Regular session end-to-end: 3956 bills, 12374 vote events, zero errors, ~13hr runtime (throttled RPM). Confirms the version-link fix works against live data and the retry fix survives connection resets that killed earlier attempts.

jessemortenson:
Looks good to me, merging

PR #5792, opened 2026-09-02
MA: include not-voting total in declared vote counts

### LINKED ISSUE #5791: MA: declared vote counts never include not-voting/other total
## Summary

In `scrapers/ma/votes.py`, the `HouseVoteRecordParser` class parses each House roll-call's declared totals from the journal PDF header — e.g. `148 YEAS 4 NAYS 7 N/V` — but only the yea and nay totals are ever applied to the resulting `VoteEvent`. The not-voting total is parsed into `self.total_nv` and then silently dropped.

## Mechanism

- `total_nv_re = re.compile(r"(\d+) n/v", re.IGNORECASE)` (line 359) correctly matches the declared not-voting total in the PDF text and is stored in `self.total_nv` (line 419).
- `createVoteEvent()` (around line 488) calls:
  ```python
  vote.set_count("yes", self.total_yea)
  vote.set_count("no", self.total_nay)
  ```
  but never calls `vote.set_count("not voting", self.total_nv)`.
- Meanwhile, individual not-voting legislators ARE itemized in the roll: `vote_dictionary["X"] = "not voting"` and `vote.vote(vote_dictionary[vote_val], name)` (around line 500) records each "X" vote as `"not voting"`.
- `VoteEvent.set_count()` only appends/updates an entry in `counts` when it's explicitly called — there's no default zero-fill for unused options — so without a call for `"not voting"`, that key is simply absent from `counts` entirely, not even present as 0.

The net effect: the declared `counts` array on the `VoteEvent` never has a `"not voting"` entry, while the itemized roll (`votes` array) can have several members recorded as not voting — an internal inconsistency in the scraped data.

## Real example (live data)

Live source: [House Journal Roll Calls, 2025 (194th session)](https://malegislature.gov/Journal/House/194/2025/RollCalls) — page 57 of that PDF reads:

```
H. 4227 On enactment  Yea and Nay  06/18/2025 02:44 PM
No. 57  148 YEAS  4 NAYS  7 N/V
```

7 legislators (Murray, Nguyen, O'Day, Rogers D., Sousa, Wells, Pease) are individually marked "X" (not voting) in that roll.

**Scraper output before this fix** (`counts` has no `"not voting"` entry at all):
```json
"counts": [
  {"option": "yes", "value": 148},
  {"option": "no", "value": 4}
]
```

**Scraper output after this fix**:
```json
"counts": [
  {"option": "yes", "value": 148},
  {"option": "no", "value": 4},
  {"option": "not voting", "value": 7}
]
```

`7` now matches the count of `"not voting"` entries actually itemized in that same vote's `votes` array.

## Fix

A one-line fix adds `vote.set_count("not voting", self.total_nv)` alongside the existing yes/no count calls, using the same `"not voting"` label already used by `vote.vote()` so the declared count and itemized roll stay consistent.

Verified against a live scrape of the current (194th) session: of 323 vote events scraped, 212 have not-voting members, and after the fix the declared not-voting count matches the itemized roll count in all 212 cases (0 mismatches).

alexkost819:
## Summary

`HouseVoteRecordParser` in `scrapers/ma/votes.py` already parses the House journal's declared not-voting total (e.g. `7 N/V`) into `self.total_nv`, but `createVoteEvent()` never applied it via `vote.set_count()` — only yes/no counts were set. Meanwhile individual not-voting legislators are itemized in the roll as `"not voting"`. Result: the declared `counts` on the VoteEvent always showed 0 for not-voting even when the itemized roll had several.

Fix is a one-line addition:

```python
vote.set_count("not voting", self.total_nv)
```

using the same `"not voting"` label the roll parsing already uses, so the declared count and itemized roll stay consistent.

Fixes #5791

## Verification

Ran a full live scrape of MA votes for the 194th (current) session: `os-update ma votes session=194th --scrape`. Result: 323 vote events scraped; 212 have not-voting members; declared not-voting counts matched the itemized roll count in all 212 cases (0 mismatches).

Verification output (scraped JSON) is attached here: https://github.com/alexkost819/openstates-scrapers/releases/download/ma-vote-fix-verify-bcf2c8745/ma-vote-fix-verification.zip

Done with Claude Code

### DIFF
--- a/scrapers/ma/votes.py
+++ b/scrapers/ma/votes.py
@@ -487,6 +487,7 @@ class HouseVoteRecordParser:
 
         vote.set_count("yes", self.total_yea)
         vote.set_count("no", self.total_nay)
+        vote.set_count("not voting", self.total_nv)
 
         vote_dictionary = {
             "Y": "yes",


jessemortenson (approved):
LGTM, merging

PR #5796, opened 2026-09-02
NH: Narrow governor signed

showerst:
NH was incorrectly classifying some bills as signed, such as https://gc.nh.gov/bill_status/legacy/bs2016/bill_docket.aspx?lsr=2818&sy=2026&sortoption=&txtsessionyear=2026 because they had an action like "Conference Committee Report; Not Signed Off".

@jessemortenson h/u this may require a manual data correction.


jessemortenson:
Thanks for the heads up, Tim

PR #5798, opened 2026-09-02
KY: fix vote count mismatches in bill vote PDF parsing

### LINKED ISSUE #5794: KY: vote count mismatch from fragile line-count heuristic in PDF parsing
## Summary

`scrape_votes()` in `scrapers/ky/bills.py` (around lines 383-391 on `main`) parses KY roll-call vote PDFs and can produce a mismatch between the declared vote counts (from the PDF's `YEAS:`/`NAYS:`/`ABSTAINED:`/`NOT VOTING:` header numbers) and the itemized per-legislator vote list actually recorded.

## Mechanism

The `NOT VOTING` section harvests names using:

```python
lines_to_go_through = math.ceil(not_voting / len(line.split()))
```

This divides the declared `not_voting` count by the **token count of the header line itself** (e.g. `"NOT VOTING : 5"` → 4 tokens), not by how many names actually appear per line in the names block below it (which is a fixed 5-column layout, matching the YEAS/NAYS sections). Since 4 ≠ 5, this guess is wrong whenever `not_voting` doesn't happen to divide evenly against a stride of 4, and it over-harvests an extra line — which is often the next page's header text or blank content, producing a bogus extra "voter" entry.

A related, distinct off-by-one bug lives in the same function's `ABSTAINED`/`PASSES` section-parsing loop: a hand-rolled lookahead re-reads the first names line of that section twice, double-counting those voters whenever the section has a blank line between its header and its first names line.

## Real example

Source PDF (live, still fetchable): https://apps.legislature.ky.gov/record/25rs/hb90/vote_history.pdf — the Senate vote on `PASS HB 90 W/ scs1 scta1` (RSN# 3506, 3/13/2025). Its actual text layout is:

```
NAYS : 0

PASSES : 5

Berg                        Herron                          Thomas             Yates
ChambersArmstrong

NOT VOTING : 4

Carpenter                   Neal                            Smith              Webb
```

Declared: `PASSES: 5` + `NOT VOTING: 4` → `other` count should be 9.

**Before the fix**, scraping this real PDF with current `main` produces (trimmed to the relevant fields):

```json
{
  "motion_text": "PASS HB 90 W/ scs1 scta1",
  "counts": [
    {"option": "yes", "value": 29},
    {"option": "no", "value": 0},
    {"option": "other", "value": 9}
  ],
  "other_voters": ["Berg", "Herron", "Thomas", "Yates", "Berg", "Herron", "Thomas", "Yates", "ChambersArmstrong", "Carpenter", "Neal", "Smith", "Webb"]
}
```

Declared `other: 9` but 13 people itemized — the ABSTAINED/PASSES off-by-one double-reads `Berg, Herron, Thomas, Yates` (the first line of the `PASSES` block).

**After the fix**, the same PDF produces:

```json
{
  "motion_text": "PASS HB 90 W/ scs1 scta1",
  "counts": [
    {"option": "yes", "value": 29},
    {"option": "no", "value": 0},
    {"option": "other", "value": 9}
  ],
  "other_voters": ["Berg", "Herron", "Thomas", "Yates", "ChambersArmstrong", "Carpenter", "Neal", "Smith", "Webb"]
}
```

9 declared, 9 itemized, no duplicates.

The `NOT VOTING` line-count heuristic bug (the primary one described above) reproduces the same class of mismatch whenever `not_voting` doesn't align with the header's 4-token stride; e.g. `not_voting=5` with the real 5-column layout over-harvests a bogus 6th "voter" pulled from the following page's header text.

## Fix

Fixed in #TBD (PR incoming): replaces both heuristics with straightforward boundary/count-based scans (matching the approach already used for YEAS/NAYS), verified against a full real scrape of KY's 2025 Regular Session House chamber (907 bills, 399 vote events) — declared vote counts now match itemized counts in all 399 votes.

alexkost819:
## Summary

Fixes #5794.

`scrape_votes()` in `scrapers/ky/bills.py` parses KY roll-call vote PDFs and had two bugs causing declared vote counts (from the PDF's header numbers) to not match the itemized per-legislator vote list:

1. The `NOT VOTING` section used `math.ceil(not_voting / len(header_line.split()))` to guess how many lines of names to harvest — decoupled from how many names actually appear per line (a fixed 5-column layout), so it would over-harvest a bogus trailing "name" from the next page/header whenever the guess overshot.
2. The `ABSTAINED`/`PASSES` section had an off-by-one in a hand-rolled lookahead loop that re-read the first names line twice, double-counting those voters.

Both are replaced with straightforward boundary/count-based scans, mirroring the approach already used for the YEAS/NAYS sections in the same file.

## Verification

Ran a full real scrape of KY's 2025 Regular Session, House chamber (`os-update ky bills session=2025RS chamber=lower --scrape`): 907 bills, 399 vote events. Declared vote counts now match itemized per-legislator counts in **all 399** votes (previously mismatched whenever a vote's not-voting/abstained counts happened to trigger either heuristic — see #5794 for a real before/after example).

Verification artifact (full scrape JSON output + the count-comparison script/results): https://github.com/alexkost819/openstates-scrapers/releases/download/ky-vote-fix-verify-6d1a877ae/ky-vote-fix-verification.zip

Also added `scrapers/ky/tests/test_vote_parsing.py`, a regression check against two real vote-history PDFs (fixtures included) that reproduce each bug — fails against the old code, passes against the fix.

## Test plan

- [x] `python3 -m py_compile scrapers/ky/bills.py`
- [x] `flake8` clean
- [x] `python3 -m scrapers.ky.tests.test_vote_parsing` passes
- [x] Full real scrape of 2025RS House chamber: 0 declared/itemized mismatches across 399 vote events (down from mismatches whenever either heuristic misfired)

Done with Claude Code


jessemortenson (approved):
LGTM, merging

PR #5800, opened 2026-09-02
TX: fix garbled voter names and duplicate vote events

### LINKED ISSUE #5795: TX: voter names truncated/split by fragile comma-splitting and trailing-char strip
## Summary

`scrapers/tx/votes.py`'s `names()` function (used to parse Yeas/Nays/Present/Absent
voter lists out of TX House/Senate journal pages) corrupts voter names on short
lists (fewer than 7 semicolon-separated parts).

## Mechanism

1. `scrapers/tx/votes.py` lines ~96-99 (before fix):

   ```python
   split_name_list = text.split(";")
   if len(split_name_list) < 7:
       # probably failed to properly split on semi-colons; try commas:
       split_name_list = text.split(",")
   ```

   TX names legitimately contain a disambiguating comma for shared surnames,
   e.g. `"Davis, A."`, `"Lopez, J."`, `"Bell, C."`. When a short voter list
   (Absent/Present/Excused, which often has only a handful of names) triggers
   this fallback, the naive `split(",")` shreds these into two fragments:
   `"Davis, A."` -> `"Davis"` + `"A"`.

   Confirmed on a real, live journal page
   (`https://journals.house.texas.gov/HJRNL/89R/HTML/89RDAY65FINAL.HTM`, 2025-05-15):
   the raw paragraph `"Absent — Bowers; Davis, A.; Dean; Simmons."` has only 4
   semicolon-parts, triggers the fallback, and `names()` returned
   `['Bowers', 'Davis', 'A', 'Dean', 'Simmon']` -- both `"Davis, A."` shredded
   into `"Davis"` / `"A"`, *and* `"Simmons"` truncated to `"Simmon"` by bug #2
   below.

2. A related heuristic a few lines further down, used when a name segment still
   contains an embedded label like `"Absent — Name"`:

   ```python
   if " — " in n:
       n = n.split()[-1]
   ```

   This grabs the last *whitespace token*, which breaks the same
   comma-disambiguated names when they immediately follow the em-dash label,
   e.g. `"Absent — Jones, V.; Lambert; Smithee."` -> `names()` returns
   `['V', 'Lambert', 'Smithee']` (should be `"Jones, V"`, not bare `"V"`).
   Confirmed on the same live journal page.

3. Line ~124 (before fix):

   ```python
   # Get rid of trailing '.'
   names[-1] = names[-1][0:-1]
   ```

   This unconditionally chops the last character off the last name in the
   list, assuming it's always a trailing period. When that assumption is
   wrong, it truncates a real character. Confirmed on the same live journal
   page: `"Present, not voting — Mr. Speaker; Vasut(C); Virdell."` ->
   `names()` returned `['Present', 'Speaker', 'Vasut(C)', 'Virdell']` in one
   variant and, before other pieces of this fix, `"Virdell"` -> `"Virdel"`
   and `"Smithee"` -> `"Smithe"` in others -- real Representatives' names
   losing their last letter. Similarly, `"Vasut(C)."` -> `"Vasut(C"` (an
   unbalanced-paren truncation matching the shape of the originally-reported
   `"Harris(C"` garbling).

## Impact

Vote records for TX bills contain garbled/truncated/split voter names instead
of the real legislator names, making it impossible to correctly attribute
individual votes.

## Fix

A fix is included in PR (see linked PR) that:
- Makes the comma-fallback split skip a comma only when it's immediately
  followed by a lone trailing initial (i.e. a genuine `"Surname, Initial."`
  suffix-disambiguation, ending the name), while still splitting on commas
  that are genuine separators (including TX Senate's *leading*-initial
  disambiguation style, e.g. `"Hancock, A. Hinojosa, J. Hinojosa, Huffman"`).
- Changes the em-dash-label strip to split on the literal `" — "` separator
  instead of grabbing the last whitespace token.
- Only strips the last character of the final name when it's actually a
  trailing `"."`.

### LINKED ISSUE #5797: TX: duplicate vote-event records from overlapping vote-paragraph scanners
## Summary

`scrapers/tx/votes.py`'s `votes()` function chains four independent scanners
over the same set of journal paragraphs:

```python
def votes(root, session, chamber):
    for vote in record_votes_with_yeas(root, session, chamber):
        yield vote
    for vote in viva_voce_votes(root, session, chamber):
        yield vote
    for vote in record_votes_with_short_count_notation(root, session, chamber):
        yield vote
    for vote in local_calendar_votes(root, session, chamber):
        yield vote
```

## Mechanism

Each scanner independently `xpath`s over `//div[@class="textpara"]` (or a
subset of it) with its own trigger condition, and these trigger conditions
are not mutually exclusive:

- `record_votes_with_yeas` (~line 592-611 before fix) triggers on any
  `div[@class="textpara"]` whose text contains "yeas" (case-insensitive).
- `record_votes_with_short_count_notation` (~line 476-489 before fix)
  triggers on a paragraph containing a trailing `(N - N - N)` tally, e.g.
  `SB 2479 (... - no) (111 - 27 - 1)`.

A single paragraph that both mentions "Yeas"/"Nays" counts by name (e.g. `...
by (Record 2040): 122 Yeas, 17 Nays, ...`) *and* ends with a `(N - N - N)`
short-count notation would satisfy both scanners' `is_valid` checks and get
turned into two separate `VoteEvent`s for the same underlying vote -- one
from each scanner, both yielded from the same `votes()` call for the same
paragraph.

`local_calendar_votes` and `viva_voce_votes` (also chained in the same loop)
add further overlapping trigger surfaces over the same paragraph set.

## Impact

A single real legislative vote can be recorded twice (or more) as separate
`vote_event` records for the same bill/chamber/date, double-counting votes
and confusing downstream consumers.

Note: `TXVoteScraper.scrape_journal` (in the same file) has a
*content*-based dedupe key (`build_vote_dedupe_key`: bill + chamber + date +
motion_text + counts) that catches some duplicates after the fact, but it
does not catch every case -- e.g. when the two scanners assign different
`motion_text` to what is really the same vote (one from `mv.motion_text`
parsing, one hardcoded to `"passage"`/`"other"`), the content keys differ and
the duplicate survives. The four scanners have no shared awareness of which
journal *paragraph* has already produced a vote, which is the actual root
cause.

Live 89R sampling (House days 40, 50, 55, 60, 65, 70, 75, 80) did not turn up
a paragraph that happens to hit two scanners' triggers at once in that
particular sample, but the overlapping trigger conditions are confirmed at
the code level and are date/content-dependent, so this is a latent
correctness bug rather than one that necessarily fires on every journal page.

## Fix

A fix is included (see linked PR) that has the four scanners in `votes()`
share a "claimed paragraphs" set (keyed by normalized paragraph text),
skipping a paragraph in a later scanner once an earlier scanner has already
turned it into a `VoteEvent`.

alexkost819:
## Summary

- `names()` shredded TX's `"Surname, Initial."` names on short voter lists (`"Morales, C."` -> `"Morales"` + `"C"`), via a naive `text.split(",")` fallback and an `"Absent — "` label strip that grabbed the last whitespace token.
- `names()` also chopped the last character off the last name in every list, whether or not it was a trailing period: `"Paul"` -> `"Pau"`, `"Schofield"` -> `"Schofiel"`, `"Vasut(C)"` -> `"Vasut(C"`. All 213 vote events on 89R House day 65 had at least one garbled name.
- The four vote-paragraph scanners chained in `votes()` have non-mutually-exclusive triggers and can turn one paragraph into several `VoteEvent`s; they now share a set of claimed paragraphs, keyed by document position (journal boilerplate text repeats verbatim across bills, so text is not a usable key).

Fixes #5795. Fixes #5797.

## Bug example

SB 2052, 89R, passage (2025-05-15), House day 65 journal:

```json
{
  "counts": [
    { "option": "yes", "value": 139 },
    { "option": "no", "value": 2 },
    { "option": "not voting", "value": 2 }
  ],
  "votes": [
    { "option": "no", "voter_name": "Morales" },
    { "option": "no", "voter_name": "C" },
    { "option": "not voting", "voter_name": "Speaker" },
    { "option": "not voting", "voter_name": "Vasut(C" },
    { "option": "absent", "voter_name": "Pau" },
    { "option": "absent", "voter_name": "Coo" }
  ]
}
```

## Resolution example

Same vote in the after scrape:

```json
{
  "counts": [
    { "option": "yes", "value": 139 },
    { "option": "no", "value": 2 },
    { "option": "not voting", "value": 2 }
  ],
  "votes": [
    { "option": "no", "voter_name": "Morales, C" },
    { "option": "not voting", "voter_name": "Mr Speaker" },
    { "option": "not voting", "voter_name": "Vasut(C)" },
    { "option": "absent", "voter_name": "Paul" },
    { "option": "absent", "voter_name": "Cook" }
  ]
}
```

## Scrapes

Before and after: https://github.com/alexkost819/openstates-scrapers/releases/tag/scrape-tx-after (`tx-before.zip` / `tx-after.zip`), both from `89RDAY65FINAL.HTM` (House, `chamber=lower url_match=89RDAY65FINAL`) and `89RSJ05-15-F.HTM` (Senate, `chamber=upper url_match=05-15`), 89R.

Diff over the matched vote events:
- House: 213 vote events before and after, all 213 with voter-row changes -- 801 voter rows removed, 731 added.
- Recurring repairs: `Speaker` -> `Mr Speaker` (164), `Schofiel` -> `Schofield` (99), `Vasut(C` -> `Vasut(C)` (83), `Pau` -> `Paul` (79), `Harris(C` -> `Harris(C)` (61), `Virdel` -> `Virdell` (23), plus the comma-initial names (`Davis, A`, `Jones, V`, `Lopez, R`, `Morales, C`) that were split in two.
- Senate: 62 vote events before and after, no rows changed. No vote events added or removed in either chamber -- the paragraph-claiming change is a latent-correctness guard.

### DIFF
--- a/scrapers/tx/votes.py
+++ b/scrapers/tx/votes.py
@@ -95,8 +95,15 @@ def names(el):
     text = (el.text or "") + (el.tail or "")
     split_name_list = text.split(";")
     if len(split_name_list) < 7:
-        # probably failed to properly split on semi-colons; try commas:
-        split_name_list = text.split(",")
+        # probably failed to properly split on semi-colons; try commas.
+        # But TX names legitimately use a comma to disambiguate shared
+        # surnames, e.g. "Davis, A." or "Lopez, J." -- don't split those
+        # apart. Only skip a comma when it's immediately followed by a
+        # lone initial that ends the name (semicolon or end of string);
+        # if a capitalized word follows the initial instead (senate lists
+        # disambiguate with a *leading* initial, e.g. "Hancock, A.
+        # Hinojosa, J. Hinojosa, Huffman"), the comma is a real separator.
+        split_name_list = re.split(r",(?!\s*[A-Z]\.(?:;|$))", text)
 
     names = []
     for name in split_name_list:
@@ -112,7 +119,11 @@ def names(el):
             for n in names_still_with_semicolon:
                 n = n.strip().replace(".", "")
                 if " — " in n:
-                    n = n.split()[-1]
+                    # Take everything after the label (e.g. "Absent —"),
+                    # not just the last whitespace token -- a comma-initial
+                    # name like "Jones, V" has a space in it and would
+                    # otherwise get shredded down to just "V".
+                    n = n.split(" — ")[-1]
                 names.append(n)
         else:
             names.append(name)
@@ -120,8 +131,11 @@ def names(el):
     if names:
         # First item in the list will have stuff to ignore before an mdash
         names[0] = clean_starting_name(names[0]).strip()
-        # Get rid of trailing '.'
-        names[-1] = names[-1][0:-1]
+        # Get rid of trailing '.', if there is one -- don't blindly chop
+        # the last character, or a real trailing character (e.g. the "l"
+        # in "Virdell") gets truncated.
+        if names[-1].endswith("."):
+            names[-1] = names[-1][:-1]
 
     return names
 
@@ -142,14 +156,28 @@ def clean_starting_name(name):
     return re.split(r"[\u2014:]", name)[-1]
 
 
+def paragraph_key(el):
+    """A stable identity for a journal paragraph, used to make sure the
+    four vote scanners below (whose trigger conditions aren't mutually
+    exclusive -- e.g. a paragraph can contain both "yeas"-style text and
+    a trailing "(N - N - N)" tally) don't each turn the same paragraph
+    into its own VoteEvent."""
+    return " ".join(el.text_content().split())
+
+
 def votes(root, session, chamber):
-    for vote in record_votes_with_yeas(root, session, chamber):
+    # Shared across all four scanners so a paragraph already claimed by
+    # an earlier one is skipped by the rest.
+    claimed_paragraphs = set()
+    for vote in record_votes_with_yeas(root, session, chamber, claimed_paragraphs):
         yield vote
-    for vote in viva_voce_votes(root, session, chamber):
+    for vote in viva_voce_votes(root, session, chamber, claimed_paragraphs):
         yield vote
-    for vote in record_votes_with_short_count_notation(root, session, chamber):
+    for vote in record_votes_with_short_count_notation(
+        root, session, chamber, claimed_paragraphs
+    ):
         yield vote
-    for vote in local_calendar_votes(root, session, chamber):
+    for vote in local_calendar_votes(root, session, chamber, claimed_paragraphs):
         yield vote
 
 
@@ -473,7 +501,7 @@ short_count_notation_regex = re.compile(
 )
 
 
-def record_votes_with_short_count_notation(root, session, chamber):
+def record_votes_with_short_count_notation(root, session, chamber, claimed_paragraphs):
     # votes with short vote count notation may look like:
     # SB 422 (Cook, Patterson, and Thimesch - no) (135 - 3 - 1)
     # or
@@ -495,6 +523,10 @@ def record_votes_with_short_count_notation(root, session, chamber):
         if not mv.is_valid:
             continue
 
+        key = paragraph_key(mv.el)
+        if key in claimed_paragraphs:
+            continue
+
         if mv.passed:
             motion_text = "passage"
         else:
@@ -510,6 +542,7 @@ def record_votes_with_short_count_notation(root, session, chamber):
             bill=mv.bill_id,
             bill_chamber=mv.chamber,
         )
+        claimed_paragraphs.add(key)
 
         v.set_count("yes", mv.counts["yeas"] or 0)
         v.set_count("no", mv.counts["nays"] or 0)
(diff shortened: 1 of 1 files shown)


scrapers/tx/votes.py
```diff
@@ -888,3 +936,82 @@ def get_session_start_date_string(self, session) -> str:
...
+        record_votes_with_short_count_notation(root, "89R", "lower", claimed)
+    )
+    assert len(yeas_votes) == 1
+    assert len(short_count_votes) == 0, "second scanner should skip claimed paragraph"
+
+    print("tx/votes.py self-check OK")
+
+
+if __name__ == "__main__":
+    _self_check()
```
jessemortenson, line 1017: What's going on with this? I don't know that I've seen this pattern before. When would this `_self_check()` code be run?

alexkost819: i'll remove it, i understand the intent here but the execution isn't great. would probably be good to have some sort of UTs in this repo for some sanity checks on merge, and this was that, but again, poorly executed

alexkost819: removed

jessemortenson (approved):
LGTM, merging

PR #5802, opened 2026-09-02
MN: re-enable Senate vote extraction in bills.py

### LINKED ISSUE #5801: MN: Senate votes not scraped since Feb 2024 (extract_vote_from_action disabled)
## Summary

MN Senate floor votes have not been scraped since Feb 2024. `scrapers/mn/bills.py`'s `process_actions_table` (around line 592-601, pre-fix) has the only call site of `extract_vote_from_action` — the Senate vote extractor — commented out:

```python
# TODO: add votes back in once not broken
# Try to extract vote
# senate vote only, house votes are scraped by the vote_event.py scraper
# if current_chamber == "upper":
#     vote = self.extract_vote_from_action(
#         bill, bill_action, current_chamber, row, pages
#     )
#     if vote:
#         votes.append(vote)
```

This was disabled in [`b991c3595`](https://github.com/openstates/openstates-scrapers/commit/b991c3595165cf37edfa5c385cf2e9801afc9603) / PR [#4838](https://github.com/openstates/openstates-scrapers/pull/4838) ("MN: temp remove senate votes"), whose body says: "Scrapes have been failing with bad vote links, removing so can get the bill data online while I work on a real fix." No follow-up ever re-enabled it. `scrapers/mn/vote_events.py` only covers the House (its own file header says so), so the Senate has produced zero `VoteEvent`s for over a year and a half.

## Live verification

Testing `extract_vote_from_action` directly against the live site (SF202, 2025-2026 session, https://www.revisor.mn.gov/bills/bill.php?b=Senate&f=SF202&ssn=0&y=2025) shows the underlying scrape/parse logic itself works fine today against the current site markup — a 62-0 "Third reading Passed" vote parses correctly, with full roll-call data pulled from the Senate journal PDF (https://www.senate.mn/journals/2025-2026/2025021004.pdf#31). Six more real votes across three other enacted bills this session (SF856, SF1075, SF1251), including a failed 19-45 floor motion with no direct page link (exercising the `pages[-1]` fallback), all parsed correctly with no exceptions.

The real risk described in the original PR is that a single malformed/missing vote link (bad query string, empty `pages` list, unexpected link shape) throws an unhandled exception and kills the *entire* bill scrape for that chamber — not just skips that one vote.

## Fix

- Re-enable the Senate `extract_vote_from_action` call site.
- Wrap the call in `try/except`, logging a warning and skipping just that vote on failure, instead of letting one bad link abort the whole scrape (this directly targets the failure mode described in #4838).
- Fix a latent bug in `extract_vote_from_action`'s no-page-link fallback: it returned `[None] * 2` instead of `None` when no page URL was resolvable at all, which would have been treated as a truthy (bogus) vote by the caller.

## Before / after (SF202, 2025-2026 session)

Before: `bills.py` yields zero `VoteEvent` objects for any Senate bill.

After (real scrape output, `scrapers/mn/bills.py` with `extract_vote_from_action` re-enabled):

```json
{
  "motion_text": "Third reading Passed",
  "start_date": "2025-02-10",
  "classification": ["passage"],
  "result": "pass",
  "bill_identifier": "SF 202",
  "bill_chamber": "upper",
  "counts": [{"option": "yes", "value": 62}, {"option": "no", "value": 0}],
  "sources": [{"url": "https://www.senate.mn/journals/2025-2026/2025021004.pdf#31", "note": ""}],
  "votes": "[62 named 'yes' voters, e.g. Abeler, Bahr, Boldon, ... Xiong; 0 'no' voters]"
}
```

(full output attached as a release asset zip on the linked PR)

alexkost819:
## Summary

MN Senate floor votes haven't been scraped since Feb 2024 (PR #4838 / commit b991c3595 disabled the only call site of `extract_vote_from_action` in `scrapers/mn/bills.py` after "bad vote links" were crashing the whole bill scrape, with a `# TODO: add votes back in once not broken` left in place). `scrapers/mn/vote_events.py` only ever covered the House, so the Senate side has produced zero `VoteEvent`s the entire time.

I re-tested `extract_vote_from_action` directly against the live revisor.mn.gov/senate.mn site for several enacted 2025-2026 bills (SF202, SF856, SF1075, SF1251) and the parsing logic itself works correctly against the current site markup — the "bad vote links" failure mode from 2024 no longer reproduces with today's page structure. So the fix here is:

- Re-enable the Senate-only call to `extract_vote_from_action` in `process_actions_table`.
- Wrap that call in `try/except`, logging a warning and skipping just that one vote on failure, so a single malformed/unexpected vote link can never again take down the whole bill scrape (this is the actual root-cause fix for the original failure mode, not just "uncomment and hope").
- Fix a latent bug in the no-page-link fallback path: it returned `[None] * 2` instead of `None` when no page URL could be resolved, which the caller's `if vote:` check would have treated as a truthy (bogus) vote object.

## Why this fix, not a rewrite

`extract_vote_from_action` already does the hard part correctly (resolves the Senate journal page via `senate.mn`'s API, downloads and parses the journal PDF with PyMuPDF, extracts yes/no roll calls) — I verified it end-to-end against six real votes this session including a failed 19-45 floor motion with a missing page link (the `pages[-1]` fallback path). Nothing here needed a rewrite; the dedicated House scraper (`vote_events.py`) is untouched and there's no overlap risk since it's chamber-gated to the House only.

## Verification

Ran the real `MNBillScraper.scrape()` for Senate bills (2025-2026 session, bill numbers 1-300) with the fix applied. Confirmed multiple real `VoteEvent`s now come through with correct chamber/date/counts/roll-call data, no exceptions, and no overlap with House votes (produced by a completely separate, chamber-gated scraper).

Full scrape output and additional per-bill verification: https://github.com/alexkost819/openstates-scrapers/releases/download/mn-vote-fix-verify-6d38286c6/mn-vote-fix-verification.zip

Fixes #5801

Done with Claude Code

### DIFF
--- a/scrapers/mn/bills.py
+++ b/scrapers/mn/bills.py
@@ -590,15 +590,24 @@ class MNBillScraper(Scraper, LXMLMixin):
             elif len(page_links) > 0:
                 pages.append(page_links[0])
 
-            # TODO: add votes back in once not broken
             # Try to extract vote
             # senate vote only, house votes are scraped by the vote_event.py scraper
-            # if current_chamber == "upper":
-            #     vote = self.extract_vote_from_action(
-            #         bill, bill_action, current_chamber, row, pages
-            #     )
-            #     if vote:
-            #         votes.append(vote)
+            if current_chamber == "upper":
+                try:
+                    vote = self.extract_vote_from_action(
+                        bill, bill_action, current_chamber, row, pages
+                    )
+                except Exception as e:
+                    # Senate journal vote links/pages are occasionally malformed
+                    # or point at content that doesn't parse as expected. Don't let
+                    # one bad vote link take down the whole bill scrape for it.
+                    self.warning(
+                        f"{bill.identifier}: failed to extract Senate vote from "
+                        f"action {bill_action['action_text']!r}: {e}"
+                    )
+                    vote = None
+                if vote:
+                    votes.append(vote)
 
         return bill_actions, votes
 
@@ -820,7 +829,7 @@ class MNBillScraper(Scraper, LXMLMixin):
                         f"No vote url in the page. trying to use the previous vote url: {vote_url}"
                     )
                     if not vote_url:
-                        return [None] * 2
+                        return None
                 else:
                     vote_url = vote_element.xpath(".//a[@href]/@href")[0]
 


jessemortenson (approved):
LGTM, merging

PR #5805, opened 2026-09-03
TX Events scraper fix for ScrapeValueError

harshitasingh360:
**Summary:** TX events scraper fails validation on empty agenda item descriptions
**Description:**
The TX events scraper was raising ScrapeValueError: '' is too short when an agenda item's text consisted only of punctuation (e.g. ":", "-"). After rstrip(" :"), these became empty strings, which fail the openstates schema's minLength: 1 requirement on agenda[].description and abort the whole scrape.
**Fix:** Skip agenda items whose description is empty after stripping trailing punctuation/whitespace, and fix the dedup check so repeated agenda text is properly filtered.
**Output -** 
[tx_Sept02.zip](https://github.com/user-attachments/files/31766793/tx_Sept02.zip)

### DIFF
--- a/scrapers/tx/events.py
+++ b/scrapers/tx/events.py
@@ -136,7 +136,12 @@ class TXEventScraper(Scraper, LXMLMixin):
             for span in p_element.xpath(xpath):
                 text = " ".join("".join(span.itertext()).split())
                 if text and text not in results:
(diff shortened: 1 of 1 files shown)


sinchana00 (commented):
The logic looks good to me.

PR #5806, opened 2026-09-03
OPEN-252: fold the bill identifier into MA vote dedupe_key, not just PDF+supplement

agent-smith-ddp:
## Summary

- Root cause of the `DuplicateItemError` that killed MA's RDS load during OPEN-193's canary (run `ma-f08d7646b9fe`): a single House roll-call PDF supplement number can legitimately be cited by more than one bill's action history — MA sometimes engrosses companion/consolidated bills together in one roll call.
- Confirmed live (not guessed) by pulling the actual scraped manifest from S3 and grouping all 504 `vote_event_*.json` files by `dedupe_key`: **Supplement #29 on 2025-04-09 is cited by both H4005's and H4010's "Passed to be engrossed" actions**, and both computed the identical `dedupe_key` (`<pdf>#29`) before this fix — `vote_events.py`'s `get_object()` replaces the *whole* match spec with `dedupe_key` alone when present, so the second bill's vote resolved to the first bill's already-imported row instead of creating its own, and the importer raised `DuplicateItemError`.
- `VoteEvent.bill` is a single FK, so this needs one row per (roll call, bill) pair, not one shared row. Fix: fold `bill.identifier` into both the House and Senate `dedupe_key` computations.
- Extracted both computations (previously inline in `scrape_bill()`) into two small, independently testable module-level functions — `_house_vote_dedupe_key()` and `_senate_vote_dedupe_key()` — rather than testing this through the full `scrape_bill()` pipeline, which would need mocking HTML/PDF fetches unrelated to this bug.
- MA's already-scraped data from `ma-f08d7646b9fe` wasn't lost — the collection manifest is still in S3 (`working-tier/ma/ma-f08d7646b9fe/_manifest.json`), so once this lands the load can be retried directly against it without re-running the ~8.2h collection.

Jira: OPEN-252 (Data Quality Enhancements epic). Note: filed originally suspecting `votes.py`'s `HouseVoteRecordParser` (a separate, legacy-ish MA vote scraping path with no dedupe_key at all) as the culprit — deeper investigation with the actual manifest data ruled that out (that path apparently didn't even run in this collection) and found the real, reproducible mechanism above instead.

## Test plan

- [x] New unit tests for `_house_vote_dedupe_key()`/`_senate_vote_dedupe_key()`, including the exact live collision (`H4005`/`H4010` sharing Supplement #29) now producing distinct keys, and same-bill re-scrapes staying idempotent.
- [x] `pytest tests/test_ma_bills.py` — 22 passed.
- [x] Full suite `pytest tests/` — 210 passed, 6 pre-existing failures in `test_classify_motion.py` unrelated to MA (confirmed via `git stash` against main, same failures with or without this change).
- [x] `black --check` clean on `scrapers/ma/bills.py`; `test_ma_bills.py`'s pre-existing formatting drift (predates this PR) left untouched, only my own added lines are black-compliant.
- [x] `flake8` clean on both changed files.


agent-smith-ddp:
Opened by mistake against the wrong repo (this checkout's `gh` default resolved to the upstream project instead of our fork) -- closing immediately, no action needed here. Sorry for the noise.

PR #5808, opened 2026-09-09
KS Scraper timeout fix

harshitasingh360:
This PR speeds up the KS votes scraper (from ~36 hours to ~3–4 hours) by fetching vote detail pages concurrently via a ThreadPoolExecutor rather than sequentially, which also resolves the GCP logging quota errors observed in production. The parsing logic remains unchanged, and because pool.map returns results in their original sequence.

### DIFF
--- a/scrapers/ks/votes.py
+++ b/scrapers/ks/votes.py
@@ -1,6 +1,6 @@
+import concurrent.futures
 import datetime
 import re
-import time
 
 import lxml.html
 
@@ -9,6 +9,13 @@ from openstates.scrape import Scraper, VoteEvent
 
 class KSVoteScraper(Scraper):
     BASE_URL = "https://www.kslegislature.gov"
+
+    # Number of vote detail pages to fetch concurrently. The old scraper
+    # fetched every ~100 KB vote page one-at-a-time, which meant a full
+    # biennium took well over a day to complete. Fetching several pages in
+    # parallel dramatically cuts the wall-clock time without changing what
+    # gets parsed or the order in which votes are emitted.
+    VOTE_FETCH_WORKERS = 12
 
     def scrape(self, session=None):
         meta = next(
@@ -34,7 +41,7 @@ class KSVoteScraper(Scraper):
             # Paginate until no vote rows are returned.
             list_url = (
                 f"{self.BASE_URL}/{biennium}/votes/fragment/"
-                f"?page={page}&per_page=20"
+                f"?page={page}&per_page=100"
             )
 
             try:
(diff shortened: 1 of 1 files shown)


sinchana00 (commented):
LGTM

scrapers/ks/votes.py
```diff
@@ -59,37 +71,55 @@ def scrape(self, session=None):
                 if not vote_url.startswith("http"):
                     vote_url = f"{self.BASE_URL}{vote_url}"
 
+                page_votes.append((bill, vote_url))
+
+            # Fetch the vote detail pages in parallel. Fetching (network I/O)
```
jessemortenson, line 76: I appreciate you thinking out of the box to suggest an improvement here. This does raise a concern: how is this code managing the number of requests we are making to the source website in a given time span? things like

x requests / sec
x requests / min

Historically the repo has had a convention of pretty strictly making requests in serial (no concurrency). The motivation is that we want to be polite/responsible scrapers, and not hit government websites with 1000 requests/min. Taking a government website down reduces transparency (residents can't access it) so that's counter to the mission. As a result of generally having serial requests, the requests-per-minute can be controlled by the `-r` [RPM flag](https://docs.openstates.org/contributing/scrapers/#running-your-first-scraper).

So it looks like this code would break the convention, and probably break the RPM flag as well (I haven't tried though). I'm not strictly opposed to making an exception here, but to consider that I still need to see substantive guardrails in place. What limits does this concurrent model operate under?

harshitasingh360: Thanks for the review, and I fully agree on being a responsible scraper. 
I will revert the changes and think of a better approach.

jessemortenson: One possibility, and this would also be a new thing... if the problem here is that there are many many vote detail pages to fetch, is it possible that we could cache some of them between scraper runs (optionally)? 

We generally don't want to cache bill detail pages, because that information changes regularly. But perhaps a vote detail page is very unlikely to change, once it is initially posted?

I'm speaking naively, because I haven't gone into the details of the KS scraper recently like you have. Maybe you could spot check some on the Wayback Machine (internet archive), if the vote detail pages are archived and have multiple snapshots over time?

Another way to test could be: implement the caching layer, but initially have the scraper check if there is a difference between the previously cached response and the just-fetched response? And if it doesn't report any changes over a week or something, we could feel relatively safe that that those pages aren't changing. Or you could just write a standalone script that caches once and then checks for any new differences with subsequent fetches over time, and just run that locally daily for a week or so.

A response caching system could be valuable down the road, so experimenting a bit may pay off. The scrapelib/openstates-scrapers does have a caching system (that is invoked via `--fastmode` flag at runtime), however a) I'm pretty sure it caches all requests, which we don't want, and b) uses local filesystem to store data (which would not persist between scraper DAG runs). A basic caching system for scrapers would probably  need to use a Google Cloud Storage bucket (maybe with a 30 day object expiration or something) and a consistent object naming convention to be able to fetch those.

Well: caching is a non-trivial option to explore. So if you see easier ideas to try, then feel free to put this suggestion on the shelf.

harshitasingh360: This approach makes sense to me. Let me try doing more research on this and try implementing this approach.

